PR #63, opened 2014-01-06
Add shallow submodule init support.

monaka:
As you know, recent git (since 1.8.x) supports shallow submodule cloning.
This patch is just to add API method.
I'll send pull requests to git-plugin after this patch was approved.


cdbennett:
:+1: I would like to see this implemented.  Maybe it's not perfectly ideal design, but it would be great to have in projects with many submodules having a large history

PR #158, opened 2015-01-22
Allow clean to remove abandoned submodules

Vynce:
"git clean -fdx" will not remove submodule folders that are no longer
relevant. This commit adds an additional -f argument so that unused
submodule folders are also removed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -514,7 +514,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     public void clean() throws GitException, InterruptedException {
         reset(true);
-        launchCommand("clean", "-fdx");
+        launchCommand("clean", "-ffdx");
     }
 
     public ObjectId revParse(String revName) throws GitException, InterruptedException {
@@ -800,7 +800,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     	if (recursive) {
             args.add("--recursive");
     	}
-    	args.add("git clean -fdx");
+        args.add("git clean -ffdx");
 
     	launchCommand(args);
     }


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -800,7 +800,7 @@ public void submoduleClean(boolean recursive) throws GitException, InterruptedEx
     	if (recursive) {
             args.add("--recursive");
     	}
-    	args.add("git clean -fdx");
+        args.add("git clean -ffdx");
```
MarkEWaite: I'd rather not have the indentation change.  Can you put the white space on the front of the line back the way it was before?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -514,7 +514,7 @@ public void execute() throws GitException, InterruptedException {
 
     public void clean() throws GitException, InterruptedException {
         reset(true);
-        launchCommand("clean", "-fdx");
+        launchCommand("clean", "-ffdx");
```
MarkEWaite: The last time I considered adding this, I think I found cases that warned me against it while writing tests.  

Could you provide at least one test which confirms that this truly does what you expect it to do?

MarkEWaite:
Can you provide a similar implementation for JGit so that we keep the JGit based functionality as close to parity with the command line git implementation as we can?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
Vynce: Sure, no problem. There's a mix of tabs and spaces used for indenting in several places, but perhaps that should be cleaned up in a separate commit.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
Vynce: Will do.

Vynce:
I'll see if I can figure out how to perform the same operation with JGit.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
MarkEWaite: I recommend against submitting a separate commit to cleanup indentation.  I'm afraid I'll just reject an indentation cleanup pull request.  That's not how this particular project chooses to operate.

Early in my submissions to the plugin I proposed that we switch the plugin to use an automatic formatting step in the maven build process. I've used automatic formatting on projects at work for years and have been very happy with the diff minimization and format standardization which results.  The other developers on the project were pretty strongly opposed to the automatic formatting idea (creates a "diff wall", damages the layout of calls to fluent interfaces, etc.)

I've accepted that, for this project, I will minimize white space changes from one commit to the next, and will ignore the inconsistencies which exist within the existing source base.

MarkEWaite:
Please refer to the discussions in https://github.com/jenkinsci/git-client-plugin/pull/163 for more information about the impact of this type of change.

MarkEWaite:
:-1: When a matrix project is checked out, in some cases the resulting workspace on the node which hosts the root of the matrix build will include a checkout of the repository.  Beneath that root checkout are any of the actual builds which run on that matrix build.

With that structure (root repo, with nested workspaces under the root repo), a "git clean -xffd" at the root checkout will delete the nested workspaces.

For example, I have a root of a matrix workspace where git clean -xfd reports:

```
jenkins@mark-pc1:~/jobs/JENKINS-26660-submodule-checkout-does-not-clean-deleted-submodules/workspace$ git clean -n -xfd
Would skip repository jdk/java-6/label/master
Would skip repository jdk/java-7/label/master
Would skip repository jdk/java-8/label/master
Would skip repository label/master
Would skip repository modules/firewall
Would skip repository modules/ntp
```

The same command with the "extra f" reports:

```
jenkins@mark-pc1:~/jobs/JENKINS-26660-submodule-checkout-does-not-clean-deleted-submodules/workspace$ git clean -n -xffd
Would remove jdk/
Would remove label/
Would remove modules/
```

PR #160, opened 2015-01-30
Set push.default = simple

martinda:
Submitting for review.


martinda:
I decided the user should change their global .gitconfig file and remove problematic line in the push section (default = nothing) rather than change the test.

PR #163, opened 2015-03-23
More complete "clean" behavior.

markdoliner:
Specifically, remove untracked directories even if they're managed as
part of another git repository. See the documentation for the -d switch
in "git help clean" for a full description.

This _seems_ like the right behavior to me. I can't think of a time
when it would make sense to not delete an untracked directory just because
it happens to be a git repo.

I haven't tested this. Haven't tried to compile. Haven't run unit tests.
Etc.

I was experiencing a problem where part of my build process downloads a
library from the internet. The library contains a .git file with a .gitdir
directive. This caused "git clean -fdx" to error-out during the next build
attempt when run as part of the "Clean after checkout" Git SCM additional
behavior, because it refused to delete the library.

You might be able to reproduce the problem by creating a simple project
in Jenkins. The test project will need to be checked out via Jenkins' Git
source code management plugin. As part of your build steps, add this:
    mkdir some_test_dir
    echo 'gitdir: /foo/bar' > some_test_dir/.git
Then add the "Clean after checkout" additional behavior under Source
Code Management.

Related bug reports:
https://issues.jenkins-ci.org/browse/JENKINS-21369
https://issues.jenkins-ci.org/browse/JENKINS-23694

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -588,7 +588,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      */
     public void clean() throws GitException, InterruptedException {
         reset(true);
-        launchCommand("clean", "-fdx");
+        launchCommand("clean", "-ffdx");
     }
 
     /** {@inheritDoc} */
@@ -907,7 +907,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     	if (recursive) {
             args.add("--recursive");
     	}
-    	args.add("git clean -fdx");
+    	args.add("git clean -ffdx");
 
     	launchCommand(args);
     }


MarkEWaite:
The git client plugin is used by so many users that I'm hesitant to change fundamental behavior without the benefit of tests which confirm that the new behavior is correct and complete.  Can you provide tests which show there is a problem without your commit?

You might refer to the comments on https://github.com/jenkinsci/git-client-plugin/pull/158 for more details on what I think is needed before this type of pull request would be accepted.

For example, I think that if we need more complete clean behavior, we need it for both the CliGit implementation and the JGit implementation.  I think that if we need more complete clean behavior, we need one or more automated tests which confirm that the clean behavior is correct for both the CliGit and JGit implementation.

markdoliner:
MarkEWaite: That's a totally reasonable request. Unfortunately I don't know when I'll be able to spend time writing a test (and figuring out how to build the code and run the test, etc).

If anyone else sees this comment, wants to work on this, and has time, please feel free! I'd certainly appreciate it!

GyrosGeier:
This has the ugly side effect of wiping out submodules that still exist though, and those have to be re-downloaded.

This could be avoided by using "git -d clean.requireForce=false clean -dx", but then again that wouldn't remove submodules that are no longer present in the current checkout.

MarkEWaite:
It also has the side effect of ignoring the existing implementation of [submodule clean](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L903)  and [submodule reset](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L887) which are already in the plugin.

I suspect that submoduleClean and submoduleReset don't meet the need that the original submitter is trying to address.  There are enough different use cases to address that I think we really need test cases before this type of behavioral change is accepted into the plugin.

GyrosGeier:
The original use case appears to be "I have dropped a submodule, and would like to have the subdirectory removed when updating the master project", which is a fair expectation in a CI environment -- however the additional "force" flag wipes all submodules, not just those that were dropped.

As soon as a single "force" flag is in effect, the default behaviour of silently skipping directories with their own .git directory goes out of the window. Because clean.requireForce defaults to true, everyone just uses a force flag.

I'm writing a message to the git mailing list outlining the problem.

dasein:
+1 

This issue is hitting us now because of the use of GO in our build environments. A lot of GO projects checkout repos into their build environment using "go get" (e.g. src/github.com/golang/glog). Although the directory where these repos are checked out are included in .gitignore, they do not get cleaned up with git clean -fdx:

```
$ git clean -fdx
Skipping repository src/github.com/armon/go-metrics
```

vs 

```
$ git clean -ffdx
Removing src/
````

Obviously there are workarounds to this problem, but it would be a nice feature to include in the plugin.

**Comment by MarkEWaite** [comment]
I can see use cases (like the "go get" case described by @dasein ) where it would be good to have the more aggressive clean that is described in this pull request.  However, this pull request makes the more aggressive clean the default and seems like it risks frustrating or hindering more users than it helps.

Would it be better, safer and less intrusive to modify the "Clean before checkout" and "Clean after checkout" extensions to allow an additional checkbox to clean even subdirectories containing a ".git" directory?

Submodules in the current repository which are initialized (`submodule init`) and have been updated (`submodule update`) seem to be left unharmed by the `git clean -xffd` on the git versions I checked (git 2.4.2, git 1.7.1, git 1.7.10.4).  Subdirectories which contain a .git directory but are *not* a result of submodule initialization are left untouched by the current `git clean -xfd` and are removed by `git clean -xffd`.

I continue to believe that a change like this definitely needs to be surrounded by tests which show the various interesting failure modes and success modes related to use with submodule processing enabled vs. disabled, with workspaces constructed using techniques like the "go get" case, etc.

**Comment by GyrosGeier** [comment]
Problem is, in a matrix build these submodules aren't abandoned.

**Comment by MarkEWaite** [comment]
:-1: for the proposed pull request specifically for the reason mentioned by @GyrosGeier .  When a matrix project is checked out, in some cases the resulting workspace on the node which hosts the root of the matrix build will include a checkout of the repository.  Beneath that root checkout are any of the actual builds which run on that matrix build.

With that structure (root repo, with nested workspaces under the root repo), a "git clean -xffd" at the root checkout will delete the nested workspaces.

For example, I have a root of a matrix workspace where git clean -xfd reports:

```
jenkins@mark-pc1:~/jobs/JENKINS-26660-submodule-checkout-does-not-clean-deleted-submodules/workspace$ git clean -n -xfd
Would skip repository jdk/java-6/label/master
Would skip repository jdk/java-7/label/master
Would skip repository jdk/java-8/label/master
Would skip repository label/master
Would skip repository modules/firewall
Would skip repository modules/ntp
```

The same command with the "extra f" reports:

```
jenkins@mark-pc1:~/jobs/JENKINS-26660-submodule-checkout-does-not-clean-deleted-submodules/workspace$ git clean -n -xffd
Would remove jdk/
Would remove label/
Would remove modules/
```

bukzor:
In that case, you want `git clean -fdxf -e /jdk -e /label -e /modules`.

The `-e` option adds exclusions to the clean which are not ignored by the `-x` option.

bukzor:
What's more, we don't use and likely never will the matrix plugin in our large jenkins deployment, but we'd still like to have this functionality. Currently our only option is to disable the "clean" option from the git plugin in all our jobs, then add a step to run `git clean -fdxf`. Is this the intended workflow in that case?

MarkEWaite:
I don't think you understood the complete set of reasons why the pull request was closed without being merged.

In my ordered priority for reasons not to accept this pull request:

1. Change of default behavior is likely to harm many more users than it helps
2. No tests means that each plugin release must be interactively tested to assure this did not regress
3. Impact of clean -xffd on submodules is likely to frustrate submodule users
4. Impact of clean -xffd on matrix projects is likely to frustrate matrix project users

More details on those concerns:

## Change of default behavior

The change of default behavior for all users of the plugin means that all users are forced to accept the new behavior.  Your proposed change will change the behavior for over 60 000 Jenkins installations.  I'm confident there are many users of the existing behavior in those 60 000+ installations.  If I accept your pull request, then I'll likely have to implement another change to make it optional and to restore the original default behavior.

## No Tests

The absence of tests from the pull request means that interactive testing will be needed for each plugin release to assure the change is still having the desired result.  If I accept your pull request "as-is", I accept the responsibility to either write the tests myself, or to test the plugin at every release to assure that your change of behavior has not been lost.

## Submodules

Removing submodules with a clean will require that they are downloaded again.  Some submodules are large and mostly static, so users will see the repeated download of those submodules as a waste of time and bandwidth.

## Matrix projects

The directory names are examples from a common plugin use case, but they are definitely not the only things that would need to be excluded.

If you really want "clean -xffd", you'll need to add it as an additional, optional behavior, with tests to confirm the new behavior is working as expected.

Could you explain why you need to disable the clean option in your jobs?  If you decide to add a "really-clean" (clean -xffd) as part of the commands in your job, I don't think that prevents you from also using the plugin's own "clean before checkout" or "clean after checkout" optional behaviors.

bukzor:
I did see the other arguments, but only the fourth is valid, so out was the
only one I addressed.

1. This only differs from existing behavior in a small edge case, wherein
the change results in the expected behavior: a pristine git checkout. I
sincerely doubt you'll find anyone complaining that their git-clean has
resulted in a git that's too clean. Such people don't, in fact, want their
gits cleaned. I believe that the status quo exists solely because the
original dev didn't know of -ff, or they would have used it.
2. The community has already offered to write tests. The only reason it
isn't done is upstream has already rejected the change. I'll personally
write the tests if I have an indication that this is an otherwise
acceptable change.
3. Submodules are treated as tracked files, and won't be cleaned regardless
of the number of -f options given. I can demonstrate this upon request. So,
-ff will only result in *removed* submodules being deleted, which is a
strictly desirable behavior for such projects.

This brings the discussion back to the start of my previous post.

MarkEWaite:
I think we simply disagree on the most crucial point.  You say that it differs from existing behavior in a small edge case (multi-configuration projects, which I use frequently), while I say that the difference from existing behavior is a risk to the many users of the plugin.

Why not take the earlier suggestion and make the added clean argument optional instead of changing the default behavior of the plugin?

If you're willing to write tests for the change, aren't you also willing to make the change in a way that preserves compatibility for existing users?

I seem to have missed where the community has offered to write tests for this change.  I'm not sure how I missed that, since I've read through at least 3 different pull requests proposing the same one line change, with no tests to verify the change in behavior from before the change to after the change.

I'm not willing to change the default behavior of the plugin, because I'm aware of at least one case where it has unexpected results (multi-configuration projects).  I am willing to consider the change as an optional setting.

PR #164, opened 2015-04-02
Fix failure in test_push_tags

dbroady1:
Currently (with 1.16.1), when testing git-client-plugin, I am getting
the following error in test_push_tags:

--------------------------------------------------------
The current branch master has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream /tmp/hudson4099265862244982437test master
--------------------------------------------------------

The reason that this is happening is that test_push_tags() is using the
PushCommand() pattern, but is failing to specify a refspec.  Also, this
test case is creating a remote url, but is not setting up the master
branch to track that remote url.  So, without a refspec, the behaviour
depends on whether CLI or JGIT is used.

In CLI, it uses whatever push.default specifies.  In my case, it was set
to tracking/upstream.  With this setting, it requires that the upstream be
configured correctly, or it will error out with the above error message.  I
tested with all other values, and only the matching value passed the test.

In JGIT, it uses Transport.REFSPEC_PUSH_ALL.  This corresponds to
push.default = matching, which means that it works.

If you are wondering why this was not hit before, by default, push.default
is matching.  Also, all other instances of git.push() in the test harness
uses the 2 argument version, which requires the refspec().  Hence, no problem.

To make this test deterministic, specifying the refspec via
.ref("master"). fixes it for both configurations.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -929,7 +929,7 @@ public abstract class GitAPITestCase extends TestCase {
         /* Add tag to working repo and without pushing it to the bare repo */
         w.tag("tag1");
         assertTrue("tag1 wasn't created", w.git.tagExists("tag1"));
-        w.git.push().to(new URIish(bare.repoPath())).tags(false).execute();
+        w.git.push().ref("master").to(new URIish(bare.repoPath())).tags(false).execute();
         assertFalse("tag1 wasn't pushed", bare.cmd("git tag").contains("tag1"));
 
         /* Add tag to working repo without pushing it to the bare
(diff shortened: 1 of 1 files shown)


MarkEWaite:
I think you've discovered a very interesting condition that we should explore further.  Global settings affect the behavior of the plugin (at least according to a recent bug report), and they may affect the JGit implementation differently than the CliGit implementation.

Any suggestions of techniques to permute the relevant global configuration values so that the tests can confirm expected behaviors with the various global configurations?

dbroady1:
For the JGIt implementation, I think that the first step is to have the push code check the default value of the push.default and change the ref setting appropriately.  After that, then you could probably setup a proper test harness that loops over possible values of this particular default, and calls the actual PushCommand in a try-catch, and during the catch does whatever is tied to that particular default.  It would have to be per-test case setup, because some of the test cases don't setup the proper parent and some do.

PR #165, opened 2015-04-06
(WIP) Allow JGit to perform Submodule operations.

dbroady1:
The code in CliGitAPIImpl is sufficiently general as to not require
anything special from the command-line.  So, we can just move it to
LegacyCompatibleGitAPIImpl to allow to use JGit with submodules.

This fixes bug JENKINS-26026.

There are no extra test cases needed, as the exists ones that test submodule support for the CLI are sufficient for testing JGit.

BTW, this is just the initial submodule support.  I will submit another pull request for getting the SubmoduleUpdate commands for --remote and --recursive to work for JGit.


jacob-keller:
I would advocate for removing this particular functionality from users of git-client, as the JGit implementation has a succinct comment of "who uses this" for it's empty implementation. In addition, it seems very very fishy.. Too smart for its own good.

MarkEWaite:
I'm unwilling to remove that legacy functionality from the command line git implementation because I can't predict the portion of the 60000+ installations which depend on it. I don't think we should spread that implementation into JGit, but I'd rather not risk breaking any existing user.

jacob-keller:
That makes sense. I think long term it would be something to consider for a large version bump, because it seems like a concept that isn't very git-like, but it doesn't appear to break anything right now. Definitely avoiding moving it into JGit makes sense.

PR #166, opened 2015-04-10
Use Java 6 compatible equals method instead of Java 7 specific method

MarkEWaite:
No need to make the git client plugin fail to compile on Java 6 just
for the convenience of a single new class in Java 7.  The Java 7
java.util.Objects class is a nicety, not a necessity.

@ndeloof - any objection to me fixing my mistake by removing the
dependency on Java 7 which I introduced earlier this month?

### DIFF
--- a/src/main/java/hudson/plugins/git/Revision.java
+++ b/src/main/java/hudson/plugins/git/Revision.java
@@ -156,6 +156,9 @@ public class Revision implements java.io.Serializable, Cloneable {
             return false;
         }
         Revision other = (Revision) obj;
-        return java.util.Objects.equals(sha1, other.sha1);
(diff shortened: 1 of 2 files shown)


MarkEWaite:
This pull request passed all the tests on Java 6 and Java 7 on Debian, Ubuntu, CentOS, Windows 8 x64 and Windows 7 x86.

KostyaSha:
Another possible variant can be copy this method to git utils class and put `//TODO remove after java7 switch`
```
    public static boolean equals(Object a, Object b) {
        return (a == b) || (a != null && a.equals(b));
    }
```

MarkEWaite:
Rebased onto the master branch after tests confirmed passing

PR #167, opened 2015-05-09
(WIP) Fix for submodule authentification JENKINS-20941

dstraub:
Because git credential helper doesn't work for submodules (the submodule config is only avaiable after the first submodule update --init - and this requires authentification ...), the submodule url in .git/conf is updated with username/password. The url  is restored after the update, so that no credentials remains in the config.
Also for subsequent updates is the same procedure used, because thats easier insted to set/remo credentials helpers in .git/modules/submodulename/conf .


MarkEWaite:
I'm closing this pull request in favor of the work that has been progressing on PR180.  The PR180 changes are larger, but are supported by tests and are looking reasonable during interactive testing as well.

PR #168, opened 2015-05-13
JENKINS-20879 - make sure $SSH_ASKPASS is not ignored

thomasgl-orange:
I'm also having the errors described in JENKINS-20879 when trying to poll a git repository using a passphrase-protected SSH private key (with the GitCli implementation - as already mentionned in the Jira report, JGit already works fine).  
Looking at the implementation, I understand that the passphrase is provided to git/ssh via a shell script and the `$SSH_ASKPASS` trick.  But according to the ssh manpage, this can only work if `$DISPLAY` is set, because this mechanism is usually dedicated to graphical environments.

     SSH_ASKPASS           If ssh needs a passphrase, it will read the passphrase from the current terminal if it was run from a
                           terminal.  If ssh does not have a terminal associated with it but DISPLAY and SSH_ASKPASS are set, it
                           will execute the program specified by SSH_ASKPASS and open an X11 window to read the passphrase.
                           This is particularly useful when calling ssh from a .xsession or related script.  (Note that on some
                           machines it may be necessary to redirect the input from /dev/null to make this work.)

So, the idea of this PR is pretty simple: ensure some `$DISPLAY` variable is exported in the SSH wrapper script (the `$GIT_SSH` one), so that `$SSH_ASKPASS` actually works. I'm assuming it won't have too much other side effect on the `git` command behavior.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1530,6 +1530,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         File ssh = File.createTempFile("ssh", ".sh");
         PrintWriter w = new PrintWriter(ssh);
         w.println("#!/bin/sh");
+        // ${SSH_ASKPASS} might be ignored if ${DISPLAY} is not set
+        w.println("[ -z \"${DISPLAY}\" ] && export DISPLAY=:123.456");
         w.println("ssh -i \"" + key.getAbsolutePath() + "\" -o StrictHostKeyChecking=no \"$@\"");
         w.close();
         ssh.setExecutable(true);


MarkEWaite:
I merged your pull request into a private repository and deployed it to my Jenkins server with access to various Debian, Ubuntu, CentOS, and Windows machines. As far as I can tell from my initial experiments, it is not working on the environments I'm testing.

However, since authentication is an easy place for me to make mistakes, I'll continue testing to see if my mistakes are somehow hiding the positive results of your proposed change.

Here are some of the steps I've taken as part of my verification attempts:
# Create a git repository that is only accessible through ssh
# Create a Jenkins job attempting to access that ssh based git repo, without credentials, confirm expected failure
# Add a passphrase protected private key to Jenkins as a global credential
# Use that passphrase protected private key with the Jenkins job (without yet allowing that key access to the git repository), confirm expected failure
# Add the public key of the passphrase protected private key to ssh account so that git repo is accessible
# Confirm from a shell that git repo is not accessible without passphrase
# Start ssh-agent in shell, add passphrase for private key, confirm git repo is accessible with passphrase
# Run Jenkins job which uses passphrase protected private key, confirm unexpected authentication failure

MarkEWaite:
Using the git-client-plugin pre-release build which includes your change, I was able to clone with command line git from a Debian 6 passphrase secured private key account to 

* Debian 6 x86
* Debian 7 x64
* Debian 8 x64
* CentOS 6 x86
* CentOS 7 x74

The fetch failed on

* Windows 7 x86
* Windows 7 x64
* Windows Home Server 2011 x64
* Windows 8.1 x64

The steps I took to duplicate the case included:

1. Create a new account on Debian 6 x86
2. Create a passphrase secured ssh private key for that account
3. Create a new git repository only accessible to that user
4. Attempt to clone that repository from a Jenkins server, confirm it fails as expected
5. Add the passphrase protected private key, confirm the clone succeeds
6. Do the same with a multi-configuration job to use all available computers

Are you willing to attempt to fix the same condition for Windows?

MarkEWaite:
I think we should include this in the plugin.  I've detected no negative side effects, and have proven to my own satisfaction that it improves the situation with the command line git client on Linux.  

I reverted to git client plugin 1.17.1 (before this fix) and confirmed that the multi-configuration job started immediately failing on Linux, both in the existing job and in a newly created copy of the job.

I again installed the pre-release of the plugin with this change, and confirmed that the Linux configurations of the job began passing immediately.

The change does not help Windows as is, and the same technique (setting a DISPLAY environment variable in the SSH_ASKPASS script) does not help either.

@ndeloof, I recommend that we merge this into the plugin because it improves the current situation, even if it does not entirely solve the problem.

ndeloof:
:+1:

PR #171, opened 2015-07-09
[JENKINS-21520] transient NoClassDefFoundError: javax/servlet/ServletException

ndeloof:


ndeloof:
Not sure why having a reference to Functions do drag some unexpected classes on slave then breaks ClassLoading, anyway inlining this on-line functions seems a reasonable workaround
@reviewbybees

MarkEWaite:
Doesn't the reference also need to be removed from src/main/java/hudson/plugins/git/GitTool.java?

Shouldn't it also be removed from src/test/java/org/jenkinsci/plugins/gitclient/WarnTempDirValueTest.java to reduce the risk that someone will see its use in a test and introduce it back into the main code?

ndeloof:
makes sense indeed

stephenc:
:bee:

PR #172, opened 2015-07-10
Added PushResult evaluation

kristian:
If JGit PushCommand call() returns without an exception, it does generally not mean all / any items have been pushed successfully. PushResult has to be evaluated to be sure.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
@@ -10,6 +10,9 @@ import static org.eclipse.jgit.lib.Constants.R_HEADS;
 import static org.eclipse.jgit.lib.Constants.R_REMOTES;
 import static org.eclipse.jgit.lib.Constants.R_TAGS;
 import static org.eclipse.jgit.lib.Constants.typeString;
+import static org.eclipse.jgit.transport.RemoteRefUpdate.Status.OK;
+import static org.eclipse.jgit.transport.RemoteRefUpdate.Status.UP_TO_DATE;
+
 import hudson.FilePath;
 import hudson.Util;
 import hudson.model.TaskListener;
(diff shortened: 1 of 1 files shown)


MarkEWaite:
It compiles and passes automated tests.  Unfortunately, the fact that it passes automated tests worries me, since that means the automated tests were not evaluating the [bug](https://issues.jenkins-ci.org/browse/JENKINS-20393) that JGit push does not seem to actually make the pushed change visible on the remote repository.

KayK:
Hey Mark, I agree. I checked the test org.jenkinsci.plugins.gitclient.GitAPITestCase and it doesn't seem to contain any negative test case for push. Therefore the tests pass. The expected behaviour for a failing push is an exception to be thrown, not even this is tested. If there would have been a test for this, the missing code would have been noticed.

KayK:
Mark, if you have an working example, where nothing is pushed like described in [bug](https://issues.jenkins-ci.org/browse/JENKINS-20393), this could make a good test case. I assume before the change no execption is thrown. After the pull, an exception should be thrown. What do you think?

kristian:
As reported in the [bug](https://issues.jenkins-ci.org/browse/JENKINS-20393), the build is failing if I recompile the plugin with the change.

kristian:
This pull should also allow error diagnosis for other push realted issues. It's quite late here, therefore I will try to provide a negative test case for this (similar to the [bug](https://issues.jenkins-ci.org/browse/JENKINS-20393)) tomorrow. After that you it should be fine to pull this one. For the bug I will try to find the root-cause and open another pull request.

MarkEWaite:
I believe that no push from JGit has ever worked with the plugin. I'd rather the fix for the bug be included in the pull to the git client plugin along with this change.  Showing a different diagnostic to the user is not much help, since there is nothing they can do about the failure (at least in the case I saw).

kristian:
Agreed. I will dig into this tomorrow.

kristian:
As stated in [JENKINS-20393](https://issues.jenkins-ci.org/browse/JENKINS-20393) the fix for the bug is part of the git-plugin, rather than the git-client-plugin. Therefore this pull request should be complete and should help to locate any future push related issues easier.

PR #173, opened 2015-07-11
Workaround for JGit unable to parse HEAD:branch

kristian:
Workaround for JGit unable to parse HEAD:branch, see [JENKINS-20393](https://issues.jenkins-ci.org/browse/JENKINS-20393) for details.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
@@ -1837,15 +1837,19 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 if("HEAD".equals(specs[0])) //special case for HEAD, if e.g. HEAD:branch, HEAD should be pushed to branch, we can't parse head, so assume to push branch:branch
                 	specs[0] = specs[1];
                 for(int spec=0;spec<specs.length;spec++)
-                	if(spec==0&&specs[spec].isEmpty())
-                		continue; //empty for the first spec. if fine (see https://github.com/eclipse/jgit/blob/master/org.eclipse.jgit/src/org/eclipse/jgit/transport/RefSpec.java#L104-L122)
-                	else if(!specs[spec].startsWith("refs/")&&!specs[spec].startsWith("+refs/")) {
-                		if(specs[spec].isEmpty() //generally means to push "matching" branches, hard to implement the right way, simple-fix here
-                		|| specs[spec].equals("HEAD")) //should read <git-dir>/HEAD file, which is also quite hard, simple-fix here aswell
+                	if(specs[spec].isEmpty()) { //generally means to push "matching" branches, hard to implement the right way, simple-fix here
+                		if(spec!=0) //empty for the first spec. if fine (see https://github.com/eclipse/jgit/blob/master/org.eclipse.jgit/src/org/eclipse/jgit/transport/RefSpec.java#L104-L122)
                 			specs[spec] = "refs/heads/master";
-                		else if(specs[spec].contains("/")) //assume spec. already contains subfolder, just prepend refs/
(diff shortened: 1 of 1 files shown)


kristian:
Assumed that branches may not contain /. Fixed fixRefSpec implementation.

kristian:
I think [`PushTest`](https://github.com/MarkEWaite/git-client-plugin/blob/master-JGit-push/src/test/java/org/jenkinsci/plugins/gitclient/PushTest.java) is a past due test case. I especially like the idea of considering both implementations (Git & JGit) in the same test case. But note, that some test cases _will_ behave differently for Git & JGit, due to the straight forward implementation of `fixRefSpec` in `JGitAPIImpl`, such as:

* ":master":
  * Git: The special refspec : (or +: to allow non-fast-forward updates) directs Git to push "matching" branches.
  * JGit: `fixRefSpec` will change the ref. spec. to "refs/heads/master:refs/heads/master"
* "HEAD:master"
  * Git: Is going to read <git-dir>/HEAD to determine the current HEAD pointer
  * JGit: `fixRefSpec` will apply the same branch to be pushed to, to be the source branch, so will choose "refs/heads/master:refs/heads/master" in this case.

Other than that, I also like the test implementation using `@Parameterized.Parameters`. Missing test cases are:

* Branches
** Empty branches
** Branches with forward slashes
** Push from one local branch to another remote branch
* Negative Test Cases (If there are any?)

MarkEWaite:
You mentioned that the test case should handle empty branches.  An empty string on the left hand side of a push refspec is interpreted as a request to delete the branch named on the right hand side.  If I want to delete the remote branch "feature/xyzzy" on the remote named "origin", I use the command:

{noformat}
git push origin :feature/xyzzy
{noformat}

kristian:
I changed the implementation of fixRefSpec to resolve the refs. from the repository! :+1: This should actually fix a lot of issues and should be good enough for 95% of the cases. Remember, before the change, the implementation wasn't working with JGit at all...

The local refs. are taken from the repository. For the remote ref. I still try to determine it "manually". Actually I also found out that JGit supports pushing e.g. "HEAD:/refs/heads/master". A quick test for possible combinations:

```
: fixed to :refs/heads/master
master fixed to refs/heads/master:refs/heads/master
:master fixed to :refs/heads/master
HEAD fixed to HEAD:refs/heads/master
HEAD:master fixed to HEAD:refs/heads/master
HEAD:branch fixed to HEAD:refs/heads/branch
master:master fixed to refs/heads/master:refs/heads/master
branch:branch fixed to refs/heads/branch:refs/heads/branch
master:branch fixed to refs/heads/master:refs/heads/branch
:refs/heads/master fixed to :refs/heads/master
refs/heads/master fixed to refs/heads/master:refs/heads/master
refs/heads/master:branch fixed to refs/heads/master:refs/heads/branch
branch:refs/heads/master fixed to refs/heads/branch:refs/heads/master
tag:tags/tag fixed to refs/tags/tag:refs/tags/tag
tags/tag fixed to refs/tags/tag:refs/tags/tag
foo/bar fixed to refs/heads/foo/bar:refs/heads/foo/bar
foo/bar:baz/qux fixed to refs/heads/foo/bar:refs/heads/baz/qux
```

MarkEWaite:
@kristian, I rebased your changes into a single commit (authored by you) and then applied changes that I think are important before this fix is merged to master.  Would you review the changes in 5c08ddc5d7a6fa67a2f3d009898681b766cd8692 and 5461726872f954fbe6e8f39bd852e0aa2f5b114c and let me know if you agree those are correct changes?

I would prefer to have those two changes included with you as the author as part of a single commit that I would merge to master.  That would avoid having the formatting fix visible in the history of the master branch.  That would mean you would need to submit a pull request which included your original changes and those two changes squashed into a single commit.

Are you willing to do that squash of changes into a single commit for JGitAPIImpl?

PR #176, opened 2015-07-24
JENKINS-29604 - Git CLI client does not propogate username

jacob-keller:
The CLI implementation of the Git client does not correctly propogate the username for ssh urls and possibly for http(s) urls. To fix this, teach the GIT_SSH scripts for Unix and Windows how to pass the username via "-l" option. In addition, for consistency, include credentials.username with the default user incase http(s) urls don't have user names assigned.

This enables use of a different remote username than the local system username, which seems to be something that was not tested today.

I am not familiar with the test framework used by this plugin, so I wasn't able to figure out how the tests were being run. Ideally we need to provide a test which exhibits this behavior so that we can catch it in the future. Help and insight regarding this would be useful.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -3,6 +3,7 @@ package org.jenkinsci.plugins.gitclient;
 
 import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
+import com.cloudbees.plugins.credentials.common.UsernameCredentials;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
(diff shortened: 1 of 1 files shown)


MarkEWaite:
The CredentialsTest is configured to use a locally visible file and directory structure for authentication test data.  I currently have several different tests included in my ~/.ssh/auth-data/repos.csv file.  The tests I have include:

{code}
* git@bitbucket.org as user markewaite with an ssh private key
* git@github.com to a private repo as user MarkEWaite with an ssh private key
* git@github.com to a private repo as user jenkins with an ssh private key
* git@github.com to a public repo as user MarkEWaite with an ssh private key
* git@github.com to a public repo as user jenkins with an ssh private key
* git@gitlab.com to a private repo as user MarkEWaite with an ssh private key
* git@gitlab.com to a public report as user MarkEWaite with an ssh private key
* http://localhost:8080/userContent.git (Jenkins userContent) as user mwaite with ssh private key
* https://github.com to a public repo as user mwaite
* https://username:password@gitlab.com to a public repo
* https://username:password@bitbucket.org to a public repo
* debian8.markwaite.net:/var/lib/git/mwaite/bin.git as user mwaite
* ssh://wheezy64b:45022/var/lib/git/mwaite/bin.git as user mwaite
{code}

I've run those test scenarios successfully on Windows 7, Windows Home Server 2011, Windows 8.1, Debian 6 x86, Debian 7 x64, Debian 8 x64, CentOS 6.6 x86, CentOS 7 x64.  As far as I can tell, your change is good in all those configurations.

MarkEWaite:
I think (b) is the better choice, but I'd like a review from @ndeloof before that change could be brought into the master branch.

The request for authenticated submodule support is the single highest voted enhancement request for the git plugin, so it would be a real victory to have a working, functional authenticated submodule access system working in the plugin.

The tests run with "mvn clean install".  Since they aren't running for you, I assume there is some important difference between your development environment and the environment used by others.

As some questions:
* Are you running MacOS or FreeBSD or OpenBSD or HP-UX or AIX or Solaris?  I develop on Ubuntu Linux and test on Windows, CentOS, Debian, and Ubuntu
* Is the git executable in your path?
* Are you running Windows with cygwin rather than the recommended msysgit?

PR #178, opened 2015-07-30
add and fix tests

jacob-keller:
I figured out that we can't do the pushNonFastForward test without first pushing some other commit, so I did some re-work and reworked the pushNoRefspec into pushCurrent and explicitly configured it to push "HEAD" so that we can get the remote ready.

It might be worth re-working how the test is run so that it is not inter-dependent, but I couldn't figure out how to do this.

The new test works with both push.default = simple and push.default = upstream, so I believe this should be correct behavior now.

We do drop testing of "no refspec" but I think that is really a broken test concept anyways unless we're going to find a way to completely 100% clear configuration during tests.

I still have some other errors which I am currently working on determining how to fix, mostly I think related to proxies. I also need to change the default remote tests as suggested before.

This series adds a submodule test, and fixes two tests which broke due to configured changes on my setup.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -1825,6 +1825,18 @@ public abstract class GitAPITestCase extends TestCase {
         );
     }
 
+    public void test_submodule_update() throws Exception {
+        w.init();
+        // Re-use the same submodule test branch
+        w.launchCommand("git","fetch",localMirror(),"tests/getSubmodules:t");
+        w.git.checkout("t");
+        w.git.submoduleInit();
+        w.git.submoduleUpdate().execute();
+
+        assertTrue("modules/firewall does not exist", w.exists("modules/firewall"));
+        assertTrue("modules/ntp does not exist", w.exists("modules/ntp"));
+    }
+
     @NotImplementedInJGit
     public void test_trackingSubmoduleBranches() throws Exception {
         if (! ((CliGitAPIImpl)w.git).isAtLeastVersion(1,8,2,0)) {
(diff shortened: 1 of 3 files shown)


MarkEWaite:
I'd like to investigate PushSimpleTest further to understand why that dependency exists between tests. The BeforeClass method in PushTest is definitely called during the run of PushSimpleTest, so the bare repository should already include a master branch, yet the failing test seems to be clear proof that there isn't a master branch.  That probably means there aren't enough assertions in the BeforeClass method in PushTest, because I think it is doing something that it is not actually doing.

jacob-keller:
On Jul 31, 2015 6:44 AM, "Mark Waite" <notifications@github.com> wrote:
>
> I'd like to investigate PushSimpleTest further to understand why that
dependency exists between tests. The BeforeClass method in PushTest is
definitely called during the run of PushSimpleTest, so the bare repository
should already include a master branch, yet the failing test seems to be
clear proof that there isn't a master branch. That probably means there
aren't enough assertions in the BeforeClass method in PushTest, because I
think it is doing something that it is not actually doing.

Pretty sure there is a master branch but not a master branch minus one.
That is origin/master^ doesn't point to anything until there are 2 commits.
Removing this test broke it because there wasn't enough history yet.

jacob-keller:
What are your thoughts on this series now? I believe i outlined the particular failure above pretty well. A fix would be to include more "commit/push" in the BeforeTest, or fix the push force test by doing a commit --amend instead of a checkout rewind.

MarkEWaite:
I like the path it is running.  I merged your changes into a private branch on my fork and ran it through my multi-platform environment.  I also pushed two more changes over your proposed 3 changes.  Could you look at [those changes](https://github.com/MarkEWaite/git-client-plugin/commits/master-jk-submodule-tests) and see if they seem sensible to you?

If they are sensible to you, then I'll merge it to master.

If you're interested in more submodule tests, you might look at my [add-submodule-test branch](https://github.com/MarkEWaite/git-client-plugin/commits/master-add-submodule-test) or my [alternate add submodule test branch](https://github.com/MarkEWaite/git-client-plugin/commits/master-add-submodule-test-a)

jacob-keller:
Ok i will look at those. I want to get more tests in place so I can work on
getting the credentials included.

I am working on an extension for the git plugin to add credentials to
complete the package.

PR #179, opened 2015-08-03
JENKINS-24728  Add ability to specify shallow clone depth

nastharl:
This is pre-work to make the base git-client plugin able to understand depth for shallow clones. There will be another pull request on the main git-plugin to take it from the UI.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -244,6 +244,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             public boolean shallow;
             public Integer timeout;
             public boolean tags = true;
+            public Integer depth = 1;
 
             public FetchCommand from(URIish remote, List<RefSpec> refspecs) {
                 this.url = remote;
@@ -271,6 +272,11 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             	return this;
             }
 
+            public FetchCommand depth(Integer depth) {
+                this.depth = depth;
+                return this;
+            }
+
             public void execute() throws GitException, InterruptedException {
                 listener.getLogger().println(
                         "Fetching upstream changes from " + url);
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -244,6 +244,7 @@ public FetchCommand fetch_() {
             public boolean shallow;
             public Integer timeout;
             public boolean tags = true;
+            public Integer depth = 1;
```
KostyaSha: why not primitive?

nastharl: Keeping with the style of other ints being passed in. No other reason.

KostyaSha: will your code work if it will be null? Do you want handle all nulls instead of using primitives?
If it not a designed use case i will prefer to skip tri-state fields.

nastharl: Upstream its always initialized to at least 1, so you would have to specify a negative number (which would also be invalid). If someone calls this code without that in mind and still sends a null through, an error will be thrown saying that null is not a valid depth (same as with negative numbers or 0). That error is thrown by the underlying git implementation and is stated clearly in the build log. If we want to disallow even saving the configuration with an invalid value, i'll definatly have to throw some more work in.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -271,6 +272,12 @@ public FetchCommand timeout(Integer timeout) {
             	return this;
             }
 
+            public FetchCommand depth(Integer depth) {
+                this.depth = depth;
+                return this;
+            }
+
+
```
KostyaSha: extra line

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
KostyaSha: This field will be null after xml deserialisation so you should prepare to nulls

nastharl: Yea i'll just handle nulls inside git-client just incase. In my testing i couldn't couldn't throw anything in that behaved unexpectedly, but you never know when someone decides to take it and run a creative way.

nastharl:
I'll squash these into 1 commit once all review comments are in.

nastharl:
Updated with everyones comments. I think its good to go?

PR #180, opened 2015-08-03
(WIP) git-client: teach submoduleUpdate how to pass credentials

jacob-keller:
Teach the submoduleUpdate call to perform submodule URL lookup, and call
"submodule update" separately for each submodule. Lookup the
credentials, and call it using launchCommandWithCredentials. This
enables submodules to correctly get the SSH or HTTP setup necessary such
that the Jenkins credentials will be passed into each submodule.

We can't just call "git submodule update" since it may be possible
(however unlikely!) that each submodule wants to use a separate
credential. Thus, perform lookup using each URL and run through a
forloop to actually update each submodule instead of depending on the
git implementation to do this for us.

JENKINS-20941 - Credentials and Submodules

NOTE: this depends on a change in the git-plugin to enable setting up default credentials or adding credentials for the submodule URL. I have submitted the pull request to git-plugin which enables advanced extensions for adding credentials to arbitrary URLs or as the default credentials.

It would be possible to just run "git submodule update" as before if we assume that "default" credentials are good enough, and supporting per-submodule credentials is not good. I believe per-module credentials should be supported which is why I went with the given implementation.

This patch together with the git-plugin patch do work in my setup, but I haven't yet had time to add more tests for this feature. It passes all of the current tests.

HTTP
--
This series also includes a fix for HTTP authentication that *should* also enable us to remove the clone = init+fetch setup. I believe we have enough tests for that particular change in behavior.

The HTTP patch removes use of the git-credential storage and instead just uses a GIT_ASKPASS script that will output username and password when given the suitable prompt. For stdin matching "Username" it outputs username, for stdin matching "Password" it outputs the password. This is much cleaner, and should even allow git-clone to work as expected without need to create credentials store.

I should note that the default test for the CSV repostiories found in auth-data requires that the repository contains a README.md which my repositories do not, resulting in test failures for no good reason. This should probably be fixed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -867,6 +867,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         return new SubmoduleUpdateCommand() {
             boolean recursive                      = false;
             boolean remoteTracking                 = false;
+            boolean parentCredentials              = false;
             String  ref                            = null;
             Map<String, String> submodBranch   = new HashMap<String, String>();
             public Integer timeout;
@@ -878,6 +879,11 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
             public SubmoduleUpdateCommand remoteTracking(boolean remoteTracking) {
                 this.remoteTracking = remoteTracking;
+                return this;
+            }
+
+            public SubmoduleUpdateCommand parentCredentials(boolean parentCredentials) {
+                this.parentCredentials = parentCredentials;
                 return this;
             }
 
@@ -959,6 +965,18 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                     }
 
                     StandardCredentials cred = credentials.get(urIish.toPrivateString());
+                    if (parentCredentials) {
+                        String parentUrl = getRemoteUrl(getDefaultRemote());
+                        URIish parentUri = null;
+                        try {
+                            parentUri = new URIish(parentUrl);
+                        } catch (URISyntaxException e) {
+                            listener.error("Invalid URI for " + parentUrl);
+                            throw new GitException("Invalid URI for " + parentUrl);
+                        }
+                        cred = credentials.get(parentUri.toPrivateString());
+
+                    }
                     if (cred == null) cred = defaultCredentials;
 
                     perModuleArgs.add(sUrl);
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -51,6 +51,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
+import java.util.regex.Matcher;
 
 
 /**
@@ -922,7 +923,47 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                         args.add("--reference", ref);
                 }
 
-                launchCommandIn(args, workspace, environment, timeout);
+
+                // We need to call submodule update for each configured
+                // submodule. Note that we can't reliably depend on the
+                // getSubmodules() since it is possible "HEAD" doesn't exist,
+                // and we don't really want to recursively find all possible
+                // submodules, just the ones for this super project. Thus,
+                // loop through the config output and parse it for configured
+                // modules.
+                String cfgOutput = null;
+                try {
+                    // We might fail if we have no modules, so catch this
+                    // exception and just return.
+                    cfgOutput = launchCommand("config", "--get-regexp", "^submodule");
+                } catch (GitException e) {
+                    listener.error("No submodules found.");
+                    return;
+                }
+
+                // Use a matcher to find each configured submodule name, and
+                // then run the submodule update command with the provided
+                // path.
+                Pattern pattern = Pattern.compile("^submodule\\.(.*)\\.url", Pattern.MULTILINE);
+                Matcher matcher = pattern.matcher(cfgOutput);
+                while (matcher.find()) {
+                    ArgumentListBuilder perModuleArgs = args.clone();
+                    String sUrl = matcher.group(1);
+
+                    URIish urIish = null;
+                    try {
+                        urIish = new URIish(getSubmoduleUrl(sUrl));
+                    } catch (URISyntaxException e) {
+                        listener.error("Invalid repository for " + sUrl);
+                        throw new GitException("Invalid repository for " + sUrl);
+                    }
+
+                    StandardCredentials cred = credentials.get(urIish.toPrivateString());
+                    if (cred == null) cred = defaultCredentials;
+
+                    perModuleArgs.add(sUrl);
+                    launchCommandWithCredentials(perModuleArgs, workspace, cred, urIish, timeout);
+                }
             }
         };
     }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1351,9 +1351,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         File key = null;
         File ssh = null;
         File pass = null;
-        File store = null;
+        File askpass = null;
         EnvVars env = environment;
-        boolean deleteWorkDir = false;
         try {
             if (credentials instanceof SSHUserPrivateKey) {
                 SSHUserPrivateKey sshUser = (SSHUserPrivateKey) credentials;
@@ -1371,36 +1370,24 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 env = new EnvVars(env);
                 env.put("GIT_SSH", ssh.getAbsolutePath());
                 env.put("SSH_ASKPASS", pass.getAbsolutePath());
+
+            } else if (credentials instanceof StandardUsernamePasswordCredentials) {
+                StandardUsernamePasswordCredentials userPass = (StandardUsernamePasswordCredentials) credentials;
+                listener.getLogger().println("using GIT_ASKPASS to set credentials " + userPass.getDescription());
+
+                if (launcher.isUnix()) {
+                    askpass = createUnixStandardAskpass(userPass);
+                } else {
+                    askpass = createWindowsStandardAskpass(userPass);
+                }
+
+                env = new EnvVars(env);
+                env.put("GIT_ASKPASS", askpass.getAbsolutePath());
+                // SSH binary does not recognize GIT_ASKPASS, so set SSH_ASKPASS also, in the case we have an ssh:// URL
+                env.put("SSH_ASKPASS", askpass.getAbsolutePath());
             }
 
             if ("http".equalsIgnoreCase(url.getScheme()) || "https".equalsIgnoreCase(url.getScheme())) {
-                if (credentials != null) {
-                    listener.getLogger().println("using .gitcredentials to set credentials");
-                    if (!isAtLeastVersion(1,7,9,0))
-                        listener.getLogger().println("[WARNING] Installed git version too old for credentials support");
-
-                    String urlWithCredentials = getGitCredentialsURL(url, credentials);
-                    store = createGitCredentialsStore(urlWithCredentials);
-
-                    // Create a temporary workspace directory in the event that no
-                    // workspace has been created.  Call git init to allow for
-                    // credentials to be stored here during execution for HTTP-based
-                    // form validation.
-                    // See https://issues.jenkins-ci.org/browse/JENKINS-21016
-                    if (workDir == null) {
-                        workDir = Util.createTempDir();
-                        deleteWorkDir = true;
-                        init_().workspace(workDir.getAbsolutePath()).execute();
-                    }
-
-                    String fileStore = launcher.isUnix() ? store.getAbsolutePath() : "\\\"" + store.getAbsolutePath() + "\\\"";
-                    if (credentials instanceof UsernameCredentials) {
-                            UsernameCredentials userCredentials = (UsernameCredentials) credentials;
-                            launchCommandIn(workDir, "config", "--local", "credential.username", userCredentials.getUsername());
-                    }
-                    launchCommandIn(workDir, "config", "--local", "credential.helper", "store --file=" + fileStore);
-                }
-
                 if (proxy != null) {
                     boolean shouldProxy = true;
                     for(Pattern p : proxy.getNoProxyHostPatterns()) {
@@ -1430,23 +1417,6 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 }
             }
 
-            List<String> aa = args.toList();
-            /* Git versions prior to 1.7.9 are not tested with the git
-             * client plugin, but are used in the community. Red Hat 6
-             * and Red Hat 5 both ship with versions prior to 1.7.9.
-             *
-             * This conditional attempts to avoid calling command line
-             * git with an argument not implemented until git
-             * 1.7.9. Git versions prior to 1.7.9 also do not support
-             * credentials, so there are other significant portions of
-             * the plugin which will not work with those older
-             * versions of git.
-             */
-            if (isAtLeastVersion(1, 7, 9, 0)) {
-                aa.add(0, "-c");
-                aa.add(1, "core.askpass=true");
-            }
-            args = new ArgumentListBuilder(aa.toArray(new String[0]));
             return launchCommandIn(args, workDir, env, timeout);
         } catch (IOException e) {
             throw new GitException("Failed to setup credentials", e);
@@ -1454,38 +1424,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             deleteTempFile(pass);
             deleteTempFile(key);
             deleteTempFile(ssh);
-            deleteTempFile(store);
-            if (store != null) {
-                try {
-                    launchCommandIn(workDir, "config", "--local", "--remove-section", "credential");
-                } catch (GitException e) {
-                    listener.getLogger().println("Could not remove the credential section from the git configuration");
-                }
-                if (deleteWorkDir) {
-                    try {
-                        Util.deleteContentsRecursive(workDir);
-                        FileUtils.deleteDirectory( workDir );
-                    } catch (IOException ioe) {
-                        listener.getLogger().println("Couldn't delete dir " + workDir.getAbsolutePath() + " : " + ioe);
-                    }
-                }
-            }
+            deleteTempFile(askpass);
         }
-    }
-
-    private File createGitCredentialsStore(String urlWithCredentials) throws IOException {
-        File store = File.createTempFile("git", ".credentials");
-        PrintWriter w = null;
-        try {
-            w = new PrintWriter(store);
-            w.print(urlWithCredentials);
-            w.flush();
-        } finally {
-            if (w != null) {
-                w.close();
-            }
-        }
-        return store;
     }
 
     private File createSshKeyFile(File key, SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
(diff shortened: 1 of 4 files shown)


MarkEWaite:
I've started the interactive testing to evaluate this.  I'm not yet willing to accept the change without automated tests, but will use the interactive testing to think about ways the test can be automated without exposing actual credentials in the test environment.

jacob-keller:
It turns out that we probably could just use the jenkins lookup for credentials, instead of storing our own set of items from in a dictionary.

Today, what we do is store credentials in a dictionary keyed by URL, which is why the git-plugin needs an update as well to provide a window for passing credentials.

I am not sure the best way to fix this. Should we just try credentials using the URIRequirementBuilder and just take the first one?

Should we change behavior of how all credentials are looked up to use the API from the CredentialsPlugin?

MarkEWaite:
I thought that credentials were already being found through the API provided by the credentials plugin.  Can you expand further what you mean by "Should we change behavior of how all credentials are looked up to use the API from the CredentialsPlugin?"

I'm not as strong with the credentials API as I'd like to be.  We may need help from @jglick to answer your specific question.

jacob-keller:
> I thought that credentials were already being found through the API provided by the credentials plugin. Can you expand further what you mean by "Should we change behavior of how all credentials are looked up to use the API from the CredentialsPlugin?"

> I'm not as strong with the credentials API as I'd like to be. We may need help from @jglick to answer your specific question.

So, git-client-plugin today exports two APIs, "addCredentials" and "setDefaultCredentials", the addCredentials puts credentials into a Map<> with the key as the URL , so it looks like "addCredentials(URL, Creds)

when we use credentials in the client plugin, we look at this map, and at the default, if there is nothing in the map matching our URLs.

Git plugin itself, sets up the Map and default slots by using the drop down menu, so when you add a remote, it has a drop down to select the credential you want to use. This drop down is filled by using the Credentials search API so that only credentials that match the URI we present are given. However, we don't tell git-client about this, and git-client doesn't set it up.

Since submodules do not already use credentials, I can have them use the credential lookup instead of consulting the original map we created.

However, we do *not* just use the credentials API, and we could eliminate the entire map and just do the lookup our selves, and take credentials that way. This would be better but has higher impact on overall project.

So, I can do that, but again, that requires more tests, more work, and is definitely less  surgical.

I think it is the right thing to do, but the only downside is that I am not sure how to select between multiple options if more than one credential matches the criteria? Maybe we just dont' care about that at all?

Basically, I think this is a preferred approach long term, but maybe not in the short term.

In addition, to get http credentials working for submodules, we need to change how we request credentials to use the ASKPASS method instead of the credentials store, since git submodule doesn't appear to use the credentials store when it does a clone.

We do get requested username and password, and we can report differently based on standard input, so at least on the Unix side I think things will work. I don't know how to implement that based on windows yet though.

jacob-keller:
Ok, I included a further patch to modify how we do HTTP authentcation as the current scheme does not work with submodule update, since git-submodule is a shell script that performs a clone, and does not pass the git authentication credentials store inwards to the submodule.

The fix I've provided is based on the wording and test of git's use of GIT_ASKPASS, and does appear to work for my system. It looks like there is already a pretty wide slew of tests in this particular area so I am not super concerned about it yet.

I am working on extending CredentialsTest to also perform submodule init, but I noticed that the current csv credentials tests all rely on a README.md file. The repositories I was going to test do not actually have a README.md file, so the tests do fail locally. It appears that the test work otherwise.

MarkEWaite:
You asked:

> However, I think we already have this issue inside of createUnixSshAskpass as well, so as no one has seriously complained.... what are your thoughts?

That's a good test that is missing from my current test environment. I need a password in a username / password pair which includes a double quote character. If it is already broken for an existing case, then I see no harm in using the same pattern here so that it is consistent.  When the fix is applied later, it can be applied to multiple locations in the same file as readily as a single location.

jacob-keller:
I am unsure if it will be safe.. I know that a variable expanded inside quotes containing a quote will be safe, but I am not sure about this text here.

For now, I think we can use what was already used, and if it is broken a test case can be provided to check that.

jacob-keller:
It also appears that we will need a different escape for windows (obviously).

Likely, % and ^ among others may cause problems.

I am trying to see if there is maybe a better way we could resolve this maybe using a temporary file? I am not sure...

I think the best thing would be some tests for this behavior.

jacob-keller:
I added a test that enables submodules during credentials checks. However, this test does fail on JGit, as JGit does not perform submodule credentials.

It appears to at least not cause failures for Git CLI version, in that the cli version will not throw any exceptions. I haven't yet figured out how to test that it did actually succeed.

jacob-keller:
Added some minimal escaping.

I also still haven't had JGit working consistently in my local tests, and I am not entirely sure of the minimal "try to clone submodules" is correct enough a test.

The tricky part is in not hard coding credentials, but the repos.csv is not setup for adding more values to it.

MarkEWaite:
I think the repos.csv file eventually needs to be switched to a more fluid format like YAML or JSON. A pure rows / columns layout is not the most flexible representation, especially when we have several different types of authentication (ssh passphrase, username/password).

jacob-keller:
I'd support YAML or JSON as the format, I think both of these could work. JSON maybe easier?

It would help for submodules as well, and we could provide fields for things such as a filename to check for instead of hard coding README.md

I re-wrote the description of the second patch and clarified all that got changed.

I'm hoping this version fixes the HTTP issue, it may also fix the issue with bitbucket/stash

jacob-keller:
Ok. I found one more issue with the repos.json "submodule" field, that it only worked for Clone test, and not Fetch test. I added one patch to fix that.

In addition, I discovered that we blindly reset GIT_ASKPASS when SSH_ASKPASS is not set. GIT_ASKPASS is used by Git, while SSH_ASKPASS is used by the SSH binary (and Git, if GIT_ASKPASS and the core.askpass fields are not set...)

Based on reading back on the git-log I was not able to determine the actual reasoning behind setting GIT_ASKPASS to /bin/echo, but I assume that it makes sense to prevent interactive credential requests, which would hang the slave builds as no one would be there to interactively enter input.

However, we were essentially resetting the GIT_ASKPASS value.

I haven't figured out why this was failing interactive tests, but I think it's because the CredentialTest only runs 3 URLs, instead of all the ones listed (by default) and this results in it picking tests and the ones I got all passed.

This has allowed me to successfully interactively test HTTP, and SSH urls. It should also allow SSH urls which are not key-based but the only SSH repositories I have access too require ssh keys and won't allow password.

I also think the repos.json test with submodules shows that submodules with credentials tests pass, though maybe you still want more coverage here?

MarkEWaite:
You can run all authentication tests with the command:

    mvn -DTEST_ALL_CREDENTIALS=true clean install

I'll need to review the history of the GIT_ASKPASS=/bin/echo setting. I'm reasonably confident it was added for carefully chosen reasons, but I will need to look further to assure that is the case.

jacob-keller:
Ok I'll run all the tests to see if any of mine fail. I believe the Jgit ssh ones do, due to "connection not authenticated", which I'm not at all sure what the issue is for that. I *think* it may be because I have an ssh agent running on my local system. (as per other discussions I found on this plugin) so I think that is a local issue.

Also, I did a git-log all the way back to start of git-client-plugin but couldn't find why it was introduced originally. I suspect we'd have to look at log of git-plugin from before the split.

My gut says it was added in order to make sure that Git didn't hang forever waiting for credentials interactively (since Jenkins wouldn't be running interactively and it would fail)

In any respect, this change here is to *use* GIT_ASKPASS so we shouldn't be overwriting it again later. Thus, I changed the behavior to only write it to =/bin/echo if we haven't set it. We definitely don't need it for SSH style urls, because we call out to SSH which does its own ask pass and relies on SSH_ASKPASS and doesn't read ASK_PASS at all.

MarkEWaite:
4 of my private credentials tests (https authenticated access to BitBucket) are passing on the master branch but fail on this pull request.  If you'll send your e-mail address to mark.earl.waite@gmail.com, I'll send you the section of my repos.json file which runs those tests.  I can't post it here, since it includes a  user name and a plaintext password.

jacob-keller:
I've pushed a new commit which also adds submodule credential support to the JGit implementation.

jacob-keller:
I just found the issue with Windows scripts. Git passes the value inside quotes, and I wasn't telling the batch script to remove the quotes. I fixed this by using %~1 instead of %1 which appears to be the correct way.

The script now works from Git BASH on Git for Windows 2.5.1, where as before it gave the same error you were seeing.

I believe this should resolve the errors you were seeing in the Windows version.

I also updated the Unix script to use /bin/sh and use case statements which are posix compatible.

jacob-keller:
I rebased this series again, I think the changes included here are solid, and while their usefullness depends on changes to git, I do think the changes are ok on their own.

I still haven't heard much about the credentials changes, and he suggested that he already had an idea of how to implement what he wants, and didn't like my initial request, so that has somewhat stalled for now..

MarkEWaite:
Thanks. I am still very much interested in this implementation.  I was puzzling over this further and wondering if we could get 80% (or more) of the final solution if we took your git-client changes and added a git-plugin checkbox "Use repo credentials" with submodules to the submodules "Additional Behaviour" in the git plugin.

That seems like it would avoid the complexity of adding URL specific credentials and the complexity of having default submodule credentials, without preventing those from being added in the future.  The git-plugin would pass the "Use repo credentials" to the git-client-plugin somehow, and the git-client-plugin would then know to apply the parent credentials to the submodules.

What do you think of that as a simplification of the user interface in the git-plugin?  

Do you think it would prevent, block, or otherwise hinder later credentials improvements?

jacob-keller:
> Thanks. I am still very much interested in this implementation. I was puzzling over this further and wondering if we could get 80% (or more) of the final solution if we took your git-client changes and added a git-plugin checkbox "Use repo credentials" with submodules to the submodules "Additional Behaviour" in the git plugin.

I think this makes very good sense. I suspect very few projects have a different set of credentials per submodules vs the original repository, as generally speaking you can't use one without the other.

> That seems like it would avoid the complexity of adding URL specific credentials and the complexity of having default submodule credentials, without preventing those from being added in the future. The git-plugin would pass the "Use repo credentials" to the git-client-plugin somehow, and the git-client-plugin would then know to apply the parent credentials to the submodules.

One method would be without change to this, simply have the git-plugin create credentials per submodule repo.

The other would be to instead just use the "default credentials", so the git-plugin sets up the repo url for submodule credentials.

> What do you think of that as a simplification of the user interface in the git-plugin?

I think this is vastly simpler for now, and a much nicer approach since probably >90% of submodules are accessible either without credentials or with the same credentials.

> Do you think it would prevent, block, or otherwise hinder later credentials improvements?

No, I think this would not prevent or block later credential improvements if we design it right.

I think the simplest way would be for git-client to expose a way to get submodule->URL mappings, which enables git-plugin to add credentials for each URL as copied from the original repo. I will look at how to code this up.

jacob-keller:
I've updated this to include a new API for enabling "use of parent credentials" which the git plugin can use. I also have a pull request to extend the git plugin.

Long term the current code will support use of specific URLs first and only then if that fails try the parent credentials.

It might be worth just making this default behavior? In this way, we wouldn't even need the API changes, nor the credentials changes in git-plugin at all?

Since I am guessing almost every submodule will use the same credentials (ssh etc) as the parent does..? 

What do you think about that? If we make it default that would mean one less thing to learn about the intricacy of how to use the plugins...

geerligs:
I am really looking forward to release of the "use of parent credentials" feature. Right now I have to supply credentials in .gitmodules when trying to checkout a project in Bitbucket Server. I am expecting this feature will solve this and use only credentials in Jenkins.

runningbird:
Same can't wait for this it will help out a ton.

Cherum:
I'm awaiting this merge as well, had to do some dirty fixes instead.

codybaxter:
@jacob-keller 
I pulled your code locally on my system and installed the plugin on my local Jenkins. I have a project that is stored in Stash that contains a submodule. When I go to build the project in Jenkins, I am still getting submodule issues. Is there something special that needs to be added to the submodule file in the parent project that Jenkins is building?

Thanks,
Cody

jacob-keller:
What issue are you seeing? Note that you also need a  equivalent pull
request in the git plugin proper which enables the feature in a checkbox in
advanced submodule behavior.

Regards
Jake

codybaxter:
@jacob-keller 
I got it working. For some reason the install of the plugin wasn't working locally and I had to manually install it. When I set it to use the parent creds it is updating the submodules correctly.

jacob-keller:
Hooray =D

codybaxter:
Any idea of when this will get merged or any areas that I can help with.

MarkEWaite:
That's good to hear.  I hope to push a git client plugin 1.20.0-beta1 to the experimental update center along with a git plugin 2.5.0-beta1.  It will likely be another day or two at least, but once they are in the experimental update center, then users will be able to switch their Jenkins to use http://updates.jenkins-ci.org/experimental/update-center.json as the update center and see Jacob's work.

codybaxter:
@jacob-keller 
Was there a reason that the flag for UseParentCreds isn't being handled like the Recursive and RemoteTracking flags?

jacob-keller:
What do you mean?

MarkEWaite:
The git client plugin 1.20.0-beta1 has been released.  The branch (poorly named, sorry about that) is [master-1.20.0-beta1](https://github.com/jenkinsci/git-client-plugin/commits/master-1.20.0-beta1).  After some period of testing, it would be merged to master and that badly named branch will be deleted.

codybaxter:
The Recursive and RemoteTracking flags are passed by something like this:
git.submoduleUpdate()
                .recursive(recursiveSubmodules)
                .remoteTracking(trackingSubmodules)

I am just wondering why the UseParentCred flag is being passed that way. Why is there a separate method to set this flag?

jacob-keller:
I can rework to do it this way. I didn't notice there was a better pattern.

codybaxter:
@MarkEWaite What is the easiest way to get 1.20 beta 1 installed on Jenkins boxes for testing.

MarkEWaite:
Download the [git client plugin 1.20.0-beta1](https://repo.jenkins-ci.org/jenkinsci/releases/org/jenkins-ci/plugins/git-client/1.20.0-beta1/git-client-1.20.0-beta1.hpi) and [git plugin 2.5.0-beta1](https://repo.jenkins-ci.org/jenkinsci/releases/org/jenkins-ci/plugins/git/2.5.0-beta1/git-2.5.0-beta1.hpi), then upload them through the "Manage Plugins" page, under the "Advanced" tab.

Alternately, you could use configure your test Jenkins instance to use the [experimental update center](http://jenkins-ci.org/content/experimental-plugins-update-center).

jacob-keller:
Just for the record please don't release this till I finish the suggested
change for the flag use parent submodules. I will have that done sometime
tomorrow.

Regards,
Jake

MarkEWaite:
No worry about me releasing this too soon.  This is a large enough change that I want at least a month of beta testing before it is visible to general users.

geerligs:
I observe that when checking the tickbox "Use credentials from default remote of parent repository" this gets unfortunately blanked after a page save.

Consequently the function is not working :-(
Any clue how to solve?

jacob-keller:
That's a weird problem

geerligs:
after checking the tickbox "use credentials" and two other options as well I find in the config.xml for the job:
-<extensions>
-<hudson.plugins.git.extensions.impl.SubmoduleOption>
<disableSubmodules>false</disableSubmodules>
<recursiveSubmodules>true</recursiveSubmodules>
<trackingSubmodules>true</trackingSubmodules>
<useParentCredentials>false</useParentCredentials>
<timeout>30</timeout>
</hudson.plugins.git.extensions.impl.SubmoduleOption>
<hudson.plugins.git.extensions.impl.WipeWorkspace/>
</extensions>

This shows that the option "use credentials" is not saved, while the other options are.
I tried to manually change the config.xml but this doesn't work :-(

It seems that apply or save page sets the "Use credentials" option to false.

geerligs:
Now with correct syntax for config.xml:
```
-<extensions>
-<hudson.plugins.git.extensions.impl.SubmoduleOption>
<disableSubmodules>false</disableSubmodules>
<recursiveSubmodules>true</recursiveSubmodules>
<trackingSubmodules>true</trackingSubmodules>
<useParentCredentials>false</useParentCredentials>
<timeout>30</timeout>
</hudson.plugins.git.extensions.impl.SubmoduleOption>
<hudson.plugins.git.extensions.impl.WipeWorkspace/>
</extensions>
```

codybaxter:
I am seeing that issue too. It looks like the useParentCred flag isn't being set or read from the config file. When I updated the config file, I am still having a build failure because it isn't using the parent cred.

MarkEWaite:
Sorry about that.  That means I should have done some minimal smoke testing before I produced the beta1.  Hopefully Jacob will have time to investigate it and see what changes are needed.  If he can't resolve it in the next few days, I'll try to find the solution.  Once a solution is found (and verified), I'll generate the beta2 builds for the git plugin and the git client plugin.

jacob-keller:
I believe I found the issue in the git-plugin, I had a typo where a variable used "useParentCreds" instead of the correct name. I've pushed an update which should resolve this issue now.

In addition, I now use the same format as recursive and tracking, so that's a plus.

jacob-keller:
Oops had a warning which I just fixed as well.

geerligs:
I'm looking forward to test the fixed beta when the build is available.

codybaxter:
Just got beta 2 and everything looks like it is working.

jacob-keller:
That's excellent news!

jacob-keller:
I'm now having the issue with it not keeping the setting checkboxed...
anyone able to help investigate this on the git-plugin?? I'm at a loss.

codybaxter:
It is saving the setting in the config file but doesn't appear to reflect that on the settings page.

geerligs:
When running beta 2 I observe:
- setting the checkbox "Use credentials from default remote of parent repository" and saving enables me to retrieve the submodules from Stash! This was not the case with beta1;
- Also the corresponding node in the config.xml is set to true.
- however when loading the configuration will not show a ticked checkbox!

So my conclusion is that the problem arises on User interface level, not on the functional level. Basically the same conclusion as Cody.

Keep up the good work: I'll be happy when this is solved.

jacob-keller:
> When running beta 2 I observe:

> * setting the checkbox "Use credentials from default remote of parent repository" and saving enables me to retrieve the submodules from Stash! This was not the case with beta1;
> * Also the corresponding node in the config.xml is set to true.
> * however when loading the configuration will not show a ticked checkbox!

> So my conclusion is that the problem arises on User interface level, not on the functional level. Basically the same conclusion as Cody.

> Keep up the good work: I'll be happy when this is solved.

Ok beta3 of the git-plugin should resolve this now. Please test again to make sure.

geerligs:
I have tested beta 3 and checkbox seems to be handled correctly now!

When checking the box I see for each submodule to be updated:
using GIT_ASKPASS to set credentials general user for Stash

Side note: since latest update of Bitbucket Server (formerly Stash) to 4.1 it seems not be required to supply credentials for submodules anymore.

martino-jelli:
Hi guys,
I am currently blocked by this issue, when do you think a new release with this fix will be available?

Thanks for your hard work!

MarkEWaite:
The change is high enough risk that it will be several months before a full
release.  You could use the pre release now and help check the behavior
with the current pre-release

codybaxter:
It looks like this change would only affect sub-modules and doesn't appear to change any thing outside of that area. Is there not a reason to release this, since submodules never worked before in most cases. If the change affected more areas, I could see the reason to what a few months but there appears to be a lot of people wanting this change.

MarkEWaite:
@codybaxter as far as I have seen, the change to support authenticated submodules also changes the way authentication is performed for many other use cases.  There are automated tests of those changes, and the automated tests are run on Debian 6, Debian 7, Debian 8, CentOS 6, CentOS 7, Windows 7 and Windows 10.  Even with that test coverage, I think that releasing a change to the authentication technique without an extended (and well attended) beta period seems too aggressive to me.

adi1133:
I'd like this PR merged into a stable release. When is 1.5 being released ?

kamranayub:
Chiming in to say git 2.5.0-beta5 and git-client 1.20.0-beta3 fix the submodule credential issue and don't appear to break anything else (yet).

MarkEWaite:
Thanks @kamranayub, that is good information.  I would appreciate (if your installation is Jenkins 1.625 or newer) if you could use git-client-plugin 2.0.0-beta1 and git-plugin 3.0.0-beta1.  Both have been released with Java 7 (no more Java 6 support), the latest JGit, and the recent changes from the master branch.

kamranayub:
@MarkEWaite sure, I'll try that!

PR #181, opened 2015-08-06
Change to fetch head rev with refs/tags

moznion:
`CliGitAPIImpl#getHeadRev(String url)` doesn't fetch refs/tags so [git-plugin](https://github.com/jenkinsci/git-plugin) can't handle any tags (ref. [https://github.com/jenkinsci/git-plugin/blob/af01132e56f767f7ee944a3b64b17aa3a0c7f198/src/main/java/hudson/plugins/git/GitSCM.java#L583](https://github.com/jenkinsci/git-plugin/blob/af01132e56f767f7ee944a3b64b17aa3a0c7f198/src/main/java/hudson/plugins/git/GitSCM.java#L583), [https://issues.jenkins-ci.org/browse/JENKINS-14917](https://issues.jenkins-ci.org/browse/JENKINS-14917)).

From the above, I changed `CliGitAPIImpl#getHeadRev(String url)` to fetch head revision with refs/tags by appending `-t` option to `git ls-remote`.

However I can't predict effect of this change. It should be another new method?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2333,6 +2333,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     public Map<String, ObjectId> getHeadRev(String url) throws GitException, InterruptedException {
         ArgumentListBuilder args = new ArgumentListBuilder("ls-remote");
         args.add("-h");
+        args.add("-t");
         args.add(url);
 
         StandardCredentials cred = credentials.get(url);


MarkEWaite:
Can you explain the use case you're hoping to satisfy by broadening the search to include tags?  

As far as I can tell, the git plugin is able to checkout tags, and build from tags, and refer to tags.  Is there a specific bug you're trying to fix?

moznion:
My use case is

- Use with SCM polling
- Specify refspec

- Specify tag by branch specifier


It looks doesn't work. In my suggestion, it caused by polling command with git (in this case: `git -c core.askpass=true ls-remote -h https://github.com/moznion/sandbox`) cannot obtain `refs/tags`.
(I checked this symptom on [docker](https://github.com/jenkinsci/docker))

Do I take mistakes?

MarkEWaite:
[JENKINS-9713](https://issues.jenkins-ci.org/browse/JENKINS-9713), [JENKINS-21952](https://issues.jenkins-ci.org/browse/JENKINS-21952), and [JENKINS-23299](https://issues.jenkins-ci.org/browse/JENKINS-23299) all describe cases where a Jenkins build runs with a tag and performs the checkout of the tag correctly.  I have job definitions for each of those bugs and have confirmed they are all able to build a tag of my choice.

Could you review those bug descriptions to see if one of them will allow you to do what you need, without requiring a code change to the plugin?

moznion:
Thank you for your suggestion but I still can't build with tags even if applied shown solutions.
Maybe [JENKINS-23299](https://issues.jenkins-ci.org/browse/JENKINS-23299) is similar to my problem. Is this issue really resolved?

I think SCM polling kicks [this method](https://github.com/jenkinsci/git-plugin/blob/af01132e56f767f7ee944a3b64b17aa3a0c7f198/src/main/java/hudson/plugins/git/GitSCM.java#L525), it calls [this](https://github.com/jenkinsci/git-plugin/blob/af01132e56f767f7ee944a3b64b17aa3a0c7f198/src/main/java/hudson/plugins/git/GitSCM.java#L550), and it calls [CliGitAPIImpl#getHeadRev(String url)](https://github.com/jenkinsci/git-plugin/blob/af01132e56f767f7ee944a3b64b17aa3a0c7f198/src/main/java/hudson/plugins/git/GitSCM.java#L583). If it can't fetch tag information by `CliGitAPIImpl#getHeadRev(String url)` then `heads` is empty, returns `NO_CHANGES` so task won't be run. And now looks like `CliGitAPIImpl#getHeadRev(String url)` doesn't fetch tag information...

FYI, my jenkins environment is following;
Jenkins version: 1.609.2
GIT Plugin version: 2.4.0
GIT Client Plugin version: 1.18.0

MarkEWaite:
Have you tried enabling "Use local workspace for polling" with the current versions of the plugins?

moznion:
Sorry, how can I enable "Use local workspace for polling"?

MarkEWaite:
It is an "Additional Behaviour" that you add to the job.
force-polling-using-workspace

MarkEWaite:
As far as I can tell, JENKINS-23299 is resolved.  My job which was created to verify JENKINS-23299 is enabled and does not build on every polling interval.

PR #182, opened 2015-09-04
implement repos.json to replace repos.csv

jacob-keller:
Obviously, a new repos.json file will be required as compared to the current repos.csv.

This format is much more manageable, and extendable. Possibly we want to change the whole way we handle items in it instead of returning an object and using the items to populate a constructor, we should maybe pass the entire thing as a Map. However, I didn't bother with that just yet.

I did add two new features that are optional to the list, submodule checkout support, and file testing, to not hardcode the README.md

Hopefully this is a solid addition that can be pulled without much fuss, as these changes will be helpful in testing the related issues with the submodule and HTTP changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,6 +87,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.googlecode.json-simple</groupId>
+      <artifactId>json-simple</artifactId>
+      <version>1.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.modules</groupId>
       <artifactId>sshd</artifactId>
       <version>1.6</version>
--- a/src/test/java/org/jenkinsci/plugins/gitclient/CredentialsTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/CredentialsTest.java
@@ -58,6 +58,7 @@ public class CredentialsTest {
     private final String username;
     private final String password;
     private final File privateKey;
+    private final String fileToCheck;
 
     private GitClient git;
     private File repo;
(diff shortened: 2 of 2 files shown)


jacob-keller:
If the test harness on Jenkins uses a repos.csv file, it will need to be updated to use a repos.json file, and the checks won't really indicate proper coverage of these changes.

MarkEWaite:
The defaultPrivateKey section of the test does not include the correct number of arguments to pass the test. Once I added the new arguments, the test (with no repos.json) passed on my Ubuntu machine. It is running through my other machines currently (with no repos.json).

The test harness on the Cloudbees server does not provide a repos.csv file.  My home test environment is the only location which I know provides a repos.csv file.

As part of evaluating the pull request, I'll convert my repos.csv file into a repos.json file.  Once the conversion is complete and verified on one computer, I'll push it to all the other test machines so that it is evaluated on Windows,  Ubuntu, CentOS, and Debian with various git versions.

jacob-keller:
Oh that is what I was missing. I couldn't figure out how that was happening. Ok, I can clean that up in my series and submit a forced-push update.

MarkEWaite:
No need for you to make that change.  There were also formatting changes needed in order to keep the formatting inside the file consistent.  I'm running a modified version of the pull request through tests now, then will rebase my changes onto yours into a single commit.

jacob-keller:
Feel free to squash things in if you like :)

MarkEWaite:
I also found that the checks which were skipping rows with incomplete data were causing tests which I'd intentionally defined to be skipped.  So long as the data includes a repository URL, it should be used for a test.  There are URL forms which embed the username and password into the URL, others which embed only the username.

jacob-keller:
Ok that makes sense I wasn't sure but the old way made sure to have enough
fields so I just added checks for all the fields.

Regards
Jake

MarkEWaite:
Rebased and merged into master branch

PR #183, opened 2015-09-07
Enabled setMessage to MergeCommand and its two implementations

albestia:
Enable the ability to customize merge message con MergeCommand

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
   </scm>
 
   <properties>
-    <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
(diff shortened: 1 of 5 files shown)


albestia:
Hi @MarkEWaite , sorry to bother you, but I haven't found a way to check a message on a certain commit.
May you help me on how to proceed with this test?
Of course, maybe the response is that we need to implement full message features to enable "Custom Merge Message" and test it correctly. What do you think?

albestia:
After checking showRevision implementation and git documentation, this is not a bad solution for acquiring revision message. Test implemented and commited.

MarkEWaite:
Rebased and merged into the master branch.

PR #184, opened 2015-09-16
[JENKINS-29631] method to get remote references by refspec patterns

ndeloof:


ndeloof:
@reviewbybees

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2419,6 +2419,13 @@ public ObjectId getHeadRev(String url, String branchSpec) throws GitException, I
     /** {@inheritDoc} */
     public Map<String, ObjectId> getRemoteReferences(String url, String pattern, boolean headsOnly, boolean tagsOnly)
             throws GitException, InterruptedException {
+        return getRemoteReferences(url, pattern != null ? Collections.singletonList(pattern) : Collections.EMPTY_LIST,
+                headsOnly, tagsOnly);
+    }
+
+    /** {@inheritDoc} */
```
KostyaSha: afair it not required to wrote this javadoc as it will be inherited by default

jglick: Right. The only case where `{@inheritDoc}` is actually necessary is if you want to write _additional_ content for an override, while still including the super content, for example if an override makes additional guarantees about implementation. Unusual in practice.

ndeloof: I've been asked on another code review to add it 
I personally don't care

stephenc: I argue that you should always add `/** {@inheritDoc} */` to record the fact that the contract has been retained as inherited. Otherwise with no comment present then you do not know if the contract has been changed or retained... inaction is not the way to record a documentation decision.

So 

* leave blank if you have not decided whether the inherited contract has been changed or not **yet**
* put `/** {@inheritDoc} */` to indicate that the contract is exactly as inherited
* put the new contract in a comment to indicate that the contract has been changed (narrowed is the preference but it may be widened.

KostyaSha: AFAIK default inheritance merges javadoc. If you changed argument or return value you should describe it and other unchanged javadocs will be merged.

KostyaSha:
```
Automatic Copying of Method Comments
The Javadoc tool has the ability to copy or "inherit" method comments in classes and interfaces under the following two circumstances. Constructors, fields and nested classes do not inherit doc comments.
Automatically inherit comment to fill in missing text - When a main description, or @return, @param or @throws tag is missing from a method comment, the Javadoc tool copies the corresponding main description or tag comment from the method it overrides or implements (if any), according to the algorithm below.
More specifically, when a @param tag for a particular parameter is missing, then the comment for that parameter is copied from the method further up the inheritance hierarchy. When a @throws tag for a particular exception is missing, the @throws tag is copied only if that exception is declared.

This behavior contrasts with version 1.3 and earlier, where the presence of any main description or tag would prevent all comments from being inherited.

Explicitly inherit comment with {@inheritDoc} tag - Insert the inline tag {@inheritDoc} in a method main description or @return, @param or @throws tag comment – the corresponding inherited main description or tag comment is copied into that spot.
The source file for the inherited method need only be on the path specified by -sourcepath for the doc comment to actually be available to copy. Neither the class nor its package needs to be passed in on the command line. This contrasts with 1.3.x and earlier releases, where the class had to be a documented class
```

> leave blank if you have not decided whether the inherited contract has been changed or not yet

This sounds like your personal addition.

KostyaSha: > leave blank if you have not decided whether the inherited contract has been changed or not yet

imho, in this case better write this explicitly "this method doesn't follow contract, undefined behaviour" rather than implicitly guess that this method is not end in implementation and you can't trust in it's behaviour.

stephenc: Look, I have my guidelines for the repositories where I am maintainer. Nicolas got push back from me for some changes he was proposing for one of "my" repositories and I am explaining **my** _preference_.

I prefer to require a concrete deliberate action on the behalf of a developer over an inaction.

I recognise that the tooling does "the mostly right thing". In fact no comment *should* be the same as `/** {@inheritDoc} */` which is fine, *but* adding the comment says "I have thought about this and spent my time typing this extra stuff to confirm my decision" while doing nothing does not convey any information.

Yes we could say "if you are not sure that the contract is the same then leave a `// TODO document` comment... That then requires a conscious effort to add... But if I forget to add it, we will never know that I didn't assess the documentation.

It is for this exact reason that **my preference in repositories that I can influence** is to use the doc comment to track methods where documentation has been considered. 

This repository is not one where I am a maintainer, so I am not dictating for this repository. I do believe my principle is logical and has merit, but each plugin maintainer is the ruler of their domain...

That Nicolas *and others* have trouble remembering which conventions hold sway where is the fault of plugin maintainers not defining a `CONTRIBUTING.md` file with their guidelines... So do not blame Nicolas for sticky bits of style, rather join the campaign for documenting style guidelines for plugins.

KostyaSha: Who knows, maybe you created this rule in ancient java times and there no other ways. Thanks for description, now i know source of this inheritDocs.

PR #186, opened 2015-09-22
prefer Node’s Launcher over a LocalLauncher

ndeloof:


ndeloof:
@reviewbybees

ndeloof:
@MarkEWaite yes indeed, IIUC this will also fix JENKINS-21342 as the Node Launcher will be decorated by BuildWrappers.

> Can you provide more details about what this resolves?
Creating a LocalLauncher don't give any other plugin an option to configure where and how git is supposed to run. A plugin to configure a custom path / executable / environment (in my case, a docker container) won't be able to adjust the command line used but CliGitAPIImpl

> Can you provide a test case that exercises the new constructor in the way that is most interesting to your use case?
I hardly can imagine how to setup such a test case, would need some way to get LocalLauncher to fail while a decorated Launcher would work ...

ydubreuil:
:bee:

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -136,6 +141,29 @@ public GitClient invoke(File f, VirtualChannel channel) throws IOException, Inte
     }
 
     /**
+     * Retrieve the {@link Launcher} to be used to run git cli.
+     */
+    private @CheckForNull Launcher getLauncher() {
```
jglick: BTW this trick will not work for Workflow builds.

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -144,5 +172,5 @@ public GitClient invoke(File f, VirtualChannel channel) throws IOException, Inte
      */
     public static final boolean USE_CLI = Boolean.valueOf(System.getProperty(Git.class.getName() + ".useCLI", "true"));
 
-    private static final long serialVersionUID = 1L;
+        private static final long serialVersionUID = 1L;
```
jglick: Gratuitous.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -149,25 +150,28 @@ private void getGitVersion() {
...
-     *
-     * @param gitExe a {@link java.lang.String} object.
-     * @param workspace a {@link java.io.File} object.
-     * @param listener a {@link hudson.model.TaskListener} object.
-     * @param environment a {@link hudson.EnvVars} object.
+     *  @param gitExe a {@link String} object.
+     * @param workspace a {@link File} object.
+     * @param listener a {@link TaskListener} object.
+     * @param environment a {@link EnvVars} object.
+     * @param launcher
```
jglick: Thoroughly useless Javadoc comment, IMO better to delete it…

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -149,25 +150,28 @@ private void getGitVersion() {
...
-        launcher = new LocalLauncher(IGitAPI.verbose?listener:TaskListener.NULL);
+        this.launcher = launcher != null ?
+                launcher :
+                new LocalLauncher(IGitAPI.verbose?listener:TaskListener.NULL);
     }
 
     /** {@inheritDoc} */
     public GitClient subGit(String subdir) {
-        return new CliGitAPIImpl(gitExe, new File(workspace, subdir), listener, environment);
+        return new CliGitAPIImpl(gitExe, new File(workspace, subdir), listener, environment, null);
```
jglick: Should you not be passing `launcher`?

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
ndeloof: @jglick is there any way to retrieve the Node's Launcher in a workflow context ?

jglick:
:bug: no test case, no clear description of the user-visible problem being solved.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: good catch

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
jglick: If `repository` is a `FilePath`, then yes that would be the right approach to get a default `Launcher` (missing any decorations etc. that might have been passed in).

ndeloof:
@jglick 
"description of the user-visible problem being solved" 
Not sure how to explain better. Maybe some _how to reproduce_ : run docker-slaves, configure a container that has no `git` and setup a git repository. Expected to fail as you configured environment to _not_ have git, but actually work

"no test case"
As I already said I don't know what I could use as a test case to demonstrate this

jglick:
> I don't know what I could use as a test case to demonstrate this

Add a `@TestExtension` `BuildWrapper` whose `decorateLauncher` creates a `DecoratedLauncher` which does something special, like using `getListener()` to print a special message. Create a job using that wrapper and `GitSCM`, build it, and verify that the special behavior occurs with your patch (and does _not_ occur otherwise).

jglick:
Or if there are functional tests in `docker-slaves` which could prove the fix in a downstream PR, that would be better than nothing.

ndeloof:
I can't create a test with a reference to `GitSCM` as this one is from git-plugin and the dependency is in reverse order.

jglick:
> I can't create a test with a reference to `GitSCM`

True, but you might be able to create an integration test in `git-plugin` which would still catch regressions (especially if we routinely ran tests with dependencies updated to snapshots).

jglick:
re:bug:

ndeloof:
@jglick do you mean there's another issue, or you consider lack of test a blocker - which I disagree, but that's another debate.

MarkEWaite:
Does the remote launcher test in [this 2014 pull request](https://github.com/R-Gerard/git-client-plugin/commit/2e5bd1a74e27dcd7a4a6dbc9a6d90f1ec5712309) help with the idea of writing a test?

ndeloof:
his one was adding a Launcher parameter so was easier to test, my PR tries to not require an API change. But maybe I'm wrong, and we should introduce `Git#withLauncher(Launcher)` and amend git-plugin accordingly.

MarkEWaite:
I'm finding unexpected test failures on one machine (my primary development machine) that I think might somehow be related to the dependency update from Jenkins core 1.532 to Jenkins core 1.580.  I'll need to gather more data on the causes of those test failures. The failures seem to be specific to that single Ubuntu 14.04 machine but don't fail the same way on two other (slower) Ubuntu 14.04 machines, nor do they fail on Debian 8 or CentOS or Windows.

ndeloof:
I can't get git-client plugin to build without failure on my MacBook, can't tell why as the CI seems happy with it.

MarkEWaite:
Are the test failures you're seeing on your MacBook in the following tests:

* org.jenkinsci.plugins.gitclient.JGitAPIImplTest.test_clone_reference_working_repo
* org.jenkinsci.plugins.gitclient.JGitAPIImplTest.test_clone_shared
* org.jenkinsci.plugins.gitclient.JGitAPIImplTest.test_getHeadRev_returns_accurate_SHA1_values
* org.jenkinsci.plugins.gitclient.JGitAPIImplTest.test_getSubmoduleUrl

Those are all JGit tests, and the same tests from command line git run successfully for me.  I will likely try varying the Jenkins version dependency to see if I can better understand what might be causing the failures.

ndeloof:
No, I get another variant :)

Tests in error: 
  CliGitAPIImplTest.runTest:30->GitAPITestCase.test_clean:734 » Git Cannot commi...
  JGitAPIImplTest.runTest:36->GitAPITestCase.test_clean:761 » JGitInternal Could...

Tests run: 574, Failures: 0, Errors: 2, Skipped: 0

jglick:
> you consider lack of test a blocker

Yes, lack of a test for a change to the way that the most popular SCM integration for Jenkins launches that SCM is a blocker. :-)

jglick:
This needs a downstream PR to `git-plugin` showing that the `Launcher` which is passed to `GitSCM.checkout` is honored.

ndeloof:
@jglick here it is : https://github.com/jenkinsci/git-plugin/pull/419

ndeloof:
@MarkEWaite I just noticed this introduce a regression with credentials support when used on a remote slave, investigating.

ndeloof:
confirmed major regression. git-client is designed to run remotely, so just can't use a RemoteLauncher

jglick:
Indeed `Launcher` is not `Serializable` and is not expected to be sent over the remoting channel; the intention is that all `Launcher`-related logic is done inside the Jenkins master. `git-client` perhaps needs to be refactored to perform logic locally, and use `Launcher`/`FilePath` remoting calls for particular steps that need to happen on an agent.

KostyaSha:
@jglick obviously git 3.0

ndeloof:
yes indeed. Impact is huge and would not be feasible in a minor release.

PR #188, opened 2015-09-28
Create license file - MIT licensing

MarkEWaite:
Request from others to include explicit licensing in plugin.

Since this was derived from the git plugin, and the git plugin is 
MIT licensed, it seems reasonable to license the client plugin
with MIT licensing as well.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,7 @@
+Copyright (c) 2013-2015
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


MarkEWaite:
Pull request is not necessary.  The licensing info is already clearly stated in the pom.xml.  A separate LICENSE.md only duplicates the information already in the pom file

KostyaSha:
LICENSE md usually contains full description of license, while pom.xml indicates only license name.

LICENSE.md
```diff
@@ -0,0 +1,7 @@
+Copyright (c) 2013-2015
```
KostyaSha: I think it started more earlier

KostyaSha: Header not the same as in https://opensource.org/licenses/MIT

KostyaSha:
This file will be also right for https://help.github.com/articles/open-source-licensing/
Info from pom.xml probably is used in Netbeans for generating headers.
Imho it not a duplication and additional file makes sense.

LICENSE.md
MarkEWaite: The git log shows that the first commit to this repository was 2013.  Prior to that, this functionality was part of the git plugin.  I didn't attempt to trace the history of any source files back into the git plugin tree.

I assumes the license file at the root level is intended to represent this plugin and its repository.  However, I'm sure there are better experts than me to decide which copyright date should be applied based on the history of the files (if that is better than history of the repository)

KostyaSha: Wasn't it imported from svn?

MarkEWaite: I wasn't sure who to place as copyright owner.  Nicolas is the first committer, Kohsuke has committed many times, and I've committed many times as well.

Any pointers to rules and regulations which define the copyright owner?  Is it the original author?

KostyaSha: Ah, i forgot that it appeared after git plugin split. 2013 seems right.

KostyaSha: You can just wrote "Many open-source contributors" Offended persons can add their real name and dates.

PR #191, opened 2015-11-06
Fixed garbled commit messages on Windows

gennady:
On windows changelog commit messages with unicode characters are
not saved correctly to changelog.xml when CliGitAPI
implementation is in use.

That happens because "git whatchanged" gives byte stream of data.
Commit messages in that stream are encoded in UTF-8.  It is
necessary to explicitly decode bytestream to strings using UTF-8
encoding, otherwise default system encoding will be used.

This should fix issues:
  https://issues.jenkins-ci.org/browse/JENKINS-6203
  https://issues.jenkins-ci.org/browse/JENKINS-14798
  https://issues.jenkins-ci.org/browse/JENKINS-23091


engycz:
This change fixes my problem with invalid characters in commit message when using Czech locale.

engycz:
Please merge this pull request. It solved https://issues.jenkins-ci.org/browse/JENKINS-23091

gennady:
Canceled in favor of https://github.com/jenkinsci/git-client-plugin/pull/194

PR #193, opened 2015-11-21
Fix JENKINS-30371 allow symlinks with jgit

MarkEWaite:
Including one more JGit jar from the JGit 3.7.1 package allows JGit to support symbolic links in some cases.

Users running Java 6 will not have symlink support because the JGit library won't be loaded.

Users running Windows with default settings will not have symlink support because Windows does not allow symbolic links by default.

Users running Linux or FreeBSD or HP-UX or AIX or other Unix-like environments should have symbolic link support so long as they are running Java 7 or newer.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -2044,13 +2044,20 @@ public abstract class GitAPITestCase extends TestCase {
         assertFalse(w.git.hasGitModules());
     }
 
+    private boolean isJava6() {
+        if (System.getProperty("java.version").startsWith("1.6")) {
+            return true;
+        }
+        return false;
+    }
+
     /**
      * core.symlinks is set to false by msysgit on Windows and by JGit
      * 3.3.0 on all platforms.  It is not set on Linux.  Refer to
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -2044,13 +2044,20 @@ public void test_hasSubmodules() throws Exception {
         assertFalse(w.git.hasGitModules());
     }
 
+    private boolean isJava6() {
+        if (System.getProperty("java.version").startsWith("1.6")) {
```
KostyaSha: 1.66 == 1.6?

MarkEWaite: Yes, as is 1.6.0_0 and 1.6.0_14 and 1.6.0_11 and 1.6.0_15 and 1.6.0_05 and 1.6.0_45.  The question is trying to answer "is this Java 6 (and not Java 7 or Java 8)", not "is this a specific version of Java 6".

Do you see a problem with that technique?

Is there a better technique for deciding that the currently executing Java runtime environment does not have support for symbolic links?

KostyaSha: `startsWith("1.6.")` looks better

PR #194, opened 2015-12-15
 Fixed garbled commit messages on Windows

gennady:
On windows changelog commit messages with unicode characters are
not saved correctly to changelog.xml when CliGitAPI
implementation is in use.

That happens because "git whatchanged" gives byte stream of data.
Commit messages in that stream are encoded in UTF-8. It is
necessary to explicitly decode bytestream to strings using UTF-8
encoding, otherwise default system encoding will be used.

This should fix issues:
https://issues.jenkins-ci.org/browse/JENKINS-6203
https://issues.jenkins-ci.org/browse/JENKINS-14798
https://issues.jenkins-ci.org/browse/JENKINS-23091


mtschneiders:
I'm looking forward to this change as I'm also having issues with encoding on the changelog file. Is there a way to test the plugin with these changes while the pull request isn't merged?

engycz:
Here is the plugin compiled for tests
https://github.com/gennady/git-client-plugin/raw/8383bd7c222b52e26b0d1b395b2eb26766f86cf7/compiled-plugin/git-client.hpi

mtschneiders:
Success.
Tested it on my environment and could not reproduce the encoding issue I was having.

PR #195, opened 2015-12-21
[JENKINS-32160] Set StrictHostKeyChecking=no even if no credentials are provided

Vlatombe:
https://issues.jenkins-ci.org/browse/JENKINS-32160

This solves a confusing situation where the user configures a job with a
ssh repository, but doesn't provide ssh credentials.

In this case, the plugin doesn't provide StrictHostKeyChecking=no to the
GIT_SSH configuration, and the build likely fails with "Host key verification
failed." unless the build slave has been pre-configured with the host key.

Then, the user tries to fix this known hosts issue, rather than providing
a ssh key which is the real fix.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1340,31 +1340,35 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                                                 Integer timeout) throws GitException, InterruptedException {
 
         File key = null;
+        String user = null;
         File ssh = null;
         File pass = null;
         File store = null;
         EnvVars env = environment;
         boolean deleteWorkDir = false;
         try {
+            env = new EnvVars(env);
             if (credentials instanceof SSHUserPrivateKey) {
                 SSHUserPrivateKey sshUser = (SSHUserPrivateKey) credentials;
                 listener.getLogger().println("using GIT_SSH to set credentials " + sshUser.getDescription());
-
                 key = createSshKeyFile(key, sshUser);
+                user = sshUser.getUsername();
                 if (launcher.isUnix()) {
-                    ssh =  createUnixGitSSH(key, sshUser.getUsername());
                     pass =  createUnixSshAskpass(sshUser);
                 } else {
-                    ssh =  createWindowsGitSSH(key, sshUser.getUsername());
                     pass =  createWindowsSshAskpass(sshUser);
                 }
-
-                env = new EnvVars(env);
-                env.put("GIT_SSH", ssh.getAbsolutePath());
                 env.put("SSH_ASKPASS", pass.getAbsolutePath());
             }
 
-            if ("http".equalsIgnoreCase(url.getScheme()) || "https".equalsIgnoreCase(url.getScheme())) {
+            if ("".equals(url.getScheme()) || "ssh".equals(url.getScheme())) {
+                if (launcher.isUnix()) {
+                    ssh = createUnixGitSSH(key, user);
+                } else {
+                    ssh = createWindowsGitSSH(key, user);
+                }
+                env.put("GIT_SSH", ssh.getAbsolutePath());
+            } else if ("http".equalsIgnoreCase(url.getScheme()) || "https".equalsIgnoreCase(url.getScheme())) {
                 if (credentials != null) {
                     listener.getLogger().println("using .gitcredentials to set credentials");
                     if (!isAtLeastVersion(1,7,9,0))
@@ -1631,7 +1635,15 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         try {
             w = new PrintWriter(ssh);
             w.println("@echo off");
-            w.println("\"" + sshexe.getAbsolutePath() + "\" -i \"" + key.getAbsolutePath() +"\" -l \"" + user + "\" -o StrictHostKeyChecking=no %* ");
+            w.print("\"" + sshexe.getAbsolutePath() + "\"");
+            if (key != null) {
+                w.print(" -i \"" + key.getAbsolutePath() + "\"");
+            }
+            if (user != null) {
+                w.print(" -l \"" + user + "\"");
+            }
+            w.print(" -o StrictHostKeyChecking=no");
+            w.println(" %* ");
             w.flush();
         } finally {
             if (w != null) {
@@ -1651,7 +1663,15 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         w.println("  DISPLAY=:123.456");
         w.println("  export DISPLAY");
         w.println("fi");
-        w.println("ssh -i \"" + key.getAbsolutePath() + "\" -l \"" + user + "\" -o StrictHostKeyChecking=no \"$@\"");
+        w.print("ssh");
+        if (key != null) {
+            w.print(" -i \"" + key.getAbsolutePath() + "\"");
+        }
+        if (user != null) {
+            w.print(" -l \"" + user + "\"");
+        }
+        w.print(" -o StrictHostKeyChecking=no");
+        w.println(" \"$@\" ");
         w.close();
         ssh.setExecutable(true);
         return ssh;
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1340,28 +1340,27 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                                                 Integer timeout) throws GitException, InterruptedException {
 
         File key = null;
-        String user = null;
         File ssh = null;
         File pass = null;
         File store = null;
         EnvVars env = environment;
         boolean deleteWorkDir = false;
         try {
-            env = new EnvVars(env);
-            if (credentials instanceof SSHUserPrivateKey) {
-                SSHUserPrivateKey sshUser = (SSHUserPrivateKey) credentials;
-                listener.getLogger().println("using GIT_SSH to set credentials " + sshUser.getDescription());
-                key = createSshKeyFile(key, sshUser);
-                user = sshUser.getUsername();
-                if (launcher.isUnix()) {
-                    pass =  createUnixSshAskpass(sshUser);
-                } else {
-                    pass =  createWindowsSshAskpass(sshUser);
-                }
-                env.put("SSH_ASKPASS", pass.getAbsolutePath());
-            }
-
             if ("".equals(url.getScheme()) || "ssh".equals(url.getScheme())) {
+                String user = null;
+                env = new EnvVars(env);
+                if (credentials instanceof SSHUserPrivateKey) {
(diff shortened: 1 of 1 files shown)


Vlatombe:
@MarkEWaite I have updated the PR as we discussed.

jacob-keller:
This change makes sense to me considering default policy is "ask" which we can't use here, and using "yes" means any new slave machine must be pre-configured with each host. Although that may add security that would negatively impact the slave setup, and thus I agree we should use StrictHostKeyChecking=no.

jglick:
Do you not need to delete `-o StrictHostKeyChecking=no` from `createUnixGitSSH`? Is it not now redundant?

MarkEWaite:
The current pull request causes failures with several of the private CredentialsTest cases that I use.  Because those tests use my personal credentials and personal repositories (on bitbucket, github, gitlab, etc.), I will need to share them privately with Vincent, or will need to understand why they fail on my own.

I need some additional time to understand why it causes those failures.

jacob-keller:
> Do you not need to delete -o StrictHostKeyChecking=no from createUnixGitSSH? Is it not now redundant?

Probably ya, actually.

@MarkEWaite I am curious exactly why the tests would fail. The documentation for StrictHostChecking=no seems to indicate no specific reason. Possibly version of ssh issue? Or maybe there is a flaw in the addition of the string which causes the GIT_SSH to be misinterpreted...

Does it fail any of the non-private tests? Do any of those even use SSH? (doubtful I suspect)

MarkEWaite:
The non-private tests generally do not use ssh.  I was unwilling to disclose private keys in the public git repository.  The private keys are referenced in my ~/.ssh/auth-data/repos.json file.

I had to run the tests again to see the failures.  The specific Ubuntu 15 and Ubuntu 14 failures seem to be related to https authentication cases, not to ssh authentication.  Some of the failures are with the password embedded in the URL, others with the username and password provided in a credential. All the failing cases with this pull request are passing on the current master branch.

jacob-keller:
I know we fixed some issues a while back sounding like that, so maybe it is just not rebased onto the tip of master...

MarkEWaite:
Those tests pass on master, and master has been merged out to the branch where I'm evaluating this pull request.  I'm pretty sure they are real failures, but need more investigation to understand why.

jacob-keller:
Is it possible to share the output of your test sans any private information that would help more eyes try to spot the issue?

MarkEWaite:
The test results embed private information in the test name for easier debugging.  If you'll send me an e-mail (mark.earl.waite at gmail.com), I'll send you a Google drive link to the results.  I can also share the failing subset of my repos.json definition file (and others) if the test results are not enough.

jglick:
Sounds like someone (and I am not personally volunteering!) needs to create a suite of functional tests using Docker fixtures that would cover all of these special scenarios. Unless of course [existing acceptance tests](https://github.com/jenkinsci/acceptance-test-harness/blob/c5e0e5bfd8d7d53a48eb5574f8f7fc6d6c7d66c5/src/test/java/plugins/GitPluginTest.java#L74-L87) already suffice to reproduce the failure.

MarkEWaite:
I volunteer to create that set of functional tests using Docker fixtures.  I already have a private Docker repository which includes many functional tests that I use to confirm understanding of bugs, and to confirm that a bug has been fixed (and stays fixed).

That Docker repository currently includes my credentials.  That's why it is private.

KostyaSha:
@MarkEWaite for what do you need credentials?

jglick:
> That Docker repository currently includes my credentials.

Ah, so the hard work has already been done—just need to create some new keypairs/passwords for test purposes so the fixtures can become public.

BTW currently only `acceptance-test-harness` supports Docker fixtures directly. IIRC @lordofthejars had some suggestions for how to use Docker fixtures from `JenkinsRule` tests. Ideally `jenkins-test-harness` would get this functionality as a first-class feature, since I think a lot of plugins could benefit.

MarkEWaite:
@KostyaSha I have a Docker Jenkins instance which includes the 20+ plugins I use for my interesting bug verification jobs (named by the bug which they are checking), my credentials for github, bitbucket, assembla, gitlab, and more.  It includes 20+ authentication jobs that test various scenarios (github with username / password, github with private key, bitbucket with username / password, bitbucket with private key).  I'm not sure it fits what @jglick is describing, but it has been very helpful to accelerate my testing of new releases of the plugin and other components.

If you want to see the publicly visible variant of that (without my credentials), you can see it in the [master-with-plugins branch|https://github.com/MarkEWaite/docker/tree/master-with-plugins]

KostyaSha:
@jglick as I already said to you i'm using docker for spinning jenkins instances with spinned slaves. JenkinsRule internals can't be used without refactoring because they are hardcoded to localhost (and i also said it already). I wouldn't suggest ATH because it a spaghetti of classes and different dependencies like guice.

If master doesn't depend on environment then everything is easier and you can spin docker container with docker-java or if you want share git you may use github-api-plugin in test scope for creating test repositories (connect is pretty simple).

KostyaSha:
@MarkEWaite IT testing can be split in 2problems:
1. spinning jenkins master in special container
2. spinning test containers with some services.
3. spinning slaves with specific env (i.e. via docker-plugin or with static jnlp slave).

It depends whether services may be spinned locally. 
1) I.e. there is no docker containers for github so you can use only public instance. For this purpose i'm using separate gh account where i create git repositories and preparing them with jgit (testing PRs functionality). All you need is token that can be injected via env vars https://github.com/jenkinsci/github-integration-plugin/blob/master/github-pullrequest-plugin/src/test/java/org/jenkinsci/plugins/github_integration/its/AbstractPRTest.java#L65
Then you can a hack and extract github connection from github-plugin https://github.com/jenkinsci/github-integration-plugin/blob/master/github-pullrequest-plugin/src/test/java/org/jenkinsci/plugins/github_integration/its/AbstractPRTest.java#L82-L102

2) For gitlab i think you can spin container, configure gitlab and do real testing. (I'm using junit ExternalResource's for tests that prepare containers).

About 1. it depends on whether you really need create isolated container with single plugins installed because extracting dependency tree is unreally hacky. The most easier way is define all plugins as one list and then spin master that can be reconfigured with CLI (working code is in https://github.com/KostyaSha/yet-another-docker-plugin/tree/master/yet-another-docker-its ). 

Using latest dependencies may lead into unreproducible tests https://github.com/MarkEWaite/docker/blob/master-with-plugins/plugins.txt better define everythin in resources and pom.xml (if version shouldn't be parametrized) to have real life mapping between tests and spinned instances.

@MarkEWaite if you need any help for reproducing FOSS issues (not a CB problems that imho delayed and put into backlog most of FOSS reports), then i can help with implementation and wrote some docker based examples.

MarkEWaite:
> Using latest dependencies may lead into unreproducible tests https://github.com/MarkEWaite/docker/blob/master-with-plugins/plugins.txt better define everythin in resources and pom.xml (if version shouldn't be parametrized) to have real life mapping between tests and spinned instances.

I may be using the wrong approach, but because I commit the plugins into the [ref/plugins directory](https://github.com/MarkEWaite/docker/tree/master-with-plugins/ref/plugins) of my docker repository, the tests and the configuration is repeatable.

I freely admit that means that I am checking binary files into the repository, specifically so that I can have repeatable tests.  It has been helpful for my interactive testing.

KostyaSha:
It may work, but you are keeping test jenkins instance outside of tests. In my example i spin fresh jenkins instances (test prepares jenkins image based on official image with plugins defined in test scope automatically) and then i do required configuration/testing with or without additional services. In this way i have all tests logic near sources and easily spinning tests with debug breakpoints (slaves debug works fine with remote back connect java debugger). All tests logic is in code. 

You may run your prepared image from docker-hub, but it shouldn't contain any private info, you can inject them before running your master in docker in different ways. Prepared master instance should pick them. 
Documentation for this tests should describe what is required for running tests i.e.
- github needs only GH_TOKEN for admin user to do any GH related testing
- fresh docker instance i.e. `docker-machine create git-plugin`

Or you may even spin clean master and then inject/configure it with keys from your env (after it started). 
In any way you should unbundle any private keys into variables or env files (i.e. `~/.git-plugin-its.properties`). That would allow people reproduce runs and contribute tests.

jglick:
> my credentials for github, bitbucket, assembla, gitlab, and more

Ah, you mean you have a container with the _client_ side of the connection. Disregard what I said—I thought you had container(s) for the _server_ side.

> `JenkinsRule` internals can't be used without refactoring because they are hardcoded to localhost

Not if the test depends on a special container for the _master_. Which is possible, but in practice most tests do not care about that. In ATH, `DockerFixture` is only for external services (including Jenkins agents as a special case); while `JenkinsController` (the definition of how to run the master) _can_ run in a Docker container, this is not under the control of the test. So you can set up, for example, a working Git server with a hardcoded account list in a container, and test making connections to it. Obviously this is only practical for a fully automated test when the server software is open source (or at least freely downloadable); testing anything specific to GitHub etc. is harder.

jglick:
> I commit the plugins into the ref/plugins directory of my docker repository

Note that if you have a container which is `FROM jenkins:…` you can more easily [use `plugins.sh`](https://github.com/jenkinsci/docker#installing-more-tools).

KostyaSha:
@jglick with ATH you can't access code internals to check or configure something, that's why i implemented custom tests tooling. ATH only provides the spaghetti of classes for spinning predefined services or master and access both with web forms. This web based testing idea sucks and just ates development time. The other part of story is that many logic is in jenkins web bindings and it not possible to do code configuration for some things in core (i.e. see how Trigger tests are done in core (hint: everywhere with roundTrip activation)). 
(/me silently: 2.0, Jenkins as a code that can't be configured from the code - paradox.)

> Which is possible, but in practice most tests do not care about that.

That's why i said that it depends on what @MarkEWaite want achieve. As soon as git-plugin will require custom master env for different tests you will realise how ATH useless. And git-plugin depends on master env: ssh keys, pollings, git binaries, different workspaces for pollings, etc. 

Some docker junit libs also appeared, but they are not more than hundred of lines that can be easily written from scratch.

jglick:
> with ATH you can't access code internals

Sure, it is for black-box testing. Anyway a diatribe against ATH is off topic when I was not requesting an ATH test, merely noting that ATH currently includes a built-in Docker fixture system which can be emulated at the cost of a bit more verbosity in other harness.

> git-plugin depends on master env: ssh keys, pollings, git binaries, different workspaces for pollings

All of these are independent of the master’s environment except for non-workspace-based polling using the CLI Git tool, which could in principle have varying behavior depending on the version of Git in the path. In this case it seems unlikely because Jenkins only calls `git-ls-remote`, which is plumbing not porcelain and probably quite stable.

MarkEWaite:
I evaluated this pull request recently on my Ubuntu 14.04 machine and on a Debian 8 machine.  Both machines are configured to run ~30 different configurations through the CredentialsTest.  One of those tests fails consistently with these changes, but passes with the current master branch.  I don't currently have time to investigate the root cause of the failure, but am unwilling to merge this change without resolving the failure.  If Vincent or someone else is interested in the details of the failure (including how to remotely access the repository which shows the failure), I can provide that information by separate e-mail .  Request to mark dot earl dot waite at gmail

Vlatombe:
@MarkEWaite I just added a new patch which shouldn't affect the http/https flow. If you get time to have a new run on your test suite I'd be interested in the results.

stephenc:
@Vlatombe at the least I think you have merge conflicts that need fixing

PR #197, opened 2016-01-04
New cherry pick sensitive changelog (between branches) command

jan-zajic:
I implemented LogCommand (alternative to ChangelogCommand). Only cli impl! It calls git cli log command - https://git-scm.com/docs/git-log, with following pattern:
A..B if you don't set cherryPick parameter, or --cherry-pick --right-only A...B if you set it to true. 

It's very important for jenkins to use in our projects - it is only way to show correct changelog between our release branches, because we havily use cherry picking where preparing release candidates to build. Without this, jenkins show much false positives commits in build changelog. We currently very depends on this modified plugins in our jenkins installation, other plugins that build on top of changelog are important for us (for example jira plugin).

I prefer to add new command instead of modify existing one, not to broke current functionality and because i don't now how implement jgit alternative of that. If someone know (and need it), you welcome.

I also add CherryPickCommand - cli and jgit implementation, because i need to create unit tests that test LogCommand and thats only possible with support of CherryPickCommand. 

I also modify git plugin to exploit this feature, we used it in our env. I create pull request https://github.com/jenkinsci/git-plugin/pull/373 to git plugin, which can be merged only after this pull request to git client plugin (it depends on this new feature).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/AbstractLogCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/AbstractLogCommand.java
@@ -1,0 +1,44 @@
+package org.jenkinsci.plugins.gitclient;
+
+import java.io.Writer;
+
+public interface AbstractLogCommand<C extends AbstractLogCommand<?>> extends GitCommand {
+
+    /**
+     * Sets the {@link java.io.OutputStream} that receives the changelog.
+     *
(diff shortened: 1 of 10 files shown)


jan-zajic:
I fix formating in client plugin changes base on review comments.

jan-zajic:
@MarkEWaite please review this after fixes, thank you

PR #200, opened 2016-02-18
git-client: add support for --force option to submodule update

jacob-keller:
The purpose of this is to extend the CLI wrapper to allow --force option which enables recovery from some specific failures when the network has intermittent issues. Particularly, if the submodule update command happens to fail due to network issue, a re-run of the command will not re-checkout the files, as the way git CLI implemenation works it will see that the submodule is already at the desired commit. The Git mailing list said the best way to handle this is to use --force, but I don't think it should be default behavior considering that behavior change which can result.

Signed-off-by: Jacob Keller jacob.keller@gmail.com

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -908,6 +908,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         return new SubmoduleUpdateCommand() {
             boolean recursive                      = false;
             boolean remoteTracking                 = false;
+            boolean force                          = false;
             String  ref                            = null;
             Map<String, String> submodBranch   = new HashMap<String, String>();
             public Integer timeout;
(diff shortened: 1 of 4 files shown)


jacob-keller:
I submitted a new version that has a test case which covers the behavior I need and the expected behavior of --force. it will force a re-checkout even if you already have it checked out. Thus, it will recheckout any files which we deleted.

jacob-keller:
One alternative to this approach is to add "--force" by default, as this is what we already do inside the git plugin for regular checkouts. It seems weird that we wouldn't also add it for submodules as well.

I haven't yet tried this change, and I think it's riskier because it is more invasive than the addition of an option. That being said, if we already know that we want it generally why not also for submodules? Options are only good if they are infact useful being both on and off.. If everyone ends up having to check "yes" why should it be optional in the first place.

Thoughts?

jacob-keller:
Note it may be possible for the same effect to be had by using the clean option, which also (not entirely documented?) performs a git-reset as well.

jacob-keller:
I am testing, but I think the clean option may actually already do what  Ineed so this pull request ist not necessary. I will report back with my results from the clean plugin and see if it is workable instead.

jacob-keller:
I updated this pull request to include launchCommandIn support, and updated it to use cmdIn() instead, which will help ensure compatibility across platforms of Git.

I do believe this functionality is still important, as the clean() extension does not appear to fully fix the issue that --force can.

PR #201, opened 2016-03-24
[JENKINS-31464] Changed the logic that sets the proxy environmental variables.

scattym:
Now if there is already an environmental variable set (like in the node configuration or at the os layer on build user login) the git-client-plugin will not set it to be what is found in the "proxy settings" section of the jenkins configuration. This should mean that for the most part there would be no change in behaviour. The only difference this would make is if someone has gone to the trouble of setting these parameters at the node level (or at the o/s layer by setting the enviornment variables for the build user). If someone has gone to the trouble of setting the parameters in either of these two ways then it's likely (but not guaranteed) that they would had set these parameters correctly.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1424,7 +1424,6 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                     }
                     if(shouldProxy) {
                         env = new EnvVars(env);
-                        listener.getLogger().println("Setting http proxy: " + proxy.name + ":" + proxy.port);
                         String userInfo = null;
                         if (proxy.getUserName() != null) {
                             userInfo = proxy.getUserName();
@@ -1434,8 +1433,25 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                         }
                         try {
                             URI http_proxy = new URI("http", userInfo, proxy.name, proxy.port, null, null, null);
-                            env.put("http_proxy", http_proxy.toString());
-                            env.put("https_proxy", http_proxy.toString());
+                            // First check that the http proxy environmental variable is not already set.
+                            if( env.get("http_proxy", "") == "" ) {
+                                listener.getLogger().println("Setting http_proxy to be " + http_proxy.toString());
+                                env.put("http_proxy", http_proxy.toString());
+                            } else {
+                                listener.getLogger().println("Leaving http_proxy set to " + env.get("http_proxy", ""));
+                            }
+
+                            // First check that the https proxy environmental variable is not already set.
+                            if( env.get("https_proxy", "") == "" ) {
+                                listener.getLogger().println("Setting https_proxy to be " + http_proxy.toString());
+                                env.put("https_proxy", http_proxy.toString());
+                            } else {
+                                listener.getLogger().println("Leaving https_proxy set to " + env.get("https_proxy", ""));
+                            }
+
+                            // If we don't need the logging then the next two lines is equivalent to the above.
+                            //env.put("http_proxy", env.get("http_proxy", http_proxy.toString()));
+                            //env.put("https_proxy", env.get("https_proxy", http_proxy.toString()));
                         } catch (URISyntaxException ex) {
                             throw new GitException("Failed to create http proxy uri", ex);
                         }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1434,7 +1434,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                         try {
                             URI http_proxy = new URI("http", userInfo, proxy.name, proxy.port, null, null, null);
                             // First check that the http proxy environmental variable is not already set.
-                            if( env.get("http_proxy", "") == "" ) {
+                            if( env.get("http_proxy", "").equals("") ) {
                                 listener.getLogger().println("Setting http_proxy to be " + http_proxy.toString());
                                 env.put("http_proxy", http_proxy.toString());
                             } else {
@@ -1442,7 +1442,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                             }
 
                             // First check that the https proxy environmental variable is not already set.
-                            if( env.get("https_proxy", "") == "" ) {
+                            if( env.get("https_proxy", "").equals("") ) {
                                 listener.getLogger().println("Setting https_proxy to be " + http_proxy.toString());
                                 env.put("https_proxy", http_proxy.toString());
                             } else {


scattym:
Both solutions (the one in use and the one commented out in the code) work for me. My hosts are now able to access git repositories through each of the proxies in each of the different environments.

The only other way I thought to alter this behaviour was to add a check box somewhere that allows jenkins to default to the current behaviour. That way there would be zero chance this breaks anyones current install. I don't really understand how the plugins work yet, but I can look into it if people prefer that approach. I would just need someone to point me in the right direction as to how to do that... 

Also, I might need a helping hand as to how to implement unit testing for this feature. I've had a quick look but I'm outa my depth at the moment. Suggestions welcome.

tremble:
@MarkEWaite on https://issues.jenkins-ci.org/browse/JENKINS-31464 you said you'd consider pull requests.  I've just hit this issue as I have slaves in different locations which need different proxies.  Could you at least comment to let us know if this is an approach you're willing to consider?

MarkEWaite:
@tremble I'm willing to consider this as an optional behavior.

I'm not willing to change the default behavior, since that tends to break many users.  If this pull request changes the default behavior, it will need to be reworked to make the behavior optional.

I am unlikely to evaluate this pull request for several months.

You can help with the evaluation of the pull request by building it yourself, running it in your environment, and reporting the results of your evaluation.

tremble:
Sounds reasonable @MarkEWaite, thanks for the feedback

tremble:
@scattym Based on Mark's feedback it may make more sense to have an option that simply disables using the plugin-downloader proxy settings and includes a note about setting http_proxy/https_proxy in the node settings

ccschneidr:
Testing this with a node which has http_proxy (and https_proxy, no_proxy) set as an environment variable: The changes do not seem to have any effect.
From looking at the code of EnvVars it seems checking env.get("https_proxy", "") will always be "", because env is a new EnvVars and is not read from the node's environment but empty.

scattym:
@ccschneidr Hey Christian. The patch does work. It was the only way we were able to get running with internal slaves being located across a number of different networks with different proxies.

We have a bunch of failures on our build servers since November last year. I am not sure, but I would guess this could co-incide with an upgrade we did. Ok. Just checked. On that date we upgraded and that included an upgrade to git-plugin-client 2.1.0 and this did not have my patch. Hence the failures.

Why do you say EnvVars will always be ""? Because of the env = EnvVars(env) line? That's not part of my patch, that's from the original code.

I can show you a log entry from the day before the upgrade, that shows the patch working (notice the line that says "Leaving http proxy set to X below":

Started by timer
[EnvInject] - Loading node environment variables.
Building remotely on machine-vm2 (x86_64 vm xenserver rhel6) in workspace /var/builduser/workspace/edaproxy
 > git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from the remote Git repository
 > git config remote.origin.url https://bitbucket.org/someplace/someproject.git # timeout=10
Fetching upstream changes from https://bitbucket.org/someplace/someproject.git
 > git --version # timeout=10
using .gitcredentials to set credentials
 > git config --local credential.username ausername # timeout=10
 > git config --local credential.helper store --file=/tmp/git8508281067351007606.credentials # timeout=10
Leaving http_proxy set to http://192.168.1.1:3128
Leaving https_proxy set to http://192.168.1.1:3128

gabemontero:
Any updates on the fate of the PR?  I've registered a vote on the associated Jenkins issue, but thought I'd chime in here as well that my team/product would also benefit from having this capability to set on a per pipeline job run basis the proxy setting for the git plugin to use, independent of the plugin manager's proxy setting.

Thanks for your attention.

MarkEWaite:
@gabemotero thanks for your interest.  This is not on the top of my list, so it likely won't be reviewed for many months.  Those who are interested in the pull request can do a great service to me (and to the community) if they take a build with the pull request included, install it and use it in their environment.  That benefits the user (meets their need) and benefits me (and the community) by giving more test time on the proposed change.

With 100 000+ installations of the plugin, I'd love to have more help evaluating pull requests for safety, reliability, compatibility, and more.  One of the ways users can do that is by helping with the testing of pull requests.

gabemontero:
Thanks for the quick response @MarkEWaite !

I should be able to give this PR a go sometime next week.  I'll report back after I have some results, and we can go from there.

thanks again

scattym:
Hi Game/Mark,

This pull request has been reported not to work for pipelines. I haven't had the chance to investigate why as I am not using pipelines.

Regards,
Matt.

Get Outlook for Android<https://aka.ms/ghei36>

gabemontero:
I was able to confirm that a `http_proxy` env var set in a pipeline was not seen by @scattym 's changes.

I'm doing some debug to see if I can zero in on what is going on.  After some scanning of the git/git-client code I *might* have an inkling on what is going on, but can't say for sure.  

I'll report back when I have a useful update, or am just stuck.

gabemontero:
OK, my inkling was correct.  Env vars in pipelines are managed differently than freestyle / legacy job, and the [git plugin's](https://github.com/jenkinsci/git-plugin) initialization of the git-client provided by this plugin does not seed the git-client with env's set in the pipeline.

There are minimally 3 ways to set env vars in pipelines:
1) via `environment { ... }`
2) via `withEnv{ .... }`
3) via `env.http_proxy = '<value>'`

None of those env vars are accessible via `build.getEnvironment(listener)`, where the var defs are `Run<?, ?> build` and `TaskListener listener`.

Those 3 forms have different levels of visibility via the various workflow api's.  I was able to prototype a change in the git-plugin were we are able to get the env vars set via 3) (`env.http_proxy = '...'`) from the `Run<?,?>` object, merge them with the env vars currently obtained via `build.getEnvironment(listener)`, and then passed to the `createClient` call in GitSCM.java in the git-plugin.

@scattym 's change in this PR then properly picked up my pipeline's `env.http_proxy` and `env.https_proxy` settings.

Took at quick peek at PR's with the `env` label in the git-plugin, and saw nothing analogous to my prototype currently in the works.

So the net I think is:
1) this PR should not be gated by pipeline support; it satisfies the req for freestyle/legacy jobs.
2) I can submit a cleaned up version of my prototype to the git-plugin

@MarkEWaite - I believe you are (at least one of) the committer/admin for the git-plugin as well.  At least that is my inference based on the commit activity.  Would it be safe to assume that similar delays like you noted in https://github.com/jenkinsci/git-client-plugin/pull/201#issuecomment-309067798 are the reality for any PR I submit ?

MarkEWaite:
@gabemontero yes, you're safe to assume similar delays are the reality for any PR you submit.  I tend to give higher priority to PR's (like this one) where other users are evaluating the PR besides the original author.

gabemontero:
ok thanks for the confirm / guidance @MarkEWaite ... first things first, I'll see about getting the PR submitted, then point folks to it to give in a try and comment there as they ask my team about it

meiswjn:
Hi @MarkEWaite,
I would be interested in testing (and refining) this Pull Request. Should I open a completely new PR?

MarkEWaite:
> Hi @MarkEWaite, I would be interested in testing (and refining) this Pull Request. Should I open a completely new PR?

Yes, you'd need to open a new pull request.  The pull request needs to retain existing behavior by default and needs to have significant automated tests to confirm that the changes are well behaved.  I don't have an http proxy environment available, so am unable to test it myself.  Because I can't test it myself, I am quite hesitant to accept changes.

meiswjn:
Thanks for the response.

As the git client plugin respects the system-wide (--system, not --global) git config, our issue with agent-specific proxies is solved. This may also serve others that stumble upon this PR in the future.

PR #203, opened 2016-04-26
squid:S2039 - Member variable visibility should be specified

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2039 - Member variable visibility should be specified

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2039

Please let me know if you have any questions.

M-Ezzat

### DIFF
--- a/src/main/java/hudson/plugins/git/IndexEntry.java
+++ b/src/main/java/hudson/plugins/git/IndexEntry.java
@@ -10,7 +10,7 @@ import java.io.Serializable;
  * @author nigelmagnay
  */
 public class IndexEntry implements Serializable {
-    String mode, type, object, file;
+    private String mode, type, object, file;
 
     /**
      * Returns the mode of this entry as a String.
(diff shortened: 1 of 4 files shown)


m-ezzat:
I am not sure whether you do not agree with these changes or not. Please close this PR if not.

MarkEWaite:
I don't have specific detailed objections to the proposed change yet, and probably won't until I evaluate it.

Some of the objects you modified are serializable objects.  Jenkins serializes objects to disc using XStream, so I can't trivially accept the changes without performing detailed verification.

There is not enough gain for users from this change compared to the other changes I'm evaluating, so it is lower priority.

I suggested you consider implementing tests for one or more reported bugs from users, rather than performing cleanup based on Sonar coding rules.  You chose to ignore my advice, as is your right, but your choice does not change my view of priorities for the git plugin and the git client plugin.

You didn't mention the types of interactive testing you performed to confirm that your changes did not harm the code base.  For example, you could take my [lts-with-plugins docker instance](https://github.com/MarkEWaite/docker/tree/lts-with-plugins) and perform interactive tests with the bug checks included in that docker instance.  You could perform interactive tests of your own and describe the results of those interactive tests.

PR #205, opened 2016-05-24
git submodule sync should be recursive

bagage:
In case of submodule submodule, `git submodule sync` won't update their URL leading to "reference is not a tree" errors.


bagage:
Dummy commit to force Jenkins rebuild, it's passing now ;-).

bagage:
However `--recursive` for `git submodule sync` is only available since git `1.8.1`. Is that an issue?

MarkEWaite:
@bagage  I don't think it is an issue to depend on a capability which is only available since git 1.8.1.  There are commercially (and community) important Linux distributions which deliver versions older than 1.8.1, and the users on those older Linux distributions are accustomed to having less than full functionality from the git client plugin and the git plugin.

For example, Red Hat 6 and CentOS 6 deliver git 1.7.1.  As far as I can tell from the [Red Hat support site](https://access.redhat.com/support/policy/updates/errata), they will continue to support Red Hat Enterprise Linux 6 through 2020.

Users running the default git version on Red Hat 6 or CentOS 6 are not able to use authentication with their git repositories.  This would be another capability they could not use with their older operating system.

You would do them a great favor if you surrounded the `--recursive` argument with a conditional to check that their git version supported the argument.  Otherwise, users on those older Linux versions will be broken in their use of `submodule sync`.

Also, could you write a test which shows the bug you are fixing.  Code changes without a matching test addition make me very nervous.

bagage:
Thanks @MarkEWaite, I wondered if `git < 1.8.1` (released on 31th Dec. 2012!) was still used but this is indeed the case. I will do as you advice:

- add version checking
- add a unit test

I'll update this PR accordingly!

bagage:
@MarkEWaite I added a non regression test but actually I think that current implementation of submodules is non-functional eg I had to hack my test to make it work:
```
        // first time we must update submodule url
        w.igit().submoduleSync();
        w.igit().submoduleUpdate(false);
        // when submodule is up-to-date, we should sync its sub-sub-module
        w.igit().submoduleSync();
        w.igit().submoduleUpdate(true);
```
The problem arises because if you have the following:
> branchA: repository -> submodule (URL 1.1) -> subsubmodule (URL 1.2)
And then you switch branch:
> branchB: repository -> submodule (URL 2.1) -> subsubmodule (URL 2.2)
To properly checkout branchB, you have to do the following:
```
git checkout branchB
git submodule sync (update submodule to URL2.1 so that we can fetch it)
git submodule update (fetch submodule)
git submodule sync --recursive (update subsubmodule to URL 2.2 - cannot be done until submodule has been fetched)
git submodule update --recursive (finally update subsubmodule)
```
Obviously, if you have 3 levels of recursivity, you will have to do it once more, etc. so I believe that either:

* `submoduleUpdate` should do the `git sync` itself: would there be any issue to force syncing before updating?
* or alternatively we should run `git submodule sync && git submodule update` for each level of recursivity. 

What do you think?

In the mean time the test fails on `JGitAPIImpl` but I am not sure how to solve this.

MarkEWaite:
@bagage thanks for the effort writing a test. I believe you found the same challenges I found when attempting to write submodule automated tests.  As far as I can tell, there are many cases where the plugin does not handle transition from one state of a submodule enabled repository to a very different state of a submodule enabled repository.

My approach in those cases has been to write tests which show the current success cases, so there is at least some greater hope that future changes won't break the current cases that work.

I believe that I've typically marked submodule test cases with the NotImplementedInJGit annotation.  I believe JGit support submodules, but the submodule implementation in the git plugin expects more than the JGit implementation provides.

bagage:
@MarkEWaite thanks for the hint about `NotImplementedInJGit`, test is now passing! But I still believe my test hack is not the proper way to handle it. Have you any thoughts about this?

bagage:
Hi - simplified the issue to have at least one case when `sync` should be used with `--recursive` no matter what.

src/test/resources/specialBranchRepoCreate.sh
```diff
@@ -1,136 +1,136 @@
...
-git tag -a "master" -m "Tag test 'master'"
-
-# End
-git checkout master
-
-# Create files for src/test/resources/ and cleanup
-jar cvf ../specialBranchRepo.zip ./
-git ls-remote . > ../specialBranchRepo.ls-remote
-cd ..
+#!/bin/sh
```
MarkEWaite: Please remove src/test/resources/specialBranchRepoCreate.sh from the pull request.  It appears to only be a change of line termination character.

MarkEWaite:
See #405 for a similar request

MarkEWaite:
Closing for now, will reopen when submodule enhancements come to the top of the stack.  Marked with the `Later` milestone so that it can be found again.

PR #206, opened 2016-06-14
[JENKINS-35687] Add simple git lfs support

matthauck:


MarkEWaite:
I'd really like some form of automated tests to be included with the pull request.  Without automated tests, we then must rely on the plugin developer to interactively test this feature before every plugin release.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -312,6 +318,13 @@ public void execute() throws GitException, InterruptedException {
                 warnIfWindowsTemporaryDirNameHasSpaces();
 
                 launchCommandWithCredentials(args, workspace, cred, url, timeout);
+
+                if (withLFS) {
+                    ArgumentListBuilder lfsArgs = new ArgumentListBuilder();
+                    lfsArgs.add("lfs");
```
MarkEWaite: Wouldn't it be better to check that LFS is available in the git version being used, rather than just throw an exception when it fails?

matthauck: Since it is an explicit setting that one would turn on, it makes most sense to me to just fail if it isn't present, rather than just log it and continue the build (which will mostly likely fail in some not-well-defined way, when it, for example, tries to extract a zip file that is not a zip file since lfs).

If you would rather thIs do a check, is there a pattern in the git client plugin I can reference for checking for git feature support?

matthauck:
Wasn't quite sure how to add an automated test that did not require having git lfs installed and have something to fetch... Or maybe there is a mock cli that you have setup that can just check if commands were run?

PR #207, opened 2016-06-16
[JENKINS-20356] Fix SSH_ASKPASS on windows

matthauck:
The SSH_ASKPASS script was being setup improperly on windows since
it was echoing quotes as well as the actual echo command itself
since `@echo off` was not being specified.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1464,6 +1464,12 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 env.put("GIT_SSH", ssh.getAbsolutePath());
                 env.put("SSH_ASKPASS", pass.getAbsolutePath());
 
+                // supply a dummy value for DISPLAY if not already present
+                // or else ssh will not invoke SSH_ASKPASS
+                if (!env.containsKey("DISPLAY")) {
+                    env.put("DISPLAY", ":");
+                }
+
             } else if (credentials instanceof StandardUsernamePasswordCredentials) {
                 StandardUsernamePasswordCredentials userPass = (StandardUsernamePasswordCredentials) credentials;
                 listener.getLogger().println("using GIT_ASKPASS to set credentials " + userPass.getDescription());
@@ -1558,7 +1564,10 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     private File createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOException {
         File ssh = File.createTempFile("pass", ".bat");
         try (PrintWriter w = new PrintWriter(ssh, Charset.defaultCharset().toString())) {
-            w.println("echo \"" + quoteWindowsCredentials(Secret.toString(sshUser.getPassphrase())) + "\"");
+            // avoid echoing command as part of the password
+            w.println("@echo off");
+            // no need for quotes on windows echo -- they will get echoed too
+            w.println("echo " + Secret.toString(sshUser.getPassphrase()));
             w.flush();
         }
         ssh.setExecutable(true);


matthauck:
This solves https://issues.jenkins-ci.org/browse/JENKINS-20356 for me

matthauck:
Well, forgot, there was one more thing here. Need to also add the `DISPLAY` environment variable if not present or else ssh will not invoke the SSH_ASKPASS.

matthauck:
Weird. Not sure why this status got changed. I saw it build successfully once...

matthauck:
Where can we see the build output for this failure? The "Details" link takes me to `http://unconfigured-jenkins-location/job/github-jenkinsci/job/git-client-plugin/job/PR-207/1/`...

MarkEWaite:
Sorry @matthauck , the report "this commit cannot be built" is a result of my experiments with a Docker instance which I regularly start and stop in various configurations.  It is configured with the GitHub Organization Folders plugin and is (correctly) using my credentials to publish those results back to GitHub.

My apologies for distracting you with my experiments with the Jenkins plugin for GitHub Organization Folders.  I'll likely remove my prototype job definition so that it does not continue publishing these experimental attempts to build pull requests.

matthauck:
Ah ok, cool

matthauck:
Just checking whether this has any chance of getting in any time soon?

MarkEWaite:
Yes,  it has a chance.  I started my interactive treating of it last week by creasing a service based Jenkins nose,  then tried some tests.  Unfortunately,  the initial treats without the change did not fail. 

I also compared the proposed change to the code on the 2.0.0-beta branch. 

I need to do more testing,  particularly of cases where the secret might require quotes on windows

matthauck:
awesome! :balloon:

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1518,7 +1518,10 @@ private File createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOExcepti
         PrintWriter w = null;
         try {
             w = new PrintWriter(ssh);
-            w.println("echo \"" + Secret.toString(sshUser.getPassphrase()) + "\"");
+            // avoid echoing command as part of the password
+            w.println("@echo off");
+            // no need for quotes on windows echo -- they will get echoed too
+            w.println("echo " + Secret.toString(sshUser.getPassphrase()));
```
MarkEWaite: This code only seems to be invoked when a passphrase has been assigned to the private key.  I was hoping it was the RSA private key (rather than the passphrase for the key) because that would mean it was encoding user input rather than echoing the user input without any encoding.  My hopes were wrong.

Since the resulting string from Secret.toString(sshUser.getPassPhrase()) is not encoded, it may contain one or more of the [Windows command processor special characters](http://ss64.com/nt/syntax-cmd.html) '&', '<', '>'. '(',  ')', '@', '^', '|'.

If the user provided passphrase includes one or more of those special characters, then the surrounding double quotes are needed (at least per the "Quote Characters in a command" section of the [ss64 command syntax site](http://ss64.com/nt/syntax-cmd.html).

If the user provided passphrase does not include one or more of those special characters, then the surrounding double quotes are not needed (and probably must not be included).

The original bug report [JENKINS-20356](https://issues.jenkins-ci.org/browse/JENKINS-20356) made no mention of an ssh passphrase being required to see the bug, yet this change seems to only be executed if the user provided an ssh passphrase protected private key.  

Does your use case require a passphrase protected private key?

MarkEWaite: When I created a private key using a passphrase which intentionally includes command processor special characters (like pipe, caret, less than, greater than, at sign, etc.), then try to use that passphrase ( `<mark|earl^waite>@gmail.com`) with the code, the initial attempt to authenticate with that repository (from the web user interface), reports:

    Failed to connect to repository : Command "C:\Program Files\git\bin\git.exe -c core.askpass=true ls-remote -h mark-pc1.markwaite.net:/var/lib/git/hasphrase/bin.git HEAD" returned status code 128:
    stdout: 
    stderr: 'earlwaite' is not recognized as an internal or external command, 
    operable program or batch file. 
    'earlwaite' is not recognized as an internal or external command, 
    operable program or batch file. 
    Permission denied, please try again. 
    'earlwaite' is not recognized as an internal or external command, 
    operable program or batch file. 
    Permission denied, please try again. 
    'earlwaite' is not recognized as an internal or external command, 
    operable program or batch file. 
    Permission denied (publickey,password). 
    fatal: Could not read from remote repository.
    
    Please make sure you have the correct access rights
    and the repository exists.

matthauck: Eek. Definitely don't want to open up a code injection vulnerability. :cry:

And yes, the key we are using does indeed have a passphrase so that the private key cannot be grabbed from the jenkins UI. That way the private is not usable without gaining access to the jenkins server and grabbing the (obfuscated) passphrase. 

I wonder if it would be better to be changed to something like: ```cmd /C "echo foo"``` together with escaping quotes (or perhaps not allowing quotes?) in the passphrase itself.

MarkEWaite:
I don't think your change opened any more of an injection vulnerability than was already there.  If someone has permissions to register and use credentials, I believe they have sufficient permissions to do very serious damage, without needing to use passphrase hacks to do it.

MarkEWaite:
I think there may be a simpler way to solve the quoting problem than what I was considering.

Windows has the `type` command which will echo the contents of a file to stdout.  If we send the passphrase to a file, then use type to output the contents of that file, we should be able to avoid all of the string quoting rules.

The same technique should work for Unix variants as well.  The plugin writes the passphrase to a temporary file, then would use the `cat` command to output the contents of that file to stdout.

realrolfje:
Writing the passphrase unencrypted to a temporary file sounds like a terrible idea, security wise.

MarkEWaite:
@realrolfje, as far as I can tell, the choices are limited for communicating the passphrase to ssh. 

Today's implementation writes the unencrypted passphrase to a file as an argument the `echo` command.  If the operating system displays the arguments to `echo` in a process listing, then the unencrypted passphrase is visible during the time the echo is executing.  That also means the passphrase needs to be safely escaped on all platforms so that the echo command correctly outputs the passphrase, even in the cases where the passphrase contains interesting characters like caret (a [Windows command processor special character](http://ss64.com/nt/syntax-cmd.html)), double quote, single quote, ampersand, or interesting UTF-8 characters like [zero width space](https://www.cs.tut.fi/~jkorpela/chars/spaces.html).

My implementation writes the unencrypted passphrase to a file and uses `cat` (Unix) or `type` (Windows) to send the unencrypted passphrase to ssh.

Both implementations strive to delete the unencrypted passphrase from the file system as soon as it has been passed to command line git.

Git 2.0 and later offer a credentials manager which might be a viable replacement, but requiring git 2.0 as a minimum version excludes the default git shipped with many major operating systems (Red Hat 5, 6, and 7, CentOS 5, 6, and 7, Debian 7).

realrolfje:
I got here because I had this problem (Windows/Temp/ssh822323982key File not found with passphrase on private key) after upgrading the git plugin to 2.5.3. Just downgraded the git plugin to version 2.5.2 in my local Jenkins installation and the problem seems to be gone.

Regarding git: version 2.0 has been released over two years ago, and can be easily upgraded on all the Linux distributions you mentioned. To me it sounds reasonable to make a new version of the git plugin which depends on git 2.0 and up. But I might be biased.

MarkEWaite:
I'm not yet willing to force a command line git upgrade on the more than 80 000 installations of the git plugin.

matthauck:
Indeed. We do not all have the luxury of only building on modern OS's. 😞 

Also, with respect to the security of the private key's password, it is indeed not secure as it is. To test some of this stuff on windows, I had written a little .bat script that scans the TEMP directory and copies over the ssh/password files. So, yeah, exploitable -- assuming your build machines are not locked down. The solution we have taken is to ensure our build VMs are locked down in a secure / segregated network.

With respect to git credential API stuff, that would be interesting to look into to support / take advantage of _if present_, while keeping the current workflow for clients without the newer credential manager.

matthauck:
I believe this is still an issue on the latest version of git-client-plugin (v. 2.2.1). I have updated to apply against a later version of git-client-plugin to resolve the merge conflicts. 

As far as I can tell, encrypted ssh creds is basically broken on windows right now with the git-client-plugin. Any chance we can get in this fix for now rather than getting side-tracked with future potential enhancements?

matthauck:
Thanks! 🎉

PR #209, opened 2016-06-21
git.clean() is more clean

bukzor:
git clean needs *two* -f's to delete directories that have a .git

From `git help clean`:

> If an untracked directory is managed by a different Git repository, it is not removed by default. Use -f option twice if you really want to remove such a directory.


MarkEWaite:
Please refer to #163 for the detailed reasons why this pull request is being closed without merging.

PR #210, opened 2016-07-04
do not call echo from absolute path

gilligan:
`git-client-plugin` calls `echo` with an absolute path from `/bin/echo` on *-nix operating systems which causes an error on any system where `echo` may not be located there for whatever reason.

I am running jenkins on [NixOS](http://nixos.org) which does not adhere to [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) meaning `echo` is not located in `/bin/echo`. 

This PR changes the code such that echo is always called without absolute path so it is executed from `$PATH`.

 I can't think of any good reason why executing it from an absolute path should be favourable and with this change it will work fine if `echo` is located elsewhere.

This fixes [JENKINS-36255](https://issues.jenkins-ci.org/browse/JENKINS-36255)


KostyaSha:
👍

gilligan:
@MarkEWaite thanks a lot for merging this. Is there any chance of having a new release some time soon that would include this PR? That would really help us a lot.

PR #211, opened 2016-07-06
let CliGitAPIImpl rely on FilePath+Launcher remote API while JGitAPIIml is remotable.

ndeloof:
This (long) pull-request is about making CLIGitAPIImpl rely on `FilePath` and `Launcher` and have logic run on master JVM (as expected by the remoting abstraction). As JGitAPIImpl relies on JGit - so on java.io.File - it has to run fully remotely, so had to split logic between implementations.

Most changes are about moving from `File` to `FilePath` and adding `IOException` to API methods.

Also removed getRepository that was deprecated since 1.1 as not being remotable.

Final goal is to let git-client rely on configured Launcher, which hold the computer environment or customized behavior from caller (using `DecoratedLauncher`)
see https://issues.jenkins-ci.org/browse/JENKINS-30600, also applies to git ran inside Pipeline as computer's NodeProperties are exposed to git by Launcher.

### DIFF
--- a/src/main/java/hudson/plugins/git/GitAPI.java
+++ b/src/main/java/hudson/plugins/git/GitAPI.java
@@ -3,6 +3,7 @@ package hudson.plugins.git;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.EnvVars;
 import hudson.FilePath;
+import hudson.Launcher;
 import hudson.model.TaskListener;
 import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.lib.PersonIdent;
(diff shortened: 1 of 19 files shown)


ndeloof:
GitAPITestCase#test_checkout_interrupted works locally (at least "work on my machine").

ndeloof:
@reviewbybees

ndeloof:
As changes are significant, maybe this should target 2.x ?

MarkEWaite:
I agree that this is a better fit for 2.0.0-beta.

Since there's already a 2.0.0-beta branch on the git client, and it contains significant improvements (submodule authentication, JGit 4.3, JDK 7, try with resources) it would be best if you did the merge onto the 2.0.0-beta branch rather than having me attempt to perform the merge.

This also gives a chance for you (and others) to review the changes I made for git client plugin 2.0.0-beta1 (and for the earlier releases of git client plugin 1.20.0-beta1, beta2, beta3, and beta4).

src/main/java/hudson/plugins/git/GitAPI.java
```diff
@@ -62,19 +48,24 @@ public GitAPI(String gitExe, FilePath repository, TaskListener listener, EnvVars
         this(gitExe, repository, listener, environment);
     }
 
+    @Deprecated
+    public GitAPI(String gitExe, FilePath repository, TaskListener listener, EnvVars environment) throws IOException, InterruptedException {
+        this(gitExe, repository, listener, environment, (Launcher) null);
```
KostyaSha: why do you need cast null to launcher?

ndeloof:
PR on 2.0.0-beta branch: https://github.com/jenkinsci/git-client-plugin/pull/212

src/main/java/hudson/plugins/git/GitAPI.java
ndeloof: to disambiguate constructor call vs (deprecated) GitAPI(String, FilePath, TaskListener, EnvVars, String)

jglick:
Does this follow up #186?

ndeloof:
@jglick yes indeed, but doesn't seem to get enough attention so I gave up maintaining this PR

PR #212, opened 2016-07-06
let CliGitAPIImpl rely on FilePath+Launcher remote API

ndeloof:
This (long) pull-request is about making CLIGitAPIImpl rely on FilePath and Launcher and have logic run on master JVM (as expected by the remoting abstraction). As JGitAPIImpl relies on JGit - so on java.io.File - it has to run fully remotely, so had to split logic between implementations.

Most changes are about moving from File to FilePath and adding IOException to API methods.

Also removed getRepository that was deprecated since 1.1 as not being remotable.

Final goal is to let git-client rely on configured Launcher, which hold the computer environment or customized behavior from caller (using DecoratedLauncher)
see https://issues.jenkins-ci.org/browse/JENKINS-30600, also applies to git ran inside Pipeline as computer's NodeProperties are exposed to git by Launcher.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2323,13 +2323,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     @Override
     public <T> T withRepository(final RepositoryCallback<T> callable) throws IOException, InterruptedException {
-        return workspace.act(new MasterToSlaveFileCallable<T>() {
-            @Override
-            public T invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
-                final Repository repository = new RepositoryBuilder().setWorkTree(f).build();
-                return callable.invoke(repository, channel);
-            }
-        });
+        return workspace.act(new WithRepository<>(callable));
     }
 
     /**
@@ -2634,4 +2628,17 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
 
+    private static class WithRepository<T> extends MasterToSlaveFileCallable<T> {
+        private final RepositoryCallback<T> callable;
+
+        public WithRepository(RepositoryCallback<T> callable) {
+            this.callable = callable;
+        }
+
+        @Override
+        public T invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
+            final Repository repository = new RepositoryBuilder().setWorkTree(f).build();
+            return callable.invoke(repository, channel);
+        }
+    }
 }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1551,14 +1551,14 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     private FilePath createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
         FilePath ssh = workspace.createTextTempFile("pass", ".bat",
-            "echo \"" + Secret.toString(sshUser.getPassphrase()) + "\"", false);
+            "echo \"" + quoteWindowsCredentials(Secret.toString(sshUser.getPassphrase())) + "\"", false);
         return ssh;
     }
 
     private FilePath createUnixSshAskpass(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
         FilePath ssh = workspace.createTextTempFile("pass", ".sh",
               "#!/bin/sh\n"
-            + "echo \"" + Secret.toString(sshUser.getPassphrase()) + "\"", false);
+            + "echo \"" + quoteUnixCredentials(Secret.toString(sshUser.getPassphrase())) + "\"", false);
         ssh.chmod(0700);
         return ssh;
     }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1727,10 +1727,6 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         return ssh;
     }
 
-    private String launchCommandIn(FilePath workDir, String... args) throws GitException, InterruptedException {
-        return launchCommandIn(new ArgumentListBuilder(args), workDir);
-    }
-
     private String launchCommandIn(ArgumentListBuilder args, FilePath workDir) throws GitException, InterruptedException {
         return launchCommandIn(args, workDir, environment);
     }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -21,19 +21,25 @@ import hudson.plugins.git.GitLockFailedException;
 import hudson.plugins.git.IGitAPI;
 import hudson.plugins.git.IndexEntry;
 import hudson.plugins.git.Revision;
+import hudson.remoting.Callable;
+import hudson.remoting.VirtualChannel;
 import hudson.util.ArgumentListBuilder;
 import hudson.util.Secret;
 
+import jenkins.MasterToSlaveFileCallable;
+import jenkins.security.MasterToSlaveCallable;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.jgit.lib.Constants;
 import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.lib.Ref;
 import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.lib.RepositoryBuilder;
 import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
 import org.eclipse.jgit.transport.RefSpec;
 import org.eclipse.jgit.transport.RemoteConfig;
 import org.eclipse.jgit.transport.URIish;
+import org.kohsuke.stapler.framework.io.IOException2;
 import org.kohsuke.stapler.framework.io.WriterOutputStream;
 
 import java.io.*;
@@ -54,6 +60,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
 
+import static org.eclipse.jgit.lib.Constants.*;
+
 
 /**
  * Implementation class using command line CLI ran as external command.
@@ -73,6 +81,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     static final String SPARSE_CHECKOUT_FILE_PATH = ".git/info/sparse-checkout";
     static final String TIMEOUT_LOG_PREFIX = " # timeout=";
     private static final String INDEX_LOCK_FILE_PATH = ".git" + File.separator + "index.lock";
+    private final FilePath workspace;
     transient Launcher launcher;
     TaskListener listener;
     String gitExe;
@@ -157,19 +166,39 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * @param listener a {@link hudson.model.TaskListener} object.
      * @param environment a {@link hudson.EnvVars} object.
      */
-    protected CliGitAPIImpl(String gitExe, File workspace,
-                         TaskListener listener, EnvVars environment) {
-        super(workspace);
-        this.listener = listener;
+    protected CliGitAPIImpl(String gitExe, FilePath workspace,
+                         TaskListener listener, EnvVars environment, Launcher launcher) {
+        this.workspace = workspace;
+        this.listener = listener != null ? listener : TaskListener.NULL;
         this.gitExe = gitExe;
         this.environment = environment;
 
-        launcher = new LocalLauncher(IGitAPI.verbose?listener:TaskListener.NULL);
+        this.launcher = launcher != null ?
+            launcher : new Launcher.LocalLauncher(IGitAPI.verbose?listener:TaskListener.NULL);
+    }
+
+    /** {@inheritDoc} */
+    @Deprecated
+    public boolean hasGitModules(String treeIsh) throws GitException {
+        return hasGitModules();
+    }
+
+    /** {@inheritDoc} */
+    public boolean hasGitModules() throws GitException {
+        try {
+            return workspace.child(".gitmodules").exists();
+        } catch (SecurityException ex) {
+            throw new GitException(
+                    "Security error when trying to check for .gitmodules. Are you sure you have correct permissions?",
+                    ex);
+        } catch (Exception e) {
+            throw new GitException("Couldn't check for .gitmodules", e);
+        }
     }
 
     /** {@inheritDoc} */
     public GitClient subGit(String subdir) {
-        return new CliGitAPIImpl(gitExe, new File(workspace, subdir), listener, environment);
+        return new CliGitAPIImpl(gitExe, workspace.child(subdir), listener, environment, launcher);
     }
 
     /**
@@ -178,8 +207,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
      * @throws java.lang.InterruptedException if interrupted.
      */
-    public void init() throws GitException, InterruptedException {
-        init_().workspace(workspace.getAbsolutePath()).execute();
+    public void init() throws GitException, InterruptedException, IOException {
+        init_().workspace(workspace.getRemote()).execute();
     }
 
     /**
@@ -189,7 +218,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
      * @throws java.lang.InterruptedException if interrupted.
      */
-    public boolean hasGitRepo() throws GitException, InterruptedException {
+    public boolean hasGitRepo() throws GitException, InterruptedException, IOException {
         if (hasGitRepo(".git")) {
             // Check if this is a valid git repo with --is-inside-work-tree
             try {
@@ -204,16 +233,16 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
     /**
-     * Returns true if the parameter GIT_DIR is a directory which
+     * Returns true if the parameter dir is a directory which
      * contains a git repository.
      *
-     * @param GIT_DIR a {@link java.lang.String} object.
-     * @return true if GIT_DIR has a git repository
+     * @param dir a {@link java.lang.String} object.
+     * @return true if dir has a git repository
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
      */
-    public boolean hasGitRepo( String GIT_DIR ) throws GitException {
+    public boolean hasGitRepo( String dir ) throws GitException {
         try {
-            File dotGit = new File(workspace, GIT_DIR);
+            FilePath dotGit = workspace.child(dir);
             return dotGit.exists();
         } catch (SecurityException ex) {
             throw new GitException("Security error when trying to check for .git. Are you sure you have correct permissions?",
@@ -282,7 +311,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 return this;
             }
 
-            public void execute() throws GitException, InterruptedException {
+            public void execute() throws GitException, InterruptedException, IOException {
                 listener.getLogger().println(
                         "Fetching upstream changes from " + url);
 
@@ -318,12 +347,12 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
     /** {@inheritDoc} */
-    public void fetch(URIish url, List<RefSpec> refspecs) throws GitException, InterruptedException {
+    public void fetch(URIish url, List<RefSpec> refspecs) throws GitException, InterruptedException, IOException {
         fetch_().from(url, refspecs).execute();
     }
 
     /** {@inheritDoc} */
-    public void fetch(String remoteName, RefSpec... refspec) throws GitException, InterruptedException {
+    public void fetch(String remoteName, RefSpec... refspec) throws GitException, InterruptedException, IOException {
         listener.getLogger().println(
                                      "Fetching upstream changes"
                                      + (remoteName != null ? " from " + remoteName : ""));
@@ -350,7 +379,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
     /** {@inheritDoc} */
-    public void fetch(String remoteName, RefSpec refspec) throws GitException, InterruptedException {
+    public void fetch(String remoteName, RefSpec refspec) throws GitException, InterruptedException, IOException {
         fetch(remoteName, new RefSpec[] {refspec});
     }
 
@@ -439,7 +468,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 return this;
             }
 
-            public void execute() throws GitException, InterruptedException {
+            public void execute() throws GitException, InterruptedException, IOException {
 
                 URIish urIish = null;
                 try {
@@ -452,7 +481,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 listener.getLogger().println("Cloning repository " + url);
 
                 try {
-                    Util.deleteContentsRecursive(workspace);
+                    workspace.deleteContents();
                 } catch (Exception e) {
                     e.printStackTrace(listener.error("Failed to clean the workspace"));
                     throw new GitException("Failed to delete workspace", e);
@@ -461,7 +490,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 // we don't run a 'git clone' command but git init + git fetch
                 // this allows launchCommandWithCredentials() to pass credentials via a local gitconfig
 
-                init_().workspace(workspace.getAbsolutePath()).execute();
+                init_().workspace(workspace.getRemote()).execute();
 
                 if (shared) {
                     if (reference == null || reference.isEmpty()) {
@@ -473,27 +502,24 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 }
 
                 if (reference != null && !reference.isEmpty()) {
-                    File referencePath = new File(reference);
+                    FilePath referencePath = workspace.child(reference);
                     if (!referencePath.exists())
                         listener.error("Reference path does not exist: " + reference);
                     else if (!referencePath.isDirectory())
                         listener.error("Reference path is not a directory: " + reference);
                     else {
                         // reference path can either be a normal or a base repository
-                        File objectsPath = new File(referencePath, ".git/objects");
+                        FilePath objectsPath = referencePath.child(".git/objects");
                         if (!objectsPath.isDirectory()) {
                             // reference path is bare repo
-                            objectsPath = new File(referencePath, "objects");
+                            objectsPath = referencePath.child("objects");
                         }
                         if (!objectsPath.isDirectory())
                             listener.error("Reference path does not contain an objects directory (no git repo?): " + objectsPath);
                         else {
                             try {
-                                File alternates = new File(workspace, ".git/objects/info/alternates");
-                                try (PrintWriter w = new PrintWriter(alternates)) {
-                                    // git implementations on windows also use
-                                    w.print(objectsPath.getAbsolutePath().replace('\\', '/'));
-                                }
+                                FilePath alternates = workspace.child(".git/objects/info/alternates");
+                                alternates.write(objectsPath.getRemote().replace('\\', '/'), null);
                             } catch (FileNotFoundException e) {
                                 listener.error("Failed to setup reference");
                             }
@@ -673,13 +699,13 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
      * @throws java.lang.InterruptedException if interrupted.
      */
-    public void clean() throws GitException, InterruptedException {
+    public void clean() throws GitException, InterruptedException, IOException {
         reset(true);
         launchCommand("clean", "-fdx");
     }
 
     /** {@inheritDoc} */
-    public ObjectId revParse(String revName) throws GitException, InterruptedException {
+    public ObjectId revParse(String revName) throws GitException, InterruptedException, IOException {
 
         String arg = sanitize(revName + "^{commit}");
         String result = launchCommand("rev-parse", arg);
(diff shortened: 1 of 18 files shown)


ndeloof:
Not sure about CliGitAPIImplTest.test_checkout_interrupted_with_existing_lock failure on CI. Works on my machine (I hate to say that), maybe a race condition ?

ndeloof:
@reviewbybees

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1872,29 +1901,8 @@ public void execute() throws GitException, InterruptedException {
...
-            if (branches.size() == 1) {
-                listener.getLogger().println("Seen 1 remote branch");
-            } else {
-                listener.getLogger().println(MessageFormat.format("Seen {0} remote branches", branches.size()));
-            }
-
-            return branches;
-        }
+    public Set<Branch> getRemoteBranches() throws GitException, InterruptedException, IOException {
+        return parseBranches(launchCommand("branch", "-r", "-v", "--no-abbrev"));
```
ndeloof: wonder why the initial code was relying on JGit within the CLIGit implementation. Would prefer not to mix such changes in this PR, but had no choice here.

ndeloof:
@MarkEWaite test failure fixed, open for review

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1486,15 +1512,11 @@ private String launchCommandWithCredentials(ArgumentListBuilder args, File workD
...
-            List<String> privateKeys = sshUser.getPrivateKeys();
-            for (String s : privateKeys) {
-                w.println(s);
-            }
-        }
-        new FilePath(key).chmod(0400);
+    private FilePath createSshKeyFile(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
+        FilePath key = workspace.createTempFile("ssh", "key");
+        List<String> privateKeys = sshUser.getPrivateKeys();
+        key.write(StringUtils.join(privateKeys, "\n"), null);
```
jtnord: smells like a bug to me the OS may be windows which does not use '\n' but '\r\n'

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1510,51 +1532,42 @@ private String quoteUnixCredentials(String str) {
...
-            w.println("@if (%arg:~0,8%)==(Username) echo " + quoteWindowsCredentials(creds.getUsername()));
-            w.println("@if (%arg:~0,8%)==(Password) echo " + quoteWindowsCredentials(Secret.toString(creds.getPassword())));
-        }
-        askpass.setExecutable(true);
+    private FilePath createWindowsStandardAskpass(StandardUsernamePasswordCredentials creds) throws IOException, InterruptedException {
+        FilePath askpass = workspace.createTempFile("pass", ".bat");
+        askpass.write("@set arg=%~1\n"
+                    + "@if (%arg:~0,8%)==(Username) echo " + quoteWindowsCredentials(creds.getUsername()) + "\n"
+                    + "@if (%arg:~0,8%)==(Password) echo " + quoteWindowsCredentials(Secret.toString(creds.getPassword())), null);
+        askpass.chmod(0700);
```
jtnord: what do you expect this to do on windows :-)

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1510,51 +1532,42 @@ private String quoteUnixCredentials(String str) {
...
-        File ssh = File.createTempFile("pass", ".bat");
-        try (PrintWriter w = new PrintWriter(ssh)) {
-            w.println("echo \"" + quoteWindowsCredentials(Secret.toString(sshUser.getPassphrase())) + "\"");
-            w.flush();
-        }
-        ssh.setExecutable(true);
+    private FilePath createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
+        FilePath ssh = workspace.createTempFile("pass", ".bat");
+        ssh.write("echo \"" + Secret.toString(sshUser.getPassphrase()) + "\"", null);
+        ssh.chmod(0700);
```
jtnord: what do you expect this to do on windows :-)

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: right, but git on windows relies on openssl and can read unix EOF without issue

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: FilePath has no equivalent for setExecutable I know, I'm open to any guidance here

MarkEWaite: I'm not sure that we want the unencrypted ssh passphrase written into the workspace.  Isn't that even more widely available than the system temporary directory (through the "Workspace" browsing facilities in Jenkins).

I'm considering switching from `echo` to `type` and `cat` so that the shell escaping rules of echo don't have to be applied to the passphrase.  Refer to https://github.com/MarkEWaite/git-client-plugin/pull/2/commits/22ae93a9f5d1088fb353995e69bda2cb0927c839 for the beginnings of that code.

I'd rather have this change (modify the location of the passphrase) be considered in the context of #207, since it is also touching the same area, but touching it in a way that fixes a Windows problem.

MarkEWaite: The chmod (and setExecutable) are not needed on Windows.  Files that end with the `bat` suffix are executable.

jtnord: should be in `workspace@tmp`

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1828,14 +1838,15 @@ public void execute() throws GitException, InterruptedException {
      */
     private Set<Branch> parseBranches(String fos) {
         Set<Branch> branches = new HashSet<>();
+        Set<String> links = new HashSet<String>();
         BufferedReader rdr = new BufferedReader(new StringReader(fos));
         String line;
         try {
             while ((line = rdr.readLine()) != null) {
-                if (line.length() < 44 || !line.contains(" ")) {
+                if (line.length() < 5 || !line.contains(" ")) {
```
jtnord: Seems a bit unrelated and a big change?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1846,10 +1857,24 @@ public void execute() throws GitException, InterruptedException {
...
                 // Fields are separated by one or more spaces
                 String[] branchVerboseOutput = line.substring(2).split(" +", 3);
+                if (branchVerboseOutput[1].equals("->")) {
+                    links.add(line);
+                }
                 if (branchVerboseOutput[1].length() == 40) {
                     branches.add(new Branch(branchVerboseOutput[0], ObjectId.fromString(branchVerboseOutput[1])));
                 }
             }
+            for (String link : links) {
```
jtnord: seems unrelated to the change in question?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: Moving to workspace@tmp is out of scope for this PR. 
@MarkEWaite tmp file is deleted on finally block, isn't this safe enough ? If you don't think so, can rely on FilePath#createTextTempFile which uses java.io.tmpdir

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: indeed. I had to re-implement getRemoteBranches in plain git cli as the legacy code was relying on JGIT for this CLI implementation (sic), and JGit doesn't behave the same as `git branch -r`.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: JGIT do report HEADs as plain branches, while git CLI do use `->` and reference to another branch

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
MarkEWaite: If (as @jtnord says) workspace.createTempFile() creates the file in a directory which is not visible from the workspace link in the Jenkins job, then that makes your change much better than the previous implementation.  The previous implementation wrote a plaintext secret to java.io.tmpdir, while your change writes a plaintext secret to a file into that directory hidden inside the Jenkins workspace.

Both the old and new implementation remove the plaintext secret file in the finally block.

ndeloof: applied 2c321ce8334176270983d3b59cda8fb148f8a314 to rely on createTextTempFile so changes here just act as the initial implementation. Topic about moving this tmp file to another (better?) place is unrelated.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1510,51 +1532,41 @@ private String quoteUnixCredentials(String str) {
...
-        try (PrintWriter w = new PrintWriter(askpass)) {
-            w.println("@set arg=%~1");
-            w.println("@if (%arg:~0,8%)==(Username) echo " + quoteWindowsCredentials(creds.getUsername()));
-            w.println("@if (%arg:~0,8%)==(Password) echo " + quoteWindowsCredentials(Secret.toString(creds.getPassword())));
-        }
-        askpass.setExecutable(true);
+    private FilePath createWindowsStandardAskpass(StandardUsernamePasswordCredentials creds) throws IOException, InterruptedException {
+        FilePath askpass = workspace.createTextTempFile("pass", ".bat",
+                  "@set arg=%~1\r\n"
+                + "@if (%arg:~0,8%)==(Username) echo " + quoteWindowsCredentials(creds.getUsername()) + "\n\n"
```
jtnord: `\n\n`?

ndeloof: good catch

ndeloof:
rebased

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1182,8 +1208,8 @@ public boolean isBareRepository(String GIT_DIR) throws GitException, Interrupted
         return !"false".equals(line);
     }
 
-    public boolean isShallowRepository() {
-        return new File(workspace, pathJoin(".git", "shallow")).exists();
+    public boolean isShallowRepository() throws IOException, InterruptedException {
+        return workspace.child(".git/shallow").exists();
```
jtnord: NIT:  `child(".git").child("shallow").exists()`.

ndeloof: FilePath do internally handle path separator normalization
(https://github.com/jenkinsci/jenkins/blob/fdabb525034e587bab2ec56eb7fc4c2431908232/core/src/main/java/hudson/FilePath.java#L265-L265)

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -850,14 +873,23 @@ public String getRemoteUrl(String name) throws GitException {
...
         try {
             return new RepositoryBuilder().setWorkTree(workspace).build();
         } catch (IOException e) {
             throw new GitException(e);
         }
     }
 
+    public <T> T withRepository(RepositoryCallback<T> callable) throws IOException, InterruptedException {
+        Repository repo = getRepository();
+        try {
```
jtnord: if repository is `AutoCloseable` then try-with-resources is better here - if both the callable and the close fails the initial error would be swallowed with the code as is :-(

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -1813,6 +1845,8 @@ else if (refspec != null)
             revListCommand.execute();
         } catch (InterruptedException e) {
             throw new GitException(e);
+        } catch (IOException e) {
```
jtnord: multi-catch

src/main/java/org/jenkinsci/plugins/gitclient/LegacyCompatibleGitAPIImpl.java
```diff
@@ -144,51 +120,58 @@ public void push(String remoteName, String refspec) throws GitException, Interru
...
-            final List<Tag> ret = new ArrayList<>();
-
-            for (final Map.Entry<String, Ref> tag : db.getTags().entrySet()) {
-                Ref value = tag.getValue();
-                if (value != null) {
-                    final ObjectId tagId = value.getObjectId();
-                    if (commit != null && commit.equals(tagId))
-                        ret.add(new Tag(tag.getKey(), tagId));
+    public List<Tag> getTagsOnCommit(final String revName) throws GitException, IOException, InterruptedException {
+        return withRepository(new RepositoryCallback<List<Tag>>() {
```
jtnord: Seems like this needs to be a static inner clsas otherwise the outer Object will be serialized.  (not sure if this gets sent to run on a slave - but it smells like it might...)

ndeloof: I didn't tried to fix those Deprecated methods, would prefer to remove them all but this would make this PR even harder to review

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1496,15 +1522,11 @@ private String launchCommandWithCredentials(ArgumentListBuilder args, File workD
...
-        key = File.createTempFile("ssh", "key");
-        try (PrintWriter w = new PrintWriter(key)) {
-            List<String> privateKeys = sshUser.getPrivateKeys();
-            for (String s : privateKeys) {
-                w.println(s);
-            }
-        }
-        new FilePath(key).chmod(0400);
+    private FilePath createSshKeyFile(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
+        FilePath key = workspace.createTempFile("ssh", "key");
```
jtnord: this smells insecure.  use [`WorkspaceList.tmpDir(workspace)`](https://github.com/jenkinsci/jenkins/commit/307bfc17385c14aefcb623ccc3e7144e4f01e2a3)

If you can not bump a version to that write a helper that does the same thing.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1520,51 +1542,41 @@ private String quoteUnixCredentials(String str) {
...
 
-    private File createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOException {
-        File ssh = File.createTempFile("pass", ".bat");
-        try (PrintWriter w = new PrintWriter(ssh)) {
-            w.println("echo \"" + quoteWindowsCredentials(Secret.toString(sshUser.getPassphrase())) + "\"");
-            w.flush();
-        }
-        ssh.setExecutable(true);
+    private FilePath createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
+        FilePath ssh = workspace.createTextTempFile("pass", ".bat",
```
jtnord: this smells insecure.  use [`WorkspaceList.tmpDir(workspace)`](https://github.com/jenkinsci/jenkins/commit/307bfc17385c14aefcb623ccc3e7144e4f01e2a3)

If you can not bump a version to that write a helper that does the same thing.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1520,51 +1542,41 @@ private String quoteUnixCredentials(String str) {
...
 
-    private File createUnixSshAskpass(SSHUserPrivateKey sshUser) throws IOException {
-        File ssh = File.createTempFile("pass", ".sh");
-        try (PrintWriter w = new PrintWriter(ssh)) {
-            w.println("#!/bin/sh");
-            w.println("echo '" + quoteUnixCredentials(Secret.toString(sshUser.getPassphrase())) + "'");
-        }
-        ssh.setExecutable(true);
+    private FilePath createUnixSshAskpass(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
+        FilePath ssh = workspace.createTextTempFile("pass", ".sh",
```
jtnord: this smells insecure.  use [`WorkspaceList.tmpDir(workspace)`](https://github.com/jenkinsci/jenkins/commit/307bfc17385c14aefcb623ccc3e7144e4f01e2a3)

If you can not bump a version to that write a helper that does the same thing.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1520,51 +1542,41 @@ private String quoteUnixCredentials(String str) {
...
-    private File createWindowsStandardAskpass(StandardUsernamePasswordCredentials creds) throws IOException {
-        File askpass = File.createTempFile("pass", ".bat");
-        try (PrintWriter w = new PrintWriter(askpass)) {
-            w.println("@set arg=%~1");
-            w.println("@if (%arg:~0,8%)==(Username) echo " + quoteWindowsCredentials(creds.getUsername()));
-            w.println("@if (%arg:~0,8%)==(Password) echo " + quoteWindowsCredentials(Secret.toString(creds.getPassword())));
-        }
-        askpass.setExecutable(true);
+    private FilePath createWindowsStandardAskpass(StandardUsernamePasswordCredentials creds) throws IOException, InterruptedException {
+        FilePath askpass = workspace.createTextTempFile("pass", ".bat",
```
jtnord: this smells insecure.  use [`WorkspaceList.tmpDir(workspace)`](https://github.com/jenkinsci/jenkins/commit/307bfc17385c14aefcb623ccc3e7144e4f01e2a3)

If you can not bump a version to that write a helper that does the same thing.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1520,51 +1542,41 @@ private String quoteUnixCredentials(String str) {
...
-        try (PrintWriter w = new PrintWriter(askpass)) {
-            w.println("#!/bin/sh");
-            w.println("case \"$1\" in");
-            w.println("Username*) echo '" + quoteUnixCredentials(creds.getUsername()) + "' ;;");
-            w.println("Password*) echo '" + quoteUnixCredentials(Secret.toString(creds.getPassword())) + "' ;;");
-            w.println("esac");
-        }
-        askpass.setExecutable(true);
+    private FilePath createUnixStandardAskpass(StandardUsernamePasswordCredentials creds) throws IOException, InterruptedException {
+        FilePath askpass = workspace.createTextTempFile("pass", ".sh",
```
jtnord: this smells insecure.  use [`WorkspaceList.tmpDir(workspace)`](https://github.com/jenkinsci/jenkins/commit/307bfc17385c14aefcb623ccc3e7144e4f01e2a3)

If you can not bump a version to that write a helper that does the same thing.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1681,44 +1695,39 @@ private File getSSHExeFromGitExeParentDir(String userGitExe) {
         throw new RuntimeException("ssh executable not found. The git plugin only supports official git client http://git-scm.com/download/win");
     }
 
-    private File createWindowsGitSSH(File key, String user) throws IOException {
-        File ssh = File.createTempFile("ssh", ".bat");
+    private FilePath createWindowsGitSSH(FilePath key, String user) throws IOException, InterruptedException {
+        FilePath ssh = workspace.createTempFile("ssh", ".bat");
```
jtnord: not insecure - but no need to clutter the workspace.
use [`WorkspaceList.tmpDir(workspace)`](https://github.com/jenkinsci/jenkins/commit/307bfc17385c14aefcb623ccc3e7144e4f01e2a3)

If you can not bump a version to that write a helper that does the same thing.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: My goal here is to make git-client remote-compliant, and doing this I keep the existing implementation decision. Relying on workspace@tmp is another topic.

jtnord:
:bug: there is a lot that smells insecure to me - and a lot of places where the workspace is littered with temporary files without due need.

Use [`WorkspaceList.tmpDir(workspace)`](http://javadoc.jenkins-ci.org/hudson/slaves/WorkspaceList.html#tempDir-hudson.FilePath-) if you can use a parent >= 1.652  

If you can not bump a version to that write a helper that does the same thing.

ndeloof:
@jtnord removed use of createTempFile so temp files are created under /tmp (as it was previously)

use of workspace@tmp is another (arguable) topic I don't plan to address here

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
jtnord: no if this introduces a security vulnerability (by being able to retrive credentials/keys via the workspace browser) the it is not another topic at all.

jtnord:
@ndeloof check the javadoc for [`FilePath.createTempFile()`](http://javadoc.jenkins-ci.org/hudson/FilePath.html#createTextTempFile-java.lang.String-java.lang.String-java.lang.String-)

Creates a temporary file in **this** directory and set the contents to the given text (encoded in the platform default encoding)

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
jtnord: ^^ still in the workspace...

jtnord:
:bee:

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ndeloof: So the use of `createTextTempFile(..., false)`, so java.io.tmpDir is being used for temporary files.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2317,19 +2314,15 @@ private void env(String name, String value) {
...
-    @NonNull
-    public Repository getRepository() throws GitException {
-        try {
-            return FileRepositoryBuilder.create(new File(workspace, Constants.DOT_GIT));
-        } catch (IOException e) {
-            throw new GitException("Failed to open Git repository " + workspace, e);
-        }
+    @Override
+    public <T> T withRepository(final RepositoryCallback<T> callable) throws IOException, InterruptedException {
+        return workspace.act(new MasterToSlaveFileCallable<T>() {
```
MarkEWaite: stephenc: > I am concerned about the anonymous inner class here that may pull over the outer class reference. Does it make sense to switch to an explicit named static inner class so that we clearly identify the minimal set of class references to send over the remoting reference

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -116,23 +126,37 @@ public Git using(String exe) {
...
-        jenkins.MasterToSlaveFileCallable<GitClient> callable = new jenkins.MasterToSlaveFileCallable<GitClient>() {
-            public GitClient invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
-                if (listener == null) listener = TaskListener.NULL;
-                if (env == null) env = new EnvVars();
 
-                if (exe == null || JGitTool.MAGIC_EXENAME.equalsIgnoreCase(exe)) {
+        if (env == null) env = new EnvVars();
+        if (repository == null) {
+            // Assume local directory
+            if (launcher != null) {
```
MarkEWaite: Note to self - the launcher "." directory is the directory from which java -jar slave.jar was invoked, not the slave home directory, and not the workspace directory.  Add a comment to make that clear

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -116,23 +126,37 @@ public Git using(String exe) {
...
     public GitClient getClient() throws IOException, InterruptedException {
-        jenkins.MasterToSlaveFileCallable<GitClient> callable = new jenkins.MasterToSlaveFileCallable<GitClient>() {
-            public GitClient invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
-                if (listener == null) listener = TaskListener.NULL;
-                if (env == null) env = new EnvVars();
 
-                if (exe == null || JGitTool.MAGIC_EXENAME.equalsIgnoreCase(exe)) {
+        if (env == null) env = new EnvVars();
+        if (repository == null) {
+            // Assume local directory
```
MarkEWaite: Note to self: move this comment to the `else` branch

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -116,23 +126,37 @@ public Git using(String exe) {
...
+            if (launcher != null) {
+                repository = new FilePath(launcher.getChannel(), ".");
+            } else {
+                repository = new FilePath(new File("."));
+            }
+        }
+        GitClient git;
+
+        if (exe == null || JGitTool.MAGIC_EXENAME.equalsIgnoreCase(exe)) {
+            FileCallable<GitClient> callable = new MasterToSlaveFileCallable<GitClient>() {
```
MarkEWaite: stephenc: > fairly certain this should be a named static inner class not anonymous inner

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -116,23 +126,37 @@ public Git using(String exe) {
...
+            // Assume local directory
+            if (launcher != null) {
+                repository = new FilePath(launcher.getChannel(), ".");
+            } else {
+                repository = new FilePath(new File("."));
+            }
+        }
+        GitClient git;
+
+        if (exe == null || JGitTool.MAGIC_EXENAME.equalsIgnoreCase(exe)) {
```
MarkEWaite: @ndeloof Does the exe == null condition here mean  that now the default git implementation on an agent is JGit, whether or not the job selected JGit as the implementation?

If so, that's bad.  Users want their job selected git implementation honored, even if it is the system default.

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -116,23 +126,37 @@ public Git using(String exe) {
...
+        } else {
+            // Ensure we return a backward compatible GitAPI, even API only claim to provide a GitClient
+            git = new GitAPI(exe, repository, listener, env, launcher);
+        }
+
         Jenkins jenkinsInstance = Jenkins.getInstance();
         if (jenkinsInstance != null && git != null)
             git.setProxy(jenkinsInstance.proxy);
         return git;
+
```
MarkEWaite: Remove this added blank line

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -47,6 +47,7 @@
 
 import javax.annotation.Nullable;
 
+import jenkins.model.Jenkins;
```
MarkEWaite: stephenc: > I am concerned greatly that you are importing Jenkins as that is a class that *should* (never mind that other plugins such as Subversion are bold and do bad things, but we should be getting better not worse) not be loading on agents. Given that we explicitly construct this class on a remote agent I would expect to see sufficient classloading guard methods to ensure that the methods referencing `Jenkins` are never called on execution paths which will be invoked on the agent and are only for execution paths running on the `Jenkins` master JVM... that is almost impossible to do without specifying a core dependency of 1.653+ (see `JenkinsJVM` class added in 1.653) so I think this is a bad change

MarkEWaite: @ndeloof 

stephenc: > I also believe given the usage of this class in [this method](https://github.com/jenkinsci/git-client-plugin/pull/212/files#diff-88af64f5dbd015aa6e8f00f355f34700R146) the intent is that the JGitAPIImpl is supposed to be instantiated on the agent and not on the master. As such I would expect to see a `writeReplace` method that replaces `this` with `Channel.export(GitAPI.class,this)` in those cases where `Channel.current() != null`... further as I suspect that the credentials conversion in this class is probably written to assume that it is being invoked on the master JVM it may be necessary to wrap the exported proxy object in a delegating `Serializable` hand rolled proxy that unencrypts the credentials before passing through to the remoting proxy so that the decryption happens on the master (which is the only place that has access to the secret key)

src/main/java/org/jenkinsci/plugins/gitclient/LegacyCompatibleGitAPIImpl.java
MarkEWaite: stephenc: > static inner class++

src/main/java/org/jenkinsci/plugins/gitclient/RemoteGitImpl.java
```diff
@@ -140,9 +141,9 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
             throw new IllegalStateException("Unexpected invocation: "+method);
         }
 
-        public void execute() throws GitException, InterruptedException {
+        public void execute() throws GitException, InterruptedException, IOException {
             try {
-                channel.call(new jenkins.security.MasterToSlaveCallable<Void, GitException>() {
+                channel.call(new MasterToSlaveCallable<Void, GitException>() {
```
MarkEWaite: stephenc: > static inner class

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
ndeloof: @MarkEWaite I haven't changed the condition here

ndeloof: @MarkEWaite I haven't changed the condition

ndeloof:
build on ci.jenkins.io fails due to missing `git config --global user.email`.

ndeloof:
@MarkEWaite why did you closed this PR ?

MarkEWaite:
It was closed by GitHub as a side effect of my created the 1.21.x branch and merging the 2.0.0-beta branch to the master branch, then deleting the 2.0.0-beta branch.  Since git client 2.0.0 has released, could you reopen this pull request against the master branch?  Evaluation can then continue against the master branch.

PR #214, opened 2016-08-04
[JENKINS-37185] Honor checkout timeout

superboum:
**Expectation**: When I set the checkout timeout, I'm expecting that Jenkins uses it when it comes to checkout a commit.

**Problem**: However, it appears that when you select the "Check out to a specific local branch", the first checkout command does not set correctly the timeout.

**Why it's a problem**: I'm using git lfs, and git lfs files are downloaded during the checkout option. For other reasons I can't use the git lfs pull command. 

**How to reproduce it**: 
 1. Use any version of git-plugin and git-client-plugin. 
 2. Create a freestyle project, add the following parameters: 
   1. Timeout=120 for the Advanced checkout option
   2. ** for the Checkout to a specific local branch. 
 3. Build the project.

The first checkout will be displayed with a timeout of 10 instead of 120.

**Before**:

```
Checking out Revision 2695971df91e03443bd2aa11109dca740c341448 (refs/remotes/origin/master)
 > git.exe config core.sparsecheckout # timeout=10
 > git.exe checkout -f 2695971df91e03443bd2aa11109dca740c341448 # timeout=10
 > git.exe branch -a -v --no-abbrev # timeout=10
 > git.exe branch -D master # timeout=10
 > git.exe checkout -b master 2695971df91e03443bd2aa11109dca740c341448 # timeout=120
 > git.exe rev-list 2695971df91e03443bd2aa11109dca740c341448 # timeout=10
Finished: SUCCESS
```

**After**:  
*(Check the third line)*

```
Checking out Revision 2695971df91e03443bd2aa11109dca740c341448 (refs/remotes/origin/master)
 > git.exe config core.sparsecheckout # timeout=10
 > git.exe checkout -f 2695971df91e03443bd2aa11109dca740c341448 # timeout=120
 > git.exe branch -a -v --no-abbrev # timeout=10
 > git.exe checkout -b master 2695971df91e03443bd2aa11109dca740c341448 # timeout=120
 > git.exe rev-list 2695971df91e03443bd2aa11109dca740c341448 # timeout=10
Finished: SUCCESS
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1976,7 +1976,9 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
                     if (branch!=null && deleteBranch) {
                         // First, checkout to detached HEAD, so we can delete the branch.
-                        launchCommand("checkout", "-f", ref);
+                        ArgumentListBuilder args = new ArgumentListBuilder();
+                        args.add("checkout", "-f", ref);
+                        launchCommandIn(args, workspace, environment, timeout);
 
                         // Second, check to see if the branch actually exists, and then delete it if it does.
                         for (Branch b : getBranches()) {
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -3409,6 +3409,20 @@ public abstract class GitAPITestCase extends TestCase {
         assertEquals("Wrong SHA1 as checkout of git-client-1.6.0", sha1Expected, sha1);
     }
 
+    @Bug(37185)
+    @NotImplementedInJGit /* JGit doesn't have timeout */
+    public void test_checkout_honor_timeout() throws Exception {
+        w = clone(localMirror());
+
+        int newTimeout = 7;
+        w.git.checkout().branch("master").ref("origin/master").timeout(newTimeout).deleteBranchIfExist(true).execute();
+
+        Map timeouts = new HashMap<Integer, Integer>();
+        timeouts.put(1, newTimeout);
+        timeouts.put(4, newTimeout);
+        setExpectedTimeoutAtPositions(timeouts);
+    }
+
     @Bug(25353)
     @NotImplementedInJGit /* JGit lock file management ignored for now */
     public void test_checkout_interrupted() throws Exception {
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -71,6 +71,7 @@ import com.google.common.base.Function;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.Lists;
+import java.util.Iterator;
 
 /**
  * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
@@ -388,7 +389,6 @@ public abstract class GitAPITestCase extends TestCase {
             }
         } else {
             assertEquals("Wrong timeout count", expectedTimeouts.size(), timeouts.size());
-            timeouts = expectedTimeouts;
         }
         assertEquals("Wrong timeout", expectedTimeouts, timeouts);
     }
@@ -474,7 +474,24 @@ public abstract class GitAPITestCase extends TestCase {
             setExpectedTimeouts(expected);
         }
     }
-
+    private void setExpectedTimeoutAtPositions(Map<Integer, Integer> newTimeouts) {
+        if (getTimeoutVisibleInCurrentTest()) {
+            int size = handler.getTimeouts().size();
+            List<Integer> expected = new ArrayList<Integer>(size);
+            for (int i = 0; i < size; i++) {
+                expected.add(i, CliGitAPIImpl.TIMEOUT);
+            }
+            
+            Iterator<Map.Entry<Integer, Integer>> entries = newTimeouts.entrySet().iterator();
+            while(entries.hasNext()) {
+                Map.Entry<Integer, Integer> entry = entries.next();
+                Integer position = entry.getKey();
+                Integer newTimeout = entry.getValue();
+                expected.set(position, newTimeout);
+            }
+            setExpectedTimeouts(expected);
+        }
+    }
     /** Clone arguments include:
      *   repositoryName(String) - if omitted, CliGit does not set a remote repo name
      *   shallow() - no relevant assertion of success or failure of this argument
@@ -497,7 +514,9 @@ public abstract class GitAPITestCase extends TestCase {
         assertFalse("Alternates file found: " + alternates, w.exists(alternates));
         assertFalse("Unexpected shallow clone", w.cgit().isShallowRepository());
 
-        setExpectedTimeoutWithAdjustedEnd(newTimeout);
+        Map timeouts = new HashMap<Integer, Integer>();
+        timeouts.put(2, newTimeout);
+        setExpectedTimeoutAtPositions(timeouts);
     }
 
     public void test_checkout_exception() throws Exception {
@@ -2062,7 +2081,7 @@ public abstract class GitAPITestCase extends TestCase {
         assertFalse("file2 exists and should not because not on 'branch1'", w.exists(subFile2));
         assertFalse("file3 exists and should not because not on 'branch2'", w.exists(subFile3));
 
-        setExpectedTimeoutWithAdjustedEnd(newTimeout, 2);
+        setExpectedTimeoutWithAdjustedEnd(newTimeout, 1);
     }
 
     @NotImplementedInJGit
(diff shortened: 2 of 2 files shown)


MarkEWaite:
Could I persuade you to fix the test which was trying to check this condition?  The test checks for the presence of timeout=xxx, but apparently does not check for the presence of that timeout value on every checkout statement.

superboum:
Haha, you can. I'll try to do it this night (my timezone is UTC-5).

superboum:
I don't understand the `checkTimeout` function in the `git-client-plugin` test files:

```java
//\src\test\java\org\jenkinsci\plugins\gitclient\GitAPITestCase.java
private void checkTimeout() {
    List<Integer> timeouts = handler.getTimeouts();
    if (expectedTimeouts == null) {
        expectedTimeouts = new ArrayList<Integer>();
        for (int i = 0; i < timeouts.size(); i++) {
            expectedTimeouts.add(i, CliGitAPIImpl.TIMEOUT);
        }
    } else {
        assertEquals("Wrong timeout count", expectedTimeouts.size(), timeouts.size());
        timeouts = expectedTimeouts; //The strange line
    }
    assertEquals("Wrong timeout", expectedTimeouts, timeouts);
}
```

I think this function goal is to check that timeouts read in the logs are not different that what we expect.
And there are two cases:

 * We have not set a list of expectedTimeouts. So we create an empty list. We fill it with the default timeout, as many time as we have found timeouts 
 * We have set a list of expectedTimeouts with  `setExpectedTimeoutWithAdjustedEnd()`. In this case, we compare the size. *There is practically no reason it fails. Indeed the previous function takes the timeouts count to fill it with default value and replace n time the last values with the given value.* And after that we are doing a **really strange thing**. We overwrite our tests values, stored in the timeouts variable with the value we expect

And finally we compare our tests values stored in `timeouts` with `expectedTimeouts`. But in the second case, both variable are pointing to the same object. So we are not comparing anything ?

Did I miss something ?

MarkEWaite:
> Did I miss something ?

I don't think you missed anything.  I think that line is nonsense (and I wrote the line).  Sorry about that mistake.  A test that cannot fail so long as the count of timeouts matches.  Ugh, that's bad.

superboum:
As an example, the following code will not fail:

```java
// Switch to master
int newTimeout = 6;
w.git.submoduleUpdate().remoteTracking(true).useBranch(submodDir, "master").timeout(newTimeout).execute();
assertFalse("file2 exists and should not because not on 'branch1'", w.exists(subFile2));
assertFalse("file3 exists and should not because not on 'branch2'", w.exists(subFile3));
// line 2060 of GitAPITestCase.java
// We can put any value, it will not fail the test.
// I've commented the original line and put a new one with something totally random
//setExpectedTimeoutWithAdjustedEnd(newTimeout, 2);
setExpectedTimeoutWithAdjustedEnd(43, 2);
```

Moreover, this same code (with the `newTimeout` value instead of `43`) should also fail, even if it's pass now (because of the bug). As you can see on the following screenshot from Netbeans, the `test_trackingSubmoduleBranches()` test produce only one timeout=6, the others are equal to 10. But with the function `setExpectedTimeoutWithAdjustedEnd(newTimeout, 2);` you expect 2 of them at the end.

wrongtest


Anyway, thanks for the reply, I'll do my best to fix that :D

superboum:
Some observations: 
 1. There is no timeout in the JGit tests.
 2. 2 tests are failing because of the position of the timeout: `[10,7,10,10,10] != [10,10,10,10,7]`

So, there are (at least) two options:
  1. Continue to push the timeout at the end, and comparing only that there is at least one occurence of `7` and four occurence of `10` (it's an example) but don't check the order.
  2. Removing `setExpectedTimeoutWithAdjustedEnd(final int newTimeout)` and `setExpectedTimeoutWithAdjustedEnd(final int newTimeout, int adjustmentCount)`. We will call directly the `setExpectedTimeouts(List<Integer> timeouts)` function with the list we want.

Pros and cons:
 1.  It might be more flexible according to the implementation but we can't be sure that the values are set on the command we expect.
 2. Less flexible, if we use a slightly different git implementation which run one more command with a timeout it will fail.

After writing that, I'm understanding that the `setExpectedTimeoutWithAdjustedEnd` has for sole purpose to be a helper to fill a list with the default timeout without explicitly declaring the number of timeouts.

As a solution, I propose a new function `setExpectedTimeoutAtPositions(Map<Integer, Integer> newTimeouts)`. The first one would be the position, the second one the value of the timeout. Something like:

```java
private void setExpectedTimeoutAtPositions(Map<Integer, Integer> newTimeouts) {
    if (getTimeoutVisibleInCurrentTest()) {
        int size = handler.getTimeouts().size();
        List<Integer> expected = new ArrayList<Integer>(size);
        for (int i = 0; i < size; i++) {
            expected.add(i, CliGitAPIImpl.TIMEOUT);
        }
        while(newTimeouts.hasNext()) {
            Map.Entry<Integer, Integer> entry = newTimeouts.next();
            Integer position = entry.getKey();
            Integer newTimeout = entry.getValue();
            expected.set(position, newTimeout);
        }
        setExpectedTimeouts(expected);
    }
}
```

What do you think of it ?

superboum:
I've committed a debatable proposition for the `checkTimeout()` problem in the commit 4187c8bb3e46e23c48260e524b0e53808da899ce .  As I don't know the whole Jenkins ecosystem, I don't know if this solution will fit your needs. I'm open to comments and suggestions.

superboum:
After a check, the test `test_sparse_checkout()` is now testing the timeout on the 2 checkout commands. (I've put a breakpoint at the end of the function, and have watched the variable `this.handler.messages.toString()`). Here is the output of the test with new lines added - `s/, /\n/`).

```
Logging started
 > git --version # timeout=10
 > git init C:\Users\QUENTI~1.DUF\AppData\Local\Temp\hudson3042771801376424834test # timeout=10
[hudson3042771801376424834test] $ git commit --allow-empty -m init
 > git add dir1/file1 # timeout=10
 > git add dir2/file2 # timeout=10
 > git add dir3/file3 # timeout=10
 > git commit -F C:\Users\QUENTI~1.DUF\AppData\Local\Temp\gitcommit2583045792915919086.txt # timeout=10
Cloning repository C:\Users\QUENTI~1.DUF\AppData\Local\Temp\hudson3042771801376424834test
 > git init C:\Users\QUENTI~1.DUF\AppData\Local\Temp\hudson652098560464312076test # timeout=10
Fetching upstream changes from C:\Users\QUENTI~1.DUF\AppData\Local\Temp\hudson3042771801376424834test
 > git --version # timeout=10
 > git -c core.askpass=true fetch --tags --progress C:\Users\QUENTI~1.DUF\AppData\Local\Temp\hudson3042771801376424834test +refs/heads/*:refs/remotes/origin/*
 > git config remote.origin.url C:\Users\QUENTI~1.DUF\AppData\Local\Temp\hudson3042771801376424834test # timeout=10
 > git config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/* # timeout=10
 > git config core.sparsecheckout # timeout=10
 > git config core.sparsecheckout true # timeout=10
 > git read-tree -mu HEAD # timeout=10
Command "git read-tree -mu HEAD" returned status code 128:
stdout: 
stderr: fatal: Not a valid object name HEAD
 > git checkout -f origin/master
 > git branch -a -v --no-abbrev # timeout=10
 > git checkout -b master origin/master
 > git config core.sparsecheckout # timeout=10
 > git read-tree -mu HEAD # timeout=10
 > git checkout -f origin/master
 > git branch -a -v --no-abbrev # timeout=10
 > git branch -D master # timeout=10
 > git checkout -b master origin/master
 > git config core.sparsecheckout # timeout=10
 > git read-tree -mu HEAD # timeout=10
 > git checkout -f origin/master
 > git branch -a -v --no-abbrev # timeout=10
 > git branch -D master # timeout=10
 > git checkout -b master origin/master
 > git config core.sparsecheckout # timeout=10
 > git read-tree -mu HEAD # timeout=10
 > git config core.sparsecheckout false # timeout=10
 > git checkout -f origin/master
 > git branch -a -v --no-abbrev # timeout=10
 > git branch -D master # timeout=10
 > git checkout -b master origin/master
 > git config core.sparsecheckout # timeout=10
 > git checkout -f origin/master # timeout=3
 > git branch -a -v --no-abbrev # timeout=10
 > git branch -D master # timeout=10
 > git checkout -b master origin/master # timeout=3
```

I'll try to add a timeout check in `test_checkoutBranch()` but I don't know yet if it's relevant.

MarkEWaite:
Thanks for that implementation.  It seems reasonable to me if we want to continue along the path I originally created, where the assertion checks for a specific sequence of timeout values in the output.

Would it be worth considering restating the assertion in a different form?  The mistake I made in a recent attempt to fix the checkout timeout was that I failed to check all of the calls to "git checkout" had a timeout value and had the correct timeout value.  Would it be better to rework the assertion to test that "all calls to git checkout have the expected timeout value"?

If you'd rather that I evaluate this pull request with your recent changes, I can do that as well, and will likely do that this weekend.

superboum:
> Would it be worth considering restating the assertion in a different form? 
> Would it be better to rework the assertion to test that "all calls to git checkout have the expected timeout value"?

I agree that this idea would give us more confidence in these tests, and should be considered.

But the scope of such changes seems to be huge for me, as this pull request is my first *real* pull request on the jenkins project. 

It might be the subject of another pull request ? 
Because I think there are many things to consider: 
 * How to check that other timeouts have the default value and not the value set for the checkout command ? we will have to parse the log, if there is a timeout value, get it. If there is a git checkout command, compare it with the `newTimeout` variable and if it's another command, we will have to compare it with the `DEFAULT_TIMEOUT`.
 * We will have to think how to refactor the code, to prevent some side effects (someone changing the clone timeout, so it will be different from the `DEFAULT_TIMEOUT` but it's expected.
 * Fix all the associated tests, clean every useless code

So, if you agree, I think that evaluating this pull request would be a good start for me. In any case, the final decision is yours :smile:. I would be happy to work on the refactor, but I can't give you any estimated date...  

Finally, I've added a test which explicitly test the checkout timeout in 742861e9e83e5775d48f33c99c2827cbe37944ed.

MarkEWaite:
I look your idea for a step-wise improvement.  I'll review the details of this change then we can decide on next steps.

MarkEWaite:
I merged your change, and made the changes in b797c98b424fca81ba523cebe1c259754ad53d3f that I was envisioning.  Could you review that commit and share if you detect any problems in it?

superboum:
Thanks, I'll review that ASAP

PR #215, opened 2016-08-13
[FIXED JENKINS-37391] Added @Symbols for GitTool and JGitTool

abayer:
[JENKINS-37391](https://issues.jenkins-ci.org/browse/JENKINS-37391)

cc @reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/git/GitTool.java
+++ b/src/main/java/hudson/plugins/git/GitTool.java
@@ -13,6 +13,7 @@ import hudson.tools.ToolProperty;
 import hudson.util.FormValidation;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
+import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/git/GitTool.java
```diff
@@ -137,7 +138,7 @@ public static void onLoaded() {
     }
 
 
-    @Extension
+    @Extension @Symbol("git")
```
oleg-nenashev: there may be other GitTool implementations. Maybe gitCli?

KostyaSha: `cli-git`

abayer: There is the jgit one here already. I'd tend to go with git for this one
since it's the default and use distinctive names for the others, but I am
persuadable. =)

MarkEWaite: Who will see the text of that symbol, and how will they see it?

Today, the caller to git-client-plugin defines the desired implementation by passing either `with("git")` or `with("jgit")` or by accepting the default ("git").  If this is visible to users, then "cli-git" is probably clearer than "git" (and better than the word "Default" which is currently displayed as the alternative to "jgit")

abayer: Right now, nothing will see it at all, but as I mentioned on the JIRA, this will enable Pipeline or Job DSL to refer to `GitTool` without needing the full package and class name, etc. I'm fine with "cli-git" or "git" or whatever - I just want a `@Symbol`. =)

jglick:
retro:bee:

PR #219, opened 2016-09-26
[JENKINS-30600] let CliGitAPIImpl rely on FilePath+Launcher remote API

ndeloof:
https://github.com/jenkinsci/git-client-plugin/pull/212 rebased on master

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2323,13 +2323,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     @Override
     public <T> T withRepository(final RepositoryCallback<T> callable) throws IOException, InterruptedException {
-        return workspace.act(new MasterToSlaveFileCallable<T>() {
-            @Override
-            public T invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
-                final Repository repository = new RepositoryBuilder().setWorkTree(f).build();
-                return callable.invoke(repository, channel);
-            }
-        });
+        return workspace.act(new WithRepository<>(callable));
     }
 
     /**
@@ -2634,4 +2628,17 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
 
+    private static class WithRepository<T> extends MasterToSlaveFileCallable<T> {
+        private final RepositoryCallback<T> callable;
+
+        public WithRepository(RepositoryCallback<T> callable) {
+            this.callable = callable;
+        }
+
+        @Override
+        public T invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
+            final Repository repository = new RepositoryBuilder().setWorkTree(f).build();
+            return callable.invoke(repository, channel);
+        }
+    }
 }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1551,14 +1551,14 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     private FilePath createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
         FilePath ssh = workspace.createTextTempFile("pass", ".bat",
-            "echo \"" + Secret.toString(sshUser.getPassphrase()) + "\"", false);
+            "echo \"" + quoteWindowsCredentials(Secret.toString(sshUser.getPassphrase())) + "\"", false);
         return ssh;
     }
 
     private FilePath createUnixSshAskpass(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
         FilePath ssh = workspace.createTextTempFile("pass", ".sh",
               "#!/bin/sh\n"
-            + "echo \"" + Secret.toString(sshUser.getPassphrase()) + "\"", false);
+            + "echo \"" + quoteUnixCredentials(Secret.toString(sshUser.getPassphrase())) + "\"", false);
         ssh.chmod(0700);
         return ssh;
     }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1727,10 +1727,6 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         return ssh;
     }
 
-    private String launchCommandIn(FilePath workDir, String... args) throws GitException, InterruptedException {
-        return launchCommandIn(new ArgumentListBuilder(args), workDir);
-    }
-
     private String launchCommandIn(ArgumentListBuilder args, FilePath workDir) throws GitException, InterruptedException {
         return launchCommandIn(args, workDir, environment);
     }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -21,19 +21,25 @@ import hudson.plugins.git.GitLockFailedException;
 import hudson.plugins.git.IGitAPI;
 import hudson.plugins.git.IndexEntry;
 import hudson.plugins.git.Revision;
+import hudson.remoting.Callable;
+import hudson.remoting.VirtualChannel;
 import hudson.util.ArgumentListBuilder;
 import hudson.util.Secret;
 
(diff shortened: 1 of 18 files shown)


jglick:
This purports to fix [JENKINS-30600](https://issues.jenkins-ci.org/browse/JENKINS-30600) IIUC?

ndeloof:
not only for  JENKINS-30600, but also would fix this one

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -204,16 +233,16 @@ public boolean hasGitRepo() throws GitException, InterruptedException {
...
      * contains a git repository.
      *
-     * @param GIT_DIR a {@link java.lang.String} object.
-     * @return true if GIT_DIR has a git repository
+     * @param dir a {@link java.lang.String} object.
+     * @return true if dir has a git repository
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
      */
-    public boolean hasGitRepo( String GIT_DIR ) throws GitException {
+    public boolean hasGitRepo( String dir ) throws GitException {
```
jglick: gratuitous?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -461,7 +490,7 @@ public void execute() throws GitException, InterruptedException {
                 // we don't run a 'git clone' command but git init + git fetch
                 // this allows launchCommandWithCredentials() to pass credentials via a local gitconfig
 
-                init_().workspace(workspace.getAbsolutePath()).execute();
+                init_().workspace(workspace.getRemote()).execute();
```
jglick: Not sure what method this is calling but probably it should take `FilePath` rather than `String`.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1503,15 +1528,10 @@ private String launchCommandWithCredentials(ArgumentListBuilder args, File workD
...
-        try (PrintWriter w = new PrintWriter(key)) {
-            List<String> privateKeys = sshUser.getPrivateKeys();
-            for (String s : privateKeys) {
-                w.println(s);
-            }
-        }
-        new FilePath(key).chmod(0400);
+    private FilePath createSshKeyFile(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
+        List<String> privateKeys = sshUser.getPrivateKeys();
+        FilePath key = workspace.createTextTempFile("ssh", "key", StringUtils.join(privateKeys, "\n"), false);
```
jglick: This should use `WorkspaceList.tempDir`.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1845,14 +1851,15 @@ public void execute() throws GitException, InterruptedException {
      */
     private Set<Branch> parseBranches(String fos) {
         Set<Branch> branches = new HashSet<>();
+        Set<String> links = new HashSet<String>();
         BufferedReader rdr = new BufferedReader(new StringReader(fos));
         String line;
         try {
             while ((line = rdr.readLine()) != null) {
-                if (line.length() < 44 || !line.contains(" ")) {
+                if (line.length() < 5 || !line.contains(" ")) {
```
jglick: ?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1863,10 +1870,24 @@ public void execute() throws GitException, InterruptedException {
                 // Split fields into branch name, SHA1, and rest of line
                 // Fields are separated by one or more spaces
                 String[] branchVerboseOutput = line.substring(2).split(" +", 3);
+                if (branchVerboseOutput[1].equals("->")) {
```
jglick: How is this related? I guess you are changing `getRemoteBranches` from using JGit to using CLI Git?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1970,17 +1970,19 @@ public CheckoutCommand timeout(Integer timeout) {
...
             /* Allow test of index.lock cleanup when checkout is interrupted */
             private void interruptThisCheckout() throws InterruptedException {
-                final File indexFile = new File(workspace.getPath() + File.separator
-                        + INDEX_LOCK_FILE_PATH);
+                final FilePath indexFile = workspace.child(INDEX_LOCK_FILE_PATH);
                 try {
-                    indexFile.createNewFile();
+                    // no equivalent for File#createNewFile
+                    if (!indexFile.exists()) {
+                        indexFile.touch(System.currentTimeMillis());
```
jglick: Whatever this is for, beware clock skew. Maybe just

```java
indexFile.write("", null);
```

?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2324,19 +2321,9 @@ private void env(String name, String value) {
...
-    @NonNull
-    public Repository getRepository() throws GitException {
-        try {
-            return FileRepositoryBuilder.create(new File(workspace, Constants.DOT_GIT));
-        } catch (IOException e) {
-            throw new GitException("Failed to open Git repository " + workspace, e);
-        }
+    @Override
+    public <T> T withRepository(final RepositoryCallback<T> callable) throws IOException, InterruptedException {
+        return workspace.act(new WithRepository<>(callable));
```
jglick: Of course we should be discouraging these anyway—it is pretty confusing to request CLI Git and then see JGit being sent to the agent.

src/main/java/org/jenkinsci/plugins/gitclient/Git.java
```diff
@@ -116,19 +126,33 @@ public Git using(String exe) {
...
 
-                if (exe == null || JGitTool.MAGIC_EXENAME.equalsIgnoreCase(exe)) {
+        if (env == null) env = new EnvVars();
+        if (repository == null) {
+            if (launcher != null) {
+                // the launcher "." directory is the directory from which java -jar slave.jar was invoked, not the slave home directory, and not the workspace directory
+                repository = new FilePath(launcher.getChannel(), ".");
+            } else {
+                // Assume local directory
+                repository = new FilePath(new File("."));
```
jglick: I do not think there is any legitimate reason for `repository == null` and we could just throw an exception?

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -850,14 +873,20 @@ public String getRemoteUrl(String name) throws GitException {
...
         try {
             return new RepositoryBuilder().setWorkTree(workspace).build();
         } catch (IOException e) {
             throw new GitException(e);
         }
     }
 
+    public <T> T withRepository(RepositoryCallback<T> callable) throws IOException, InterruptedException {
+        try (Repository repo = getRepository()) {
+            return callable.invoke(repo, Jenkins.MasterComputer.localChannel);
```
jglick: IIUC `withRepository` is being called in this case in an agent JVM? If so, we should not be referring to the `Jenkins` type at all—it is unsafe.

Maybe just pass null? I cannot find any `RepositoryCallback` implementations that actually use the `channel` parameter anyway. (And any such attempt would likely be a SECURITY-144 violation!) Maybe we could deprecate it, and switch to a

```java
T invoke(Repository repo) throws IOException, InterruptedException;
```

overload, with a `Util.isOverridden` fallback for compatibility.

src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java
```diff
@@ -228,7 +229,7 @@ public void test_git_version_ubuntu_14_04_ppa() {
         doTest("git version 2.2.2", versions);
     }
 
-    public void test_git_version_ubuntu_14_04_ppa_2_3_0() {
+    public void test_git_version_ubuntu_14_04_ppa_2_3_0() throws IOException, InterruptedException {
```
jglick: Tests can just throw `Exception`.

src/test/java/org/jenkinsci/plugins/gitclient/RemotingTest.java
```diff
@@ -34,7 +35,7 @@ public void testRemotability() throws Exception {
         channel.close();
     }
 
-    private static class Work implements Callable<Void,IOException> {
+    private static class Work extends MasterToSlaveCallable<Void, IOException> {
```
jglick: Get rid of the `checkRoles` override then.

ndeloof:
too much conflicts to rebase this 14 months old PR.

PR #220, opened 2016-09-26
Use java agent on ci.jenkins.io

abayer:
There are some issues on ci.jenkins.io with GIT_COMMITTER_NAME and
GIT_COMMITTER_EMAIL not getting set properly, but the "java" agents
either have properly configured ~/.gitconfig files or git is
autodetermining the email somehow. In any case, this should be better
in general.

cc @rtyler


MarkEWaite:
I'd rather fix the tests which have that dependency on command line git, but this seems like a small change even if it is only a short term change.

PR #221, opened 2016-10-24
[JENKINS-23476] Let git client default timeout be configurable on a global level.

erikhakansson:
Today, the default timeout for Git client is based on a [system property ](https://github.com/jenkinsci/git-client-plugin/blob/0aa5973b3fec1cc14a5e56d6f6e36cea4310d293/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L2629).

This limits configurability and requires a Jenkins restart to be changed. Some plugins, such as the [Multibranch pipeline plugin](https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Multibranch+Plugin) will use this timeout without allowing overriding. This will result in timeouts for large repos or slow connections. ([Example issue](https://issues.jenkins-ci.org/browse/JENKINS-39173))

This PR allows setting the default timeout in Jenkins global configuration, with fallback on the previously mentioned system property.

Fixes [JENKINS-23476](https://issues.jenkins-ci.org/browse/JENKINS-23476)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2060,6 +2060,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                     } else {
                         timeout = TIMEOUT;
                     }
+                } else {
+                    timeout = TIMEOUT;
                 }
             }
             listener.getLogger().println(" > " + command + (TIMEOUT_LOG_PREFIX + timeout));
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -4,8 +4,9 @@ package org.jenkinsci.plugins.gitclient;
 
 import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
+import com.cloudbees.plugins.credentials.common.UsernameCredentials;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
-import com.google.common.collect.Lists;
+
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.EnvVars;
@@ -26,6 +27,7 @@ import hudson.plugins.git.IndexEntry;
 import hudson.plugins.git.Revision;
 import hudson.util.ArgumentListBuilder;
 import hudson.util.Secret;
+
 import jenkins.model.Jenkins;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
@@ -39,17 +41,7 @@ import org.eclipse.jgit.transport.RemoteConfig;
 import org.eclipse.jgit.transport.URIish;
 import org.kohsuke.stapler.framework.io.WriterOutputStream;
 
-import java.io.BufferedReader;
-import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStreamWriter;
-import java.io.PrintWriter;
-import java.io.StringReader;
-import java.io.StringWriter;
-import java.io.Writer;
+import java.io.*;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -5,7 +5,7 @@ package org.jenkinsci.plugins.gitclient;
 import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
-
+import com.google.common.collect.Lists;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.EnvVars;
@@ -20,12 +20,13 @@ import hudson.plugins.git.Branch;
 import hudson.plugins.git.GitException;
 import hudson.plugins.git.GitLockFailedException;
 import hudson.plugins.git.GitObject;
+import hudson.plugins.git.GitTool;
 import hudson.plugins.git.IGitAPI;
 import hudson.plugins.git.IndexEntry;
 import hudson.plugins.git.Revision;
 import hudson.util.ArgumentListBuilder;
 import hudson.util.Secret;
-
+import jenkins.model.Jenkins;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.jgit.lib.Constants;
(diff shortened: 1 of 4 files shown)


daniel-beck:
While you're waiting for feedback, you could add tests for this. Quoting https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.md:

> Before submitting your change, please assure that you've added a test which verifies your change. There have been many developers involved in the git client plugin and there are many, many users who depend on the git-client-plugin. Tests help us assure that we're delivering a reliable plugin, and that we've communicated our intent to other developers in a way that they can detect when they run tests.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1,29 +1,27 @@
 package org.jenkinsci.plugins.gitclient;
 
-
 import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
-import com.cloudbees.plugins.credentials.common.UsernameCredentials;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
-
+import com.google.common.collect.Lists;
```
MarkEWaite: The [contributing guidelines](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.md) for the plugin ask that contributors not perform wholesale reformatting of files without consulting the maintainers.  Moving the location of an import seems like it leans toward wholesale reformatting.  Couldn't this be put back in the location where it was prior to your change, so that the comparison between versions is smaller?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -36,7 +34,17 @@
 import org.eclipse.jgit.transport.URIish;
 import org.kohsuke.stapler.framework.io.WriterOutputStream;
 
-import java.io.*;
+import java.io.BufferedReader;
```
MarkEWaite: What was your rationale for expanding the wildcard import?

src/main/resources/hudson/plugins/git/GitTool/global.jelly
```diff
@@ -31,5 +31,8 @@ THE SOFTWARE.
                      addCaption="${%Add Git}" hasHeader="true"
                      deleteCaption="${%Delete Git}"/>
     </f:entry>
+    <f:entry title="${%Git default timeout in minutes}" >
```
MarkEWaite: Is it intentional that the same default timeout value might eventually be used for command line git, JGit, and the upcoming JGit with apache http authentication?

The current timeout only applies to command line git, and I think it makes sense that a global default should apply across all git implementations.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1734,11 +1742,23 @@ private String launchCommandIn(ArgumentListBuilder args, File workDir, EnvVars e
...
+                if (jenkins != null) {
+                    GitTool.DescriptorImpl gitTool = (GitTool.DescriptorImpl) jenkins.getDescriptor(GitTool.class);
+                    if (gitTool != null && gitTool.getGitDefaultTimeout() != null && gitTool.getGitDefaultTimeout() > 0) {
+                        timeout = gitTool.getGitDefaultTimeout();
+                    } else {
+                        timeout = TIMEOUT;
+                    }
+                }
+            }
+            listener.getLogger().println(" > " + command + (TIMEOUT_LOG_PREFIX + timeout));
```
MarkEWaite: Could you add one or more tests to assure that a new default timeout has been honored in the call to command line git?  [GitAPITestCase](https://github.com/jenkinsci/git-client-plugin/blob/master/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java#L110) includes several tests which check that the timeout values are being correctly echoed in the command line git output.

I would also be curious if setting the global timeout will apply to command line git calls like "git clean".  There are one or more requests to set a longer timeout for "git clean" when using a large repository over a slow file system.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
erikhakansson: This was just my IDE doing its own thing. I'll revert.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
erikhakansson: IDE again! :) I'll revert

erikhakansson:
I've reverted the import statements that was caused by my IDE and added a unit test that should cover the case where a timeout is configured. For when a timeout is not configured, any existing test should work.

erikhakansson:
@MarkEWaite any input on the added test?

MarkEWaite:
Thanks for the pull request.  

Could you make the integer value optional in the UI so that the red text demanding "Git default timeout must be an integer value" is not displayed when the field is empty?

Could you remove the word "description" from the UI?

git-client-timeout-setting

MarkEWaite:
My interactive testing seems to show that the default timeout setting is being applied in several cases (like multi-branch indexing) but not in other cases (like multi-configuration jobs, or even in other parts of a multi-branch pipeline job).

orkenstein:
@erikhakansson, BTW, how to change timeout now?

MarkEWaite:
@orkenstein the system wide default command line git timeout is `org.jenkinsci.plugins.gitclient.Git.timeOut` and can be set as a property on the Jenkins command line that starts the Jenkins process

orkenstein:
@MarkEWaite thanks! 
We are using Docker image. Could you, please, point out how to modify Docker container?

MarkEWaite:
I set the plugin timeout for my docker images from the `docker run` command line (part of my testing).

`--env JAVA_OPTS="-Dorg.jenkinsci.plugins.gitclient.Git.timeOut=11 -Duser.timezone=America/Denver"`

orkenstein:
So, the whole command gonna look like:
```
docker run --restart=always -d --env JAVA_OPTS="-Dorg.jenkinsci.plugins.gitclient.Git.timeOut=200" --name jenkins -p 8080:8080 -p 50000:50000 --volumes-from jenkins_bak jenkins
```

amontalban:
Would love to see this implemented! I was trying to use the multi branch pipeline with thousands of branches/tags and it was time outing.

Thanks!

MarkEWaite:
@amontalban I think you can already set the git clone timeout for multi-branch pipelines.  Add the "Advanced clone options" additional behavior and you'll find a clone timeout which can be set there.

If you are encountering timeouts during your clone within a Jenkinsfile, the `checkout scm` can be prototyped using the `Pipeline Syntax` link on most job definition pages.  Choose the `checkout` pick from the menu, then define the desired git options.

If your case is not solved by those techniques, then you're welcome to download and install a [prototype build](https://drive.google.com/open?id=0B9GyYWyfT1N5OGV0VEl2TTFWTEk) of the current master branch (just after release of git client plugin 2.3.0) with this change included.

amontalban:
Hey @MarkEWaite thanks for the reply, the problem in my case was that the "Scan Multibranch Pipeline" was time outing here:

```
Fetching & pruning origin...
Fetching upstream changes from origin
 > git --version # timeout=10
using GIT_SSH to set credentials 
 > git fetch --tags --progress origin +refs/heads/*:refs/remotes/origin/* --prune
Getting remote branches...
```

Because the timeout was 10 (Not sure why it didn't appear in the `git fetch` command), I tried to change the timeout for checkout but was not used, the solution was to change it globally with `-Dorg.jenkinsci.plugins.gitclient.Git.timeOut=120` so if we can set this value inside plugin would be great.

Thanks!

dggs123:
Hello everyone, I'm facing the same problem. I changed the git global time timeout using --env JAVA_OPTS="-Dorg.jenkinsci.plugins.gitclient.Git.timeOut=2000 in jnlp slave but while git push in jenkins pipeline job getting an timeout of 10 min. 

Here's the log
```
Pushing HEAD to branch refs/heads/master at repo deis
 > git --version # timeout=200
using GIT_SSH to set credentials 
 > git push ssh://<my-repo> HEAD:refs/heads/master -f
```

In git push there is no # timeout=200.

Thanks

laeubi:
This would really be helpfull if it could be configured in the jenkins screen, any progress on this issue? I try to use a jenkins pipeline script I can not adjust/modify that currently times out ...

laeubi:
BTW it might be usefull if the timeout choosen on the pipeline job would override the default for scm checkouts inside the script for scm checkouts of other repros also...

rgl:
@MarkEWaite, the close means this was already implemented?

MarkEWaite:
No @rgl, it means that I can't process it now.  It will need to wait until later.  Since I can't process it now, there is not much benefit to evaluating this pull request in the CI jobs until it gets closer to the top of the priority queue.

PR #223, opened 2016-11-23
Show merges in generated changelog

ptzafrir:


ptzafrir:
@MarkEWaite  Thank you for taking the time to look into this.
I added tests as you requested, and changed the behavior to be optional upon explicit request.
However my test to verify current behavior fails for JGitAPIImpl and passes for CliGitAPIImpl. Could you please provide any hints as to why is this?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -853,6 +856,10 @@ public void execute() throws GitException, InterruptedException {
                     throw new GitException("Error launching git whatchanged",e);
                 }
             }
+
+			public void listMerges(boolean flag) {
```
MarkEWaite: Please use consistent formatting (spaces vs. tabs) so that the alignment of text blocks in the file doesn't become any worse than it already is.  I realize that the text formatting is already inconsistent, but I would prefer to not have new inconsistencies added.

src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
```diff
@@ -121,4 +121,10 @@
      * ChangelogCommand instance or files will be left open.
      */
     void abort();
+
+    /**
+     * Include merge commits in the changelog
+     * @param flag
```
MarkEWaite: Please provide a few words of description of the purpose of flag so that the Javadoc processor with Java 8 will not add new javadoc warnings

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ptzafrir: Sure. I'm aware of that and fixed it in other places, missed that one

MarkEWaite:
I don't know why the test would fail on JGit and pass on command line git, unless the JGit implementation is not behaving as the test expects.  That may indicate that JGit is already showing merge commits (in its default configuration), while CliGit is not showing merge commits.  If that is the case, then I recommend making JGit consistent with CliGit, since CliGit is the reference implementation, the default implementation, and the most widely used implementation.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -2654,6 +2654,30 @@ public void test_merge_with_message() throws Exception {
...
+
+        // Merge branch1 into master
+        w.git.checkout("master");
+        String mergeMessage = "Merge message to be tested.";
+        w.git.merge().setMessage(mergeMessage).setGitPluginFastForwardMode(MergeCommand.GitPluginFastForwardMode.NO_FF).setRevisionToMerge(w.git.getHeadRev(w.repoPath(), "branch1")).execute();
+        // Obtain last commit message
+        String mergeSha1 = w.git.revParse("HEAD").name();
+        check_changelog_sha1(w.git.changelog().listMerges(true),mergeSha1, "master");
+        check_changelog_sha1(w.git.changelog().listMerges(false),commitSha1, "master");
+        check_changelog_sha1(w.git.changelog(),commitSha1, "master");
```
ptzafrir: @MarkEWaite  The above 2 tests fail with JGit and pass with CliGit.
The failure for this simple scenario is that JGit revision walker sees the HEAD commit (which is a merge commit). the ignore it becuase it has more then 1 parent (so far, as expected). But then when it should look for the previous commit in the history, it fails, as if there are no commits prior to the HEAD.
Whereas CliGit return the commit of "commit1" introduced in branch1 before the merge.
I think this is a JGit bug, and I don't quite understand the walker logic to provide further hints.
Was hoping you could shed some light on this.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ptzafrir: Fixed

ptzafrir:
I ended up removing the tests of the current expected behavior, as the current implementation of JGIT has different behaviour the CliGit.
I left only the test using the new optional value. Current changelog tests can be used to verify default behaviour was not modified.
Are there any more comments on the pull requests that should be addressed to allow the merge?

ptzafrir:
I have created [JENKINS-40023](https://issues.jenkins-ci.org/browse/JENKINS-40023) to track the discrepancy between JGit and CliGit which is not related to this pull request

ptzafrir:
Rewrite tests to workaround JGit bugs and provide tests for default behaviour as well.

ptzafrir:
@MarkEWaite  I believe I have addressed all your comments. Can you have another look at the latest state of the pull request and let me know if there is more work required in order to approve it?

ptzafrir:
ping. Please let me know if this merge request will not be merged, so I can work on creating a local fork with a patch for my use case which does require showing merge commits in jenkins changelog.
I'm waiting for this to be approved, to work on a second pull request to git-plugin which will utilize this new option.

ptzafrir:
Please respond to this merge request, If this functionality is not to be added by principle at least please say so so I can drop this and revert back to local modifications of the plug in upon each release as I need this functionality added

MarkEWaite:
I am currently evaluating a proposal to add git large file support to the plugin.  I won't do further evaluation of this pull request for at least a month.  Once I start the evaluation of a pull request, it typically takes 1-2 weeks before it is accepted into the plugin.

ptzafrir:
Any updates on this?

PR #224, opened 2016-11-23
Explicitly look up the submodule path in .gitmodules 

courtarro:
...rather than assuming that the path is the same as the module name. Should fix JENKINS-37495


KostyaSha:
@MarkEWaite JFYI you can optionally add PR template https://github.com/blog/2111-issue-and-pull-request-templates https://github.com/docker/docker/blob/master/.github/PULL_REQUEST_TEMPLATE.md

courtarro:
I'm working on a test, but it will require a modification to the tests/getSubmodules branch. What is the process for contributing a pull request with that kind of change? I will need to add a submodule whose name is different from its path.

MarkEWaite:
Submit it as two pull requests, one to the tests/getSubmodules branch and one to the master branch.

PR #227, opened 2016-12-08
Exclude patterns and timeout support for git clean

sboschman:
Added support for exclude patterns (-e option) and timeout for the git clean command (see JENKINS-29893 and JENKINS-39111). Support in the git-client-plugin is required before I can open a pull request with the follow up changes to the git-plugin and make the features available in the Jenkins UI.


src/main/java/org/jenkinsci/plugins/gitclient/GitClient.java
```diff
@@ -435,6 +435,12 @@
      */
     void clean() throws GitException, InterruptedException;
 
+    /**
+     * clean_.
+     *
+     * @return a {@link org.jenkinsci.plugins.gitclient.CleanCommand} object.
+     */
+    CleanCommand clean_(); // can't use 'clean' as legacy IGitAPI already define this method
```
KostyaSha: So name it as clean2 or anything other meaningful

sboschman: Not sure how 'clean2' is more meaningful. I'm just following the naming convention used in this class. Like clone_, fetch_, init_ and revList_.

E.g.:
FetchCommand fetch_(); // can't use 'fetch' as legacy IGitAPI already define this method

src/main/java/org/jenkinsci/plugins/gitclient/CleanCommand.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface CleanCommand extends GitCommand {
+    /**
+     * exclude patterns.
+     *
+     * @param excludePatterns a {@link java.util.List} with exclude patterns.
+     * @return a {@link org.jenkinsci.plugins.gitclient.CleanCommand} object.
+     */
+    CleanCommand excludePatterns(List<String> excludePatterns);
+
```
MarkEWaite: Could you add a method submodules(boolean removeSubmodules) to allow callers to specify that directories which contain a `.git` directory should also be cleaned?

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -1178,6 +1178,10 @@ void format(RevCommit commit, @Nullable RevCommit parent, PrintWriter pw, Boolea
         }
     }
 
+    public CleanCommand clean_() {
+    	throw new UnsupportedOperationException("No JGit support!");
```
MarkEWaite: Why not an implementation for JGit?  The [JGit CleanCommand](http://download.eclipse.org/jgit/site/4.6.0.201612231935-r/apidocs/org/eclipse/jgit/api/CleanCommand.html) seems like it could support path exclusion.

sboschman: The current implementation of JGit does not support exclusion, or I am overlooking something... The only option I see is an inclusion list for paths (exact match, no regex). Probably it is possible to calculate the diff and feed JGit the entire list of paths it is allowed to clean, but that sounds like a real ugly workaround for a missing option in JGit imo.

src/main/java/org/jenkinsci/plugins/gitclient/CleanCommand.java
sboschman: Yes, think I can add that option as well. Will look into it next week.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -670,15 +672,65 @@ public void execute() throws GitException, InterruptedException {
...
      * Remove untracked files and directories, including files listed
      * in the ignore rules.
      *
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
      * @throws java.lang.InterruptedException if interrupted.
      */
     public void clean() throws GitException, InterruptedException {
-        reset(true);
-        launchCommand("clean", "-fdx");
+    	clean_().execute();
```
MarkEWaite: I think this changes behavior from the previous implementation for tracked files.  I don't think we can change that behavior.

I believe that a tracked file which was modified would be reset to its unmodified value by the previous implementation of clean().  I'm surprised there is not a test which checks that behavior, but I am reasonably confident the previous behavior was to reset the contents of tracked files, in addition to cleaning the untracked files from the repository.

Is there a compelling reason you removed the reset(true)?

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
MarkEWaite: I had hoped it would be feasible to use the exclusion list as a way to generate the JGit inclusion list.  However, I can accept that isn't workable if constructing the that inclusion list is difficult or error-prone

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -828,6 +828,12 @@ public void test_clean() throws Exception {
         w.touch(fileName2);
         w.touch(fileName, "new content");
 
+        // JGit does not support exclude patterns        
+        if (w.igit() instanceof CliGitAPIImpl)
+        {
+        	((CliGitAPIImpl)w.igit()).clean_().excludePatterns(Collections.singletonList(fileName)).execute();
```
MarkEWaite: Is there a specific reason you didn't use w.cgit() instead of casting the w.igit() return value to a CliGitAPIImpl?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
sboschman: the reset is not removed, it is part of the execute method of the CleanCommand

ide puts in tabs, so that is not intentionally, should be fixed now

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
sboschman: I was under the impression cgit/jgit methods return a new GitClient object. Opposed to the igit method which returns the current workspace GitClient. So the reasoning was to add an extra test for the CliGitAPIImplTest run using the workspace client. And skip the extra test for a JGitAPIImplTest run.

PR #225, opened 2016-11-24
Correctly handle submodules whose name and path do not match

courtarro:
## Bug to fix
The Jenkins git-client plugin currently checks out submodules by initializing them into PATH, where PATH is erroneously specified as the *name* of the submodule. In many repositories, the submodule name and path match, but this is not a git requirement. When they do not match, this results in an error condition. [JENKINS-37495](https://issues.jenkins-ci.org/browse/JENKINS-37495)

## What I did
I added a new method named `getSubmodulePath()`, which takes the name of the submodule and fetches the `path` as defined in `.gitmodules`. I updated the existing submodule update code to call this method and get the path explicitly, rather than reusing the *name* of the submodule, which was mislabeled as `sUrl`. The old `sUrl` was renamed `sModuleName` to reflect its true nature.

Automated testing isn't going to like this one...

## How to demonstrate the bug in `master`
1. Clone `https://github.com/courtarro/git-client-plugin.git` into `target/clone.git`. This repository has an additional submodule that demonstrates the bug.
2. Run the tests (`mvn test`). You can run just the test in question with `mvn -Dtest=org.jenkinsci.plugins.gitclient.CliGitAPIImplTest test`
3. Observed the error that includes `error: pathspec 'puppetssh' did not match any file(s) known to git.`. This is the bug. Git-client tried to run `git submodule update puppetssh` when it should have run `git submodule update modules/sshkeys`

## How to test that the fix works
1. Check out this branch
2. Follow steps 1 and 2 above.
3. Observe that the tests succeed.

PR #230, opened 2016-12-21
Reference the .gitmodules file to get a list of current submodules

courtarro:
# Bug to fix

The Jenkins git client currently determines the list of submodules by reading `.git/config`. Unfortunately, this file may contain references and URLs for any submodules that have been accessed in the working directory of this repository, not simply the ones that exist in the active branch. In many cases this will result in Jenkins attempting to clone non-existent submodules and fail. [JENKINS-37419](https://issues.jenkins-ci.org/browse/JENKINS-37419)

# What I did

I explicitly instruct git to read the `.gitmodules` file when gathering its list of submodules to update. By iterating through `.gitmodules`, this ensures that only submodules for *this* branch are synchronized and updated. The regex string used is the same as the one a few lines later. Along with #225, this should fix some issues related to the new submodule update approach.

# How to test

There is currently no test that shows the previous bug. This would require a checkout of a branch with a one set of submodules, then switching to a branch with a non-common set of submodules. Since there is currently an effort to rewrite the submodule tests (by @MarkEWaite, I believe), I decided to forgo writing one of my own. If this would be useful, let me know.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -961,6 +961,10 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
              * @throws InterruptedException if called methods throw same exception
              */
             public void execute() throws GitException, InterruptedException {
+                // Initialize the submodules to ensure that the git config
+                // contains the URLs from .gitmodules.
+                submoduleInit();
+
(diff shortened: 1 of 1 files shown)


courtarro:
This doesn't pass tests because the bug referenced in #225 still exists.

MarkEWaite:
Thanks.  I've included it in a [new evaluation branch](https://github.com/MarkEWaite/git-client-plugin/commits/master-PR225-PR230-renamed-submodules) which combines PR225 and the submodule tests.

PR #231, opened 2016-12-21
Properly escape special Windows Batch characters

mletterle:
This was a blocking issue for us, so I went ahead and implemented the simplest fix. 

Related JIRA issue: https://issues.jenkins-ci.org/browse/JENKINS-40166

I looked into adding UnitTests for quoting, but it didn't seem straightforward and it's platform dependent.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1515,9 +1515,18 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
     private String quoteWindowsCredentials(String str) {
-        // Assumes the only meaningful character is %, this may be
-        // insufficient.
-        return str.replace("%", "%%");
+        // Quote special characters for Windows Batch Files
+        // See: http://ss64.com/nt/syntax-esc.html
+        String quoted = str.replace("%", "%%")
+                        .replace("^", "^^")
+                        .replace(" ", "^ ")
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1515,9 +1515,18 @@ private File createSshKeyFile(File key, SSHUserPrivateKey sshUser) throws IOExce
     }
 
     private String quoteWindowsCredentials(String str) {
-        // Assumes the only meaningful character is %, this may be
-        // insufficient.
-        return str.replace("%", "%%");
+        // Quote special characters for Windows Batch Files
```
mletterle: I was hesitant to change something like that just for testing purposes, but if it's thought to be worthwhile, I can certainly do so and then add some basic tests.

MarkEWaite: I'm absolutely willing to make methods package protected for tests.  The CliGitAPIImpl class is generally not accessed directly, rather it is accessed through either the GitClient interface or the GitAPI interface, so it is unlikely that someone will mistakenly place their new class into the same class as CliGitAPIImpl, and then use that package protected method.

I started an [evaluation branch](https://github.com/MarkEWaite/git-client-plugin/commits/master-PR231-escape-windows-chars) with some preliminary tests.  They are incomplete, and riddled with problems.  You're welcome to use those for ideas, or to discard them and create the tests that you think are most sensible.  I won't be able to return to that code for at least a week.

PR #234, opened 2017-01-23
Limit the number of objects returned from rev-list

deece:
Add a new method that allows a rev-list limit to be specified. This improves performance when fetching tags for multibranch projects.


jacob-keller:
Where would you envision using this? How does the count allow you to save time?

deece:
This was used here to only fetch the relevant commit for the tag rather than its full history.
https://github.com/jenkinsci/git-plugin/pull/407/commits/883e92715e1fec323d3c19fab5a9a34ec34a4664

I'm not sure if https://github.com/jenkinsci/git-plugin/pull/528 is susceptible to the same performance issues as the former was.

deece:
The PR which depended on this was dropped, so this is no longer required.

PR #236, opened 2017-03-21
[JENKINS-40834] Add an API to resolve symbolic refs

stephenc:
- Sadly JGit needs to try and guess

@reviewbybees

(Downstream PR for git-plugin to use the API will follow - based on WIP in https://github.com/jenkinsci/git-plugin/tree/jenkins-40834 that needed this change to have a hope of working)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2614,27 +2614,29 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     @Override
     public Map<String, String> getRemoteSymbolicReferences(String url, String pattern)
             throws GitException, InterruptedException {
-        ArgumentListBuilder args = new ArgumentListBuilder("ls-remote");
-        args.add("--symref");
-        args.add(url);
-        if (pattern != null) {
-            args.add(pattern);
-        }
-
-        StandardCredentials cred = credentials.get(url);
-        if (cred == null) cred = defaultCredentials;
-
-        String result = launchCommandWithCredentials(args, null, cred, url);
-
         Map<String, String> references = new HashMap<>();
-        String[] lines = result.split("\n");
-        Pattern symRefPattern = Pattern.compile("^ref:\\s+([^ ]+)\\s+([^ ]+)$");
-        for (String line : lines) {
-            Matcher matcher = symRefPattern.matcher(line);
-            if (matcher.matches()) {
-                references.put(matcher.group(2), matcher.group(1));
-            } else  if (line.length() < 41) {
-                throw new GitException("unexpected ls-remote output " + line);
+        if (isAtLeastVersion(2, 8, 0, 0)) {
+            // --symref is only understood by ls-remote starting from git 2.8.0
+            // https://github.com/git/git/blob/afd6726309/Documentation/RelNotes/2.8.0.txt#L72-L73
+            ArgumentListBuilder args = new ArgumentListBuilder("ls-remote");
+            args.add("--symref");
+            args.add(url);
+            if (pattern != null) {
+                args.add(pattern);
+            }
+
+            StandardCredentials cred = credentials.get(url);
+            if (cred == null) cred = defaultCredentials;
+
+            String result = launchCommandWithCredentials(args, null, cred, url);
+
+            String[] lines = result.split("\n");
+            Pattern symRefPattern = Pattern.compile("^ref:\\s+([^ ]+)\\s+([^ ]+)$");
+            for (String line : lines) {
+                Matcher matcher = symRefPattern.matcher(line);
+                if (matcher.matches()) {
+                    references.put(matcher.group(2), matcher.group(1));
+                }
             }
         }
         return references;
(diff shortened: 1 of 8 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2611,7 +2611,36 @@ public ObjectId getHeadRev(String url, String branchSpec) throws GitException, I
...
+        String result = launchCommandWithCredentials(args, null, cred, url);
+
+        Map<String, String> references = new HashMap<>();
+        String[] lines = result.split("\n");
+        Pattern symRefPattern = Pattern.compile("^ref:\\s+([^ ]+)\\s+([^ ]+)$");
+        for (String line : lines) {
+            Matcher matcher = symRefPattern.matcher(line);
+            if (matcher.matches()) {
+                references.put(matcher.group(2), matcher.group(1));
+            } else  if (line.length() < 41) {
```
MarkEWaite: If it helps, you can refer to my https://github.com/MarkEWaite/git-client-plugin/tree/master-PR236-resolve-symbolic-refs branch for some additional tests I added and for the `line.isEmpty()` change.

stephenc:
@MarkEWaite I have reworked this somewhat based on your feedback.

Now the API will return an empty map if the client or remote server does not support symref reporting.

Consumers will have to decide what they want to do in the absence of the symref information, which I think is better than trying to fake out the information in the Git 1.8.4 way

MarkEWaite:
Thanks @stephenc , I'll review it and post my feedback.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -784,6 +790,67 @@ public void deleteRef(String refName) throws GitException, InterruptedException
         return references;
     }
 
+    @Override
+    public Map<String, String> getRemoteSymbolicReferences(String url, String pattern)
+            throws GitException, InterruptedException {
+        Map<String, String> references = new HashMap<>();
+        String regexPattern = null;
+        if (pattern != null) {
+            regexPattern = createRefRegexFromGlob(pattern);
```
MarkEWaite: This call to createRefRegexFromGlob creates a regex pattern which won't match "HEAD".  Line 805 seems to mandate that only HEAD is accepted as a parameter to JGit for non-empty results, but if I pass "HEAD", the results are empty for JGit but have a value for CliGit.

Refer to e89785033c15b1fec3fcb4165757a966baffc74c for a test to show the problem, and to aa92d89e7e4fd7c90c01b7e1f0e6cc210cd72874 for one alternative to fix it.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -3372,6 +3374,27 @@ public void test_getRemoteReferences_withMatchingPattern() throws Exception {
...
+        assertThat(references, hasEntry(is(Constants.HEAD), is(Constants.R_HEADS + Constants.MASTER)));
+    }
+
+    protected abstract boolean hasWorkingGetRemoteSymbolicReferences();
+
+    /**
+     * Test getRemoteSymbolicReferences with listing all references
+     */
+    public void test_getRemoteSymbolicReferences_withMatchingPattern() throws Exception {
+        assumeThat(hasWorkingGetRemoteSymbolicReferences(), is(true));
```
MarkEWaite: Unfortunately, this file is JUnit 3.  It doesn't act correctly for an assumeThat.  Need to replace it with a simple conditional.  See the [test failures](https://ci.jenkins.io/job/Plugins/job/git-client-plugin/view/change-requests/job/PR-236/lastCompletedBuild/testReport/).

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -3372,6 +3374,27 @@ public void test_getRemoteReferences_withMatchingPattern() throws Exception {
         assertTrue(references.isEmpty());
     }
 
+    /**
+     * Test getRemoteSymbolicReferences with listing all references
+     */
+    public void test_getRemoteSymbolicReferences() throws Exception {
+        assumeThat(hasWorkingGetRemoteSymbolicReferences(), is(true));
```
MarkEWaite: Unfortunately, this file is JUnit 3.  It doesn't act correctly for an assumeThat.  Need to replace it with a simple conditional.  See the [test failures](https://ci.jenkins.io/job/Plugins/job/git-client-plugin/view/change-requests/job/PR-236/lastCompletedBuild/testReport/).

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -784,6 +790,67 @@ public void deleteRef(String refName) throws GitException, InterruptedException
...
+        }
+        // HACK ALERT... JGit doesn't give the info we require hard-coding HEAD symref support only based on pre 1.8.5
```
MarkEWaite: I think this comment is now outdated, isn't it?

The later comment describes the current implementation, while I think this comment describes something previous

MarkEWaite:
@stephenc , I think this is ready to merge, so long as you are OK with me adding the changes I've implemented in https://github.com/MarkEWaite/git-client-plugin/commits/master-PR236-resolve-symbolic-refs .  I'd merge this, then squash my changes into a more coherent set of a few commits.

Is that OK with you, or would you prefer to make changes based on my last review?

stephenc:
All fine by me

PR #239, opened 2017-04-05
JENKINS-43198: restart clean as long as we catch a false jgit failure

afeblot:
The previous fix was not ok: it could leave files or directories untouched instead of deleting them.


MarkEWaite:
Test change has been merged, code change is not included in the merge, will be included in the next JGit release after 1.10.0.

PR #240, opened 2017-04-17
Add getCredentials API - credentials tracking

MarkEWaite:
Callers of the git client API are best suited to track credential usage.  They are currently able to add credentials and clear credentials.  This allows callers to also get the credentials so that they can pass those credentials to CredentialsProvider.track().

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 
   <groupId>org.jenkins-ci.plugins</groupId>
   <artifactId>git-client</artifactId>
-  <version>2.4.2-SNAPSHOT</version>
+  <version>2.5.0-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
   <name>Jenkins Git client plugin</name>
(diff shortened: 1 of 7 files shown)


KostyaSha:
@stephenc do you have any doc for credentials fingerprinting? I.e. when password used in global settings -> cloud, custom global settings and etc.

MarkEWaite:
@KostyaSha , I used the CredentialsProvider.track() and CredentialsProvider.trackAll() [javadoc](http://javadoc.jenkins-ci.org/credentials/com/cloudbees/plugins/credentials/CredentialsProvider.html#track-hudson.model.Item-C-) as my first reference.

I'm still not convinced this API will actually be needed in the git client plugin, since the implementation in the git plugin [PR490](https://github.com/jenkinsci/git-plugin/pull/490) has not yet required it.

KostyaSha:
I saw this methods, but they are item related. And unclear how to untrack when configuration changed.

stephenc:
@MarkEWaite likely you have not needed it in the git plugins because of [findCredentialById(id, type, run, domainRequirements)](https://github.com/jenkinsci/credentials-plugin/blob/a21b0d37cd4f4c566b599ef0042d5986c92acf07/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java#L944) and some other similar helpers that automatically infer tracking of the credential as required.

@KostyaSha I have lobbied extensively for a time allocation to work on the credentials user, implementer and consumer guides... currently I am approved for [JENKINS-43426](https://issues.jenkins-ci.org/browse/JENKINS-43426) (and there are some potential aspects of that which I would like to discuss with you *after* I get the bulk of that sorted out... but that is a different story)... perhaps @HRMPW could advise on where he is putting the credentials user, implementer and consumer guides on the backlog ;-)

MarkEWaite:
I'm closing this pull request because the git plugin credential tracking support does not require it and I'm not aware of any others that currently require it.  We can reopen the pull request in the future if we find a consumer that needs the API.

PR #241, opened 2017-05-05
Added support for running "git lfs pull" in submodules.

creste:
This is a follow-up to #232 "Added simple git lfs support".

These changes add a new property called `lfsRemote` to `SubmoduleUpdateCommand`.  When `lfsRemote` is not null, Jenkins will automatically run `git lfs pull <lfsRemote>` after checking out each submodule.  I only implemented this functionality in the git CLI implementation.  I could not get JGit to work.

These changes depend on a new test branch called `tests/getLargeFileSubmodules`.  I have a copy of the branch at https://github.com/creste/git-client-plugin/tree/tests/getLargeFileSubmodules.  I'm not sure how to get that branch into this repository because it is an "orphaned" branch that doesn't share any history with existing branches in this repository.  I tried to make the branch as similar as possible to the `tests/getSubmodules` branch.  My new branch also makes https://github.com/MarkEWaite/jenkins-pipeline-utils.git a submodule because that repository has files stored using git LFS.

I added a few test cases to cover the basic operations.  I tried adding test cases for JGit to confirm that it wouldn't cause problems but I couldn't get JGit to checkout the submodule used in `tests/GetLargeFileSubmodules`.  I'm not sure if that was my fault or if it is a bug in JGit.

A separate PR for jenkinsci/git-plugin is forthcoming.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -955,6 +955,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             String  ref                            = null;
             Map<String, String> submodBranch   = new HashMap<>();
             public Integer timeout;
+            String lfsRemote;
 
             public SubmoduleUpdateCommand recursive(boolean recursive) {
                 this.recursive = recursive;
@@ -983,6 +984,11 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
             public SubmoduleUpdateCommand timeout(Integer timeout) {
                 this.timeout = timeout;
+                return this;
+            }
+
+            public SubmoduleUpdateCommand lfsRemote(String lfsRemote) {
+                this.lfsRemote = lfsRemote;
                 return this;
             }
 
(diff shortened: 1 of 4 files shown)


ntwerdochlib:
@MarkEWaite What is the status of the PR?  I too am looking for this functionality.

MarkEWaite:
@ntwerdochlib the pull request has not been evaluated yet.  You can help with the run time evaluation by downloading the [latest build of the pull request](https://ci.jenkins.io/job/Plugins/job/git-client-plugin/job/PR-241/12/artifact/target/git-client.hpi) and testing it in your environment.

You're also welcome to help with the code review portion of evaluating the pull request.  Additional people evaluating pull requests is a great help for me.

MarkEWaite:
I've pushed the tests/largeFileSupport branch to the jenkinsci/git-client-plugin repository with an extra submodule included.  The additional submodule was initially created at modules/other-large-files, then was moved with `git mv` to other-modules/large-files.  You should be able to adjust the tests to use the official repository now.

MarkEWaite:
While I've been trying to understand the code and its tests, I've added the tests/getLargeFileSubmodules branch to the jenkins-pipeline-utils repository.  The tests pass except for 1, which passes on my Linux machines and seems to fail consistently on my Windows machines.

The failure message is:

```
org.jenkinsci.plugins.gitclient.GitClientTest.testCLICheckoutSubmoduleWithoutLFSWhenLFSAvailable[git]

Incorrect LFS file contents in %TEMP%\junit4861134648612575104\junit8524699918793965232\modules\large-files\uuid.txt expected:<[version https://git-lfs.github.com/spec/v1
oid sha256:75d122e4160dc91480257ff72403e77ef276e24d7416ed2be56d4e726482d86e
size 33]> but was:<[5e7733d8acc94636850cb466aec524e4]>
```

My Windows machines are running git 2.13 with LFS installed by the git installer.  I suspect that I would see the same failure if I were running my tests on a Linux machine with a version of git that is newer than 2.9.  Since Debian 9 has released and includes git 2.11.0, there is at least one Linux distribution using a git version that includes LFS specific extensions.

creste:
Thank you for the feedback Mark.  I suspect the test failure is caused by an LFS cache being used on one of the systems.  Here is an example:

```
git clone https://github.com/MarkEWaite/jenkins-pipeline-utils
cd jenkins-pipeline-utils
git checkout tests/getLargeFileSubmodules
git submodule init
# Simulate the testCLICheckoutSubmoduleWithGitLFS test case
git submodule update
# Make sure LFS downloaded the file
cat modules/large-files/uuid.txt
5e7733d8acc94636850cb466aec524e4
# Clear submodules
git submodule deinit --all
git submodule init
# Simulate the testCLICheckoutSubmoduleWithoutLFSWhenLFSAvailable test case
GIT_LFS_SKIP_SMUDGE=1 git submodule update
cat modules/large-files/uuid.txt
5e7733d8acc94636850cb466aec524e4
```
Notice that the large file was "downloaded" even though we specified GIT_LFS_SKIP_SMUDGE=1.  Actually, the file was not downloaded but was retrieved from the local LFS cache stored in the `.git/modules/large-files/lfs` directory.  This means GIT_LFS_SKIP_SMUDGE=1 doesn't actually skip the smudge.  The smudge still runs but will not download large files that aren't in the local cache.

After reviewing the differences between git 2.13 and git 2.10, it looks like these changes are no longer necessary.  In git 2.10, each large file was checked out using a separate process. In git 2.13, git now uses a single process to checkout all large files.  This significantly reduces the time it takes to clone large files.  Therefore, there is no need to force GIT_LFS_SKIP_SMUDGE=1 and then run "git lfs pull" after the submodule is cloned.

Technically, this pull request might help people still on Git 2.10 or prior, but I think most people will be upgrading over time so these changes will eventually decay anyhow.

PR #242, opened 2017-05-09
[JENKINS-43022] - Solved to exclude & include path to url

jsanchezramos:
Solution to work exclude & include path, is the issue JENKINS-43022


MarkEWaite:
Superseded by #249

PR #243, opened 2017-05-18
Fix workspace path regressions

MarkEWaite:
The security fixes in git client plugin 2.4.4 shifted from preferring the system temporary directory (with default permissions on the files) to preferring a temporary directory adjacent to the workspace directory.  That change exposed the git client plugin credentials code (command line git, with all its complications) to the path name of the workspace, where previously the credentials code only used the path name to the system temporary directory.

This fixes the following bugs:

* [JENKINS-44041](https://issues.jenkins-ci.org/browse/JENKINS-44041) - Windows authenticated git checkout fails if '(' or ')' in path to workspace
* [JENKINS-43931](https://issues.jenkins-ci.org/browse/JENKINS-43931) - Windows authenticated git checkout fails if ' ' in path to workspace
* [JENKINS-44127](https://issues.jenkins-ci.org/browse/JENKINS-44127) - Authenticated git checkout fails if '%' in path to workspace (Windows & Linux)
* [JENKINS-44301](https://issues.jenkins-ci.org/browse/JENKINS-44301) - Authenticated git checkout fails if '%' in path to workspace (Windows & Linux) 

This has completed initial testing in my automated tests, and in my interactive tests, and will run through another 1-2 days of interactive testing before I actually release it.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPITempFileTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPITempFileTest.java
@@ -1,0 +1,123 @@
+package org.jenkinsci.plugins.gitclient;
+
+import hudson.EnvVars;
+import hudson.model.TaskListener;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.jvnet.hudson.test.Issue;
+
+/**
+ * Test that createTempFile is adapting its directory name choices to match
+ * platform limitations of command line git.
+ *
+ * @author Mark Waite
+ */
+@RunWith(Parameterized.class)
+public class CliGitAPITempFileTest {
+
+    private final String workspaceDirName;
+    private final boolean mustUseSystemTempDir;
+    /* Should temp folder be in same parent dir as workspace? */
+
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPITempFileTest.java
```diff
@@ -0,0 +1,123 @@
...
+    @Rule
+    public TemporaryFolder workspaceParentFolder = new TemporaryFolder();
+
+    public CliGitAPITempFileTest(String workspaceDirName, boolean mustUseSystemTempDir) {
+        this.workspaceDirName = workspaceDirName;
+        this.mustUseSystemTempDir = mustUseSystemTempDir;
+    }
+
+    @Parameterized.Parameters(name = "{0}")
+    public static Collection workspaceDirNames() {
```
varyvol: This is not used anywhere?

MarkEWaite: That is a parameterized JUnit test.  The Parameterized annotation causes the JUnit executor to call this method to assemble the arguments to the constructor for this test class.

It allows me to readily construct many different cases which all use the same assertion.  This particular case runs about 10 tests, one for each of the different workspace directory names in the workspaceDirNames method

varyvol: Didn't know. Thanks for the explanation!

MarkEWaite:
@jglick I fully intend to remove as much of the SSH_ASKPASS "mess" as I can in a future change.

jmellor:
I can confirm that this git-client.hpi fixes JENKINS-44301 (2.4.5 bad regression: unable to handle previously-acceptable branch names)

flavius:
Original reporter of [JENKINS-44420](https://issues.jenkins-ci.org/browse/JENKINS-44420) here, I confirm it fixed it.

PR #244, opened 2017-06-06
Implement RECURSIVE_THEIRS merge strategy option

MacThrawn:
Hi,

as far as I know the current implementation of merge command does not support the parameter strategy-option for recursive merge strategy. I need to use recursive merge with the option theirs. Because JGit does not support the full set of features I implemented this feature as simple strategy RECURSIVE_THEIRS to support CLI and JGit implementation.  

Regards
Martin

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -622,7 +622,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                         args.add(strategy);    
                     }
                 }
-                
+
                 args.add(fastForwardMode);
                 args.add(rev.name());
                 launchCommand(args);
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -575,7 +575,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 this.strategy = strategy.toString();
                 return this;
             }
-            
+
             public MergeCommand setGitPluginFastForwardMode(MergeCommand.GitPluginFastForwardMode fastForwardMode) {
                 this.fastForwardMode = fastForwardMode.toString();
                 return this;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -575,7 +575,7 @@ public MergeCommand setStrategy(MergeCommand.Strategy strategy) {
                 this.strategy = strategy.toString();
                 return this;
             }
-
+
```
MarkEWaite: Please remove this pure white space change.  Formatting is already bad enough in the plugin without introducing more lines with trailing white space.

src/main/java/org/jenkinsci/plugins/gitclient/MergeCommand.java
```diff
@@ -34,8 +34,8 @@
     MergeCommand setStrategy(Strategy strategy);
 
     public enum Strategy {
-        DEFAULT, RESOLVE, RECURSIVE, OCTOPUS, OURS, SUBTREE;
-
+        DEFAULT, RESOLVE, RECURSIVE, OCTOPUS, OURS, SUBTREE, RECURSIVE_THEIRS;
+
```
MarkEWaite: Please remove the trailing white space introduced on this line

src/test/java/org/jenkinsci/plugins/gitclient/MergeCommandTest.java
```diff
@@ -225,6 +239,12 @@ public void testRecursiveStrategy() throws GitException, InterruptedException {
         assertTrue("README 1 missing on master branch", readmeOne.exists());
     }
 
+    @Test
+    public void testRecursiveTheirsStrategy() throws GitException, InterruptedException {
+        mergeCmd.setStrategy(MergeCommand.Strategy.RECURSIVE_THEIRS).setRevisionToMerge(commit1Branch2).execute();
+        assertTrue("branch 2 commit 1 not on master branch after merge", git.revList("master").contains(commit1Branch2));
```
MarkEWaite: Thanks for writing the test to confirm that the strategy option is accepted.

Would you be willing to include an assertion in the test which confirms that the behavior of the THEIRS option was as expected?  For example, could you assert that the contents of the merged file were as expected?

PR #248, opened 2017-06-30
Use parent pom 2.31

MarkEWaite:
Include requireUpperBoundDeps exclusions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,7 +116,7 @@
     <dependency>
       <groupId>org.jenkins-ci.modules</groupId>
       <artifactId>sshd</artifactId>
-      <version>1.7</version>
+      <version>1.11</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -272,27 +272,13 @@
         <artifactId>maven-enforcer-plugin</artifactId>
         <version>1.4.2.jenkins-1</version>
         <configuration>
-          <rules combine.children="append">
+          <rules>
             <requireUpperBoundDeps>
-              <excludes>
-                <!-- TODO Stapler requests 14 while Jenkins requests 11; should backdate the Stapler version -->
-                <exclude>com.google.guava:guava</exclude>
-                <!-- TODO jenkins-test-harness-htmlunit requests 1.2, newer than 1.1.x in core -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -267,6 +267,37 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>1.4.2.jenkins-1</version>
+        <configuration>
+          <rules combine.children="append">
+            <requireUpperBoundDeps>
+              <excludes>
+                <!-- TODO Stapler requests 14 while Jenkins requests 11; should backdate the Stapler version -->
+                <exclude>com.google.guava:guava</exclude>
```
jglick: This is defined in the parent. You should not repeat it here. `combine.children` works when you actually put it in the right place—on `excludes`.

pom.xml
```diff
@@ -267,6 +267,37 @@
...
+                <!-- TODO jenkins-test-harness-htmlunit requests 1.2, newer than 1.1.x in core -->
+                <exclude>commons-logging:commons-logging</exclude>
+                <!-- Stapler requests 2.x while core requests 1.x; anyway not used at runtime -->
+                <exclude>com.google.code.findbugs:jsr305</exclude>
+                 <!-- needed between Jenkins 2.36—2.60 -->
+                <exclude>org.kohsuke:access-modifier-annotation</exclude>
+                 <!-- core requests 1.4, sshd requests 1.2 -->
+                <exclude>org.jenkins-ci.modules:instance-identity</exclude>
```
jglick: Rather than excluding this violation, fix it, perhaps by upgrading `sshd-module` appropriately.

pom.xml
```diff
@@ -267,6 +267,37 @@
...
+                 <!-- core requests 1.4, sshd requests 1.1 -->
+                <exclude>org.jenkins-ci.modules:ssh-cli-auth</exclude>
+                 <!-- core requests 1.8.4, lib-jenkins-embedder requests 1.9.2 -->
+                <exclude>org.apache.ant:ant</exclude>
```
jglick: No. `jenkins-test-harness` has been fixed to not need `lib-jenkins-embedder`. If you are specifically requesting an older version, stop.

PR #253, opened 2017-07-21
[JENKINS-45729] fix getRepository() to work with symref git files

jacob-keller:
The current implementation of getRepository() fails to work if the
repository directory in question has a .git file (known as a symref)
which points to the real directory. This causes problems if you try to
directly open a submodule or worktree repository using the git client.

This happens because FileRepositoryBuilder.create() uses .setGitDir() to
set the git directory which expects only a proper git directory path.
Instead, set the worktree and have the RepositoryBuilder figure out
where the git directory is.

This implementation is identical to the JGit implementation, which is
a nice bonus.

Add additional tests for both CliGit and JGit which verify that a git
client can open a repository object properly for repositories
with .git symrefs.

Without this fix, the git plugin will fail to fetch and checkout
revisions of a working copy which use .git symref file to point to
its git meta directory, because it uses the withRepository in
order to obtain the git commit subject.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -2120,6 +2120,35 @@ public abstract class GitAPITestCase extends TestCase {
         assertSubmoduleContents(w.repo);
     }
 
+    /* Opening a git repository in a directory with a symbolic git file instead
+     * of a git directory should function properly. JGit does not implement
+     * renamed submodules, so we the submodules branch won't work with it...
+     */
+    @NotImplementedInJGit
+    public void test_with_repository_works_with_submodule() throws Exception {
+        w = clone(localMirror());
+        assertSubmoduleDirs(w.repo, false, false);
+
+        /* Checkout a branch which includes submodules (in modules directory) */
+        String subBranch = "tests/getSubmodules";
+        String subRefName = "origin/" + subBranch;
+        w.git.checkout().ref(subRefName).branch(subBranch).execute();
+        w.git.submoduleUpdate().recursive(true).execute();
+
+        /* Get a client directly on the ntp submodule */
+        GitClient ntpClient = setupGitAPI(new File(w.repo, "modules/ntp"));
+
+        /* Assert that when we invoke the repository callback it gets a
+         * functioning repository object
+         */
+        ntpClient.withRepository(new RepositoryCallback<Void>() {
+            public Void invoke(final Repository repo, VirtualChannel channel) throws IOException, InterruptedException {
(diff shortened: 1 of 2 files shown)


jacob-keller:
For reference, I believe this is also fixable in JGit by the following https://git.eclipse.org/r/#/c/101749/3 but since we already have a working pattern in JGitImpl, I think this is worth doing regardless.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -2120,6 +2120,35 @@ public void test_submodule_checkout_simple() throws Exception {
         assertSubmoduleContents(w.repo);
     }
 
+    /* Opening a git repository in a directory with a symbolic git file instead
+     * of a git directory should function properly. JGit does not implement
+     * renamed submodules, so we the submodules branch won't work with it...
+     */
+    @NotImplementedInJGit
```
MarkEWaite: I think this test can be implemented for JGit as well as for CliGit.  See the next comment for my assumption of how it can be implemented for JGit.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -2120,6 +2120,35 @@ public void test_submodule_checkout_simple() throws Exception {
...
+    public void test_with_repository_works_with_submodule() throws Exception {
+        w = clone(localMirror());
+        assertSubmoduleDirs(w.repo, false, false);
+
+        /* Checkout a branch which includes submodules (in modules directory) */
+        String subBranch = "tests/getSubmodules";
```
MarkEWaite: A JGit compatible submodule test branch is available if you use:

`String subBranch = w.git instanceof CliGitAPIImpl ? "tests/getSubmodules" : "tests/getSubmodules-jgit";`

When I do that, the test fails for JGit without your change, but also fails for JGit with your CliGitAPIImpl change.  I don't yet understand why it fails, since (as you noted) the CliGitAPIImpl and JGitAPIImpl versions of getRepository() are the same once your change is applied.

MarkEWaite: I'm exploring further and I think that the preceding test (`test_submodule_checkout_simple`) can be extended to cover both CLI git and JGit and your case.  It appears to be missing a call to submoduleInit() in order to work with JGit and the tests/getSubmodules-jgit branch.

MarkEWaite:
I submitted a [pull request](https://github.com/jacob-keller/git-client-plugin/pull/1) to extend your proposed change to test JGit as well.

jacob-keller:
Thanks for the help with the test.

PR #255, opened 2017-07-27
[JENKINS-45771] Ensure all command boolean options can be toggled, not just set once

stephenc:
Precursor to [JENKINS-45771](https://issues.jenkins-ci.org/browse/JENKINS-45771), we need the ability
to turn off shallow if another extension has turned it on.

@reviewbybees


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2369,6 +2369,7 @@ public RevListCommand all(boolean all) {
                 this.all = all;
                 return this;
             }
+
```
MarkEWaite: I know this is a silly request, but could you remove the trailing white space from that line?

There are so many formatting issues in the git client plugin and the git plugin that I'd like to avoid adding one more.

stephenc:
I thought I had done prune... :roll eyes:

PR #257, opened 2017-08-13
Disable code coverage unless COVERAGE=true passed to build

MarkEWaite:
Code coverage is used only occasionally, and typically from specific
freestyle jobs which report the coverage through graphs and annotated
source code.

Don't spend time or energy instrumenting for coverage when it won't be
used.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -183,37 +183,6 @@
         </configuration>
       </plugin>
       <plugin>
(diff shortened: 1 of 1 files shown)


MarkEWaite:
Thanks for the suggestion in the matching git plugin pull request (use the parent pom).  That will make this pom simpler and still provide coverage reporting

PR #260, opened 2017-08-17
Use jgit 4.5.3

MarkEWaite:
Fix exception handling for opening bitmap index files.

Unlikely to be detected in common Jenkins usage, but better to run the
current released version for those cases where it might be found.

See https://projects.eclipse.org/projects/technology.jgit/releases/4.5.3

@reviewbybees


oleg-nenashev:
Belated 🐝

PR #262, opened 2017-08-20
Update dependencies

MarkEWaite:
Dependency updates
* Apache httpclient 4.5.2 to 4.5.3, most recent httpclient
* parent pom 2.32 to 2.33, most recent parent pom
* structs 1.9 to 1.10, match with git plugin, latest dependency updates in structs)
* credentials 2.1.14, most recent and preferred credentials plugin

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
-      <version>4.5.2</version>
+      <version>4.5.3</version>
     </dependency>
     <dependency>
       <groupId>commons-codec</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>structs</artifactId>
-      <version>1.9</version>
+      <version>1.10</version>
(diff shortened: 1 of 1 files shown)


MarkEWaite:
@KostyaSha my thinking was that it is better to depend on the credentials version which the plugin maintainer has stated meets the criterion "I do not want to have to restart my Jenkins in the next two weeks" (the strongest of the criteria in [the list](https://wiki.jenkins.io/display/JENKINS/Credentials+Plugin)).

I'm fine leaving it at the current 2.1.13 value, though it seems surprising to me to leave it on an outdated version when the 2.1.13 is no longer in the table on the credentials plugin page at all.

Should the same logic be applied to structs 1.10 as well?

Can you provide some insights on your preference for dependency updates, and the reasons for those preferences?  That will help me better understand the use cases I should be considering.

KostyaSha:
Dependencies should be updated only when it critically required - new methods, new ABI. I don't remember anything critical and complex in structs plugin, while credentials plugin may not work with any stable by somebody opinion version in any jenkins. If somebody wants new version of credentials plugin - it should update credentials plugin, plugin maintainer should bother about his plugin, not about other.

MarkEWaite:
Good rationale.  I'll close this pull request and submit a new pull request with just the parent pom update.

PR #263, opened 2017-08-21
Update parent pom and structs dependency

MarkEWaite:
* Update parent pom from 2.32 to 2.33
* Update structs dependency from 1.9 to 1.10 to match git plugin and reduce risk of transitive dependency problems for other git client consumers

@reviewbybees


MarkEWaite:
Not needed, since the callers can choose (on their own) to use a newer structs dependency.

PR #264, opened 2017-08-24
commons-codec is provided by jenkins-core

jglick:
I found that `workflow-job`, with a `test`-scoped dep on `git`, had `requireUpperBoundDeps` violations whereby `git-client` was asking for a newer version of `commons-codec` than `jenkins-core` actually bundled.

Since the dep was `provided`, it was not being bundled in `git-client.hpi` anyway (and even if it were, without `pluginFirstClassLoader` it would be ignored).

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,8 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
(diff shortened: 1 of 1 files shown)


jglick:
> Jenkins uses older version

Precisely the point; see the PR description.

@reviewbybees done

PR #269, opened 2017-09-05
Assert exception if no changelog destination specified

MarkEWaite:
If the caller failed to provide a changelog destination, then the JGit and CliGit implementations handled the failure differently.  The CliGit implementation is considered authoritative, so the JGit implementation was extended to match CliGit.

@reviewbybees


MarkEWaite:
Infrastructure problem caused a test failure.  Closing and reopening to execute tests again.

PR #272, opened 2017-09-11
[JENKINS-46737] Improve logging when linking a reference repository.

Dohbedoh:
@reviewbybees @MarkEWaite Adding logging to show that the clone points to a reference repository

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
@@ -1423,10 +1423,12 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                             else {
                                 try {
                                     File alternates = new File(workspace, ".git/objects/info/alternates");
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -1423,10 +1423,12 @@ else if (!referencePath.isDirectory())
                             else {
                                 try {
                                     File alternates = new File(workspace, ".git/objects/info/alternates");
+                                    String absoluteReference = objectsPath.getAbsolutePath().replace('\\', '/');
+                                    listener.getLogger().println("Link to the reference repository " + absoluteReference);
```
MarkEWaite: I don't understand why you're reporting the location of the objects directory to the user.  The user passed in the location of the reference repository.  Won't they be perplexed when the value they passed is not displayed in the output?

Wouldn't it be better to display the reference repository, rather than the path to the objects directory in the reference repository?

PR #273, opened 2017-09-20
Protect against ArrayIndexOutOfBoundsException (related to JENKINS-34309)

bmleite:
If `git branch -v --no-abbrev` output contains a line like " very-long-string-with-more-than-44-characters\n", it will bypass the first `if` condition of the `while` block and will end up being split. However, since it will only contain 1 token (as there are no spaces after applying `substring(2)`), accessing `branchVerboseOutput[1]` results in ArrayIndexOutOfBoundsException. This fix is to avoid such exception and pre-validate the array size (should be greater than 1 in order to contain, at least, the branch name and the SHA1).


bmleite:
Just added an extra step to remove CR characters from the command output (to avoid any false positive branch identifications).
Also added a test case for this particular issue.

PR #276, opened 2017-10-10
[JENKINS-47373] Set git configuration to avoid test failures.

fcojfernandez:
[JENKINS-47373](https://issues.jenkins-ci.org/browse/JENKINS-47373)

If tests are executed without setting local/global git configuration (i.e. neither .git/config either ~/.gitconfig correctly set), tests in class org.jenkinsci.plugins.gitclient.JGitApacheAPIImplTest fail. That configuration should be set before executing the test. (See JIRA for more details)

@reviewbybees specially @MarkEWaite


varyvol:
🐝

PR #279, opened 2017-10-16
Add path filter to ChangeLogCommand

ansig:
This commits adds the possibility to restrict changelogs to certain
paths. It is related to adding support in the git-plugin to only
show checkouts for checked out paths (using the SparseCheckout
extension), which is implemented in: https://github.com/ansig/git-plugin/commit/7e6de1ee3c9c2e7442cc4b59e61fd4ce01073ff8

See [JENKINS-24461] for some more information on this requirement.


ansig:
@MarkEWaite Any chance of getting this merged? We would appreciate it so that we don't have to build our own version of this plugin and I think there are others as well: https://issues.jenkins-ci.org/browse/JENKINS-24461

There is a corresponding commit in git-plugin that uses this feature and I was waiting with making a pull request there since it requires this to be released first.

The change is made on master, which is now 3.0.0-beta6-SNAPSHOT. I do not know how far off that is to be released, though. Perhaps the change should be included in stable-2.7 as well?

PR #282, opened 2017-10-30
Include the results of the command in the log when it fails

trejkaz:
Proposed fix for JENKINS-47493.

Having the error in the exception is nice assuming someone can figure out where the exception renders, but most views of stage logs in Jenkins don't include the exception result at all. This patch causes it to log the stdout/stderr and status code to the log for the stage, so that it can be seen inline when it occurs.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1967,7 +1967,9 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
             String result = fos.toString(Charset.defaultCharset().toString());
             if (status != 0) {
-                throw new GitException("Command \""+command+"\" returned status code " + status + ":\nstdout: " + result + "\nstderr: "+ err.toString(Charset.defaultCharset().toString()));
+                String message = "Command \"" + command + "\" returned status code " + status + ":\nstdout: " + result + "\nstderr: " + err.toString(Charset.defaultCharset().toString());
+                listener.getLogger().println(message);
+                throw new GitException(message);
             }
 
             return result;
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -377,6 +377,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
                 warnIfWindowsTemporaryDirNameHasSpaces();
 
+                //TODO: Catch and log here?
                 launchCommandWithCredentials(args, workspace, cred, url, timeout);
             }
         };
@@ -411,6 +412,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
         StandardCredentials cred = credentials.get(url);
         if (cred == null) cred = defaultCredentials;
+        //TODO: Catch and log here?
         launchCommandWithCredentials(args, workspace, cred, url);
     }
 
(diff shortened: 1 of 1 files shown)


trejkaz:
What's the right place to put it then?

MarkEWaite:
I think it needs to be placed in the exception handling of the specific call that defined the git command that was called.

Logging the failures of all commands is not an improvement as far as I can tell. It will continually distract users with failure messages which they cannot resolve, and risk obscuring relevant messages in the noise of the irrelevant.

trejkaz:
Compared to the current behaviour, though, where you can't even figure out what went wrong because none of the information is printed at all.

MarkEWaite:
> where none of the information is printed at all

might be added "in the specific use case that you've seen", and might further be added with "ignoring all the other use cases of the 130,000+ installations of the git client plugin".

trejkaz:
I have seen about 3 separate cases of it so far, but it's okay, keep pretending your shit isn't broken.

I'm gathering a list of all the places launchCommandIn is called, but it's going to result in a lot of duplicate code at this rate because it looks like it's called a *lot*.

trejkaz:
The current call hierarchy is as follows. The three marked as **SEEN** have appeared as the last line in our build logs for stages which supposedly failed despite no error being printed in the log at the time, so those three are a given, but maybe there are others which could be problematic as well?.

    * `launchCommandIn` called from:
        * `#getRemoteTagNames`
        * `#getTagNames`
        * **SEEN** anonymous `CheckoutCommand#execute`
        * `#launchCommand`, which in turn is called from:
            * `#getGitVersion`
            * `#hasGitRepo`
            * `#reset`
            * `#clean`
            * `#revParse`
            * `#validateRevision`
            * `#describe`
            * `#showRevision`
            * `#submoduleInit`
            * `#addSubmodule`
            * `#submoduleSync`
            * `#submoduleReset`
            * `#submoduleClean`
            * `#getSubmoduleUrl`
            * `#setSubmoduleUrl`
            * `#getSubmodulePath`
            * `#getRemoteUrl`
            * `#setRemoteUrl`
            * `#addRemoteUrl`
            * `#getRemoteUrl`
            * `#setRemoteUrl`
            * `#getDefaultRemote`
            * `#isBareRepository`
            * `#tag`
            * `#createNote`
            * `#getBranches`
            * `#tagExists`
            * `#deleteBranch`
            * `#deleteTag`
            * `#lsTree`
            * `#add`
            * `#branch`
            * `#commit`
            * `#getTagMessage`
            * `#ref`
            * `#refExists`
            * `#deleteRef`
            * `#getRefNames`
            * `#merge`
            * `#getBranchesContaining`
            * `#mergeBase`
            * `#getAllLogEntries`
            * anonymous `CloneCommand#execute`
            * **SEEN** anonymous `MergeCommand#execute`
            * anonymous `RebaseCommand#execute`
            * anonymous `InitCommand#execute`
            * anonymous `SubmoduleUpdateCommand#execute`
            * anonymous `RevListCommand#execute`
        * `#launchCommandWithCredentials`, which in turn is called from:
            * `#fetch`
            * `#prune`
            * `#getHeadRev`
            * `#getRemoteReferences`
            * `#getRemoteSymbolicReferences`
            * `#push`
            * **SEEN** anonymous `FetchCommand#execute`
            * anonymous `PushCommand#execute`

trejkaz:
Well... I might drop this change request anyway.

It did help find the real error behind some operations which seemed to be passing, but now we're at a point where we want to see the output of previous "successful" commands, to figure out why they might have left the repo locked.

So really now I'd want to see all the output somehow, meaning that maybe what we really want is a global verbose setting, rather than just printing the output on error.

MarkEWaite:
I'm closing this pull request in favor of future investigation of adjustable logging level for both successful and failing command line git calls.

PR #283, opened 2017-11-02
use --no-walk when merely checking if a revision exists

jacob-keller:
When implementing isCommitInRepo() for the CLI git implementation we use
git rev-list, assuming that if this command fails then the commit
doesn't exist in the repository. This is technically correct.

However, git rev-list actually performs a complete revision walk, which
for large repositories can take a long time. For example, on the Linux
repository, git rev-list HEAD took around 30 seconds. This is entirely
due to the command producing an entire list of all revisions leading
back to the beginning of history.

But isCommitInRepo() doesn't even care about that! We can significantly
reduce the cost of this function in a number of ways. One, we could
re-implement this using various other git commands, such as git
cat-file. Or, we could ask rev-list not to actually walk the revisions.
This was added in git 1.5.3 (~2007).

Local testing found there to be minimal difference in "git cat-file -t
<id>" vs "git rev-list --no-walk <id>". Since the latter is mostly
already implemented, add support for "--no-walk" to the
RevListCommand.

This allows us to modify isCommitInRepo() to use the RevList_()
directly, and pass .nowalk(true) in order to avoid the unnecessary
revision walk.

For completeness, also implement the equivalent support in JGit. To do
so, simply use parseCommit() of the refspec provided to the command. If
all was set to true, we'll search all refs and include those as well.

Ultimately, this enables the command line implementation of the
gitClient to have a significantly faster test for whether a commit is in
a repository.

Signed-off-by: Jacob Keller <jacob.e.keller@intel.com>

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2416,6 +2416,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     public RevListCommand revList_() {
         return new RevListCommand() {
             public boolean all;
+            public boolean nowalk;
             public boolean firstParent;
             public String refspec;
             public List<ObjectId> out;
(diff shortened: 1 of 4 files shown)


jacob-keller:
Found this when debugging an apparent stall in changelog calculation. It baffled me that we'd perform an entire graph walk just to determine if a commit is in a repository. On large repositories with a long history this can change the call from taking ~30s or longer to milliseconds. This also vastly reduces the memory consumption that would occur with actually copying the entire rev-list for a big project.

jacob-keller:
It should hopefully show a nice speedup of the change log creation, which is where isCommitInRepo() is used.

PR #284, opened 2017-11-03
Fix submodule url regex [JENKINS-46054]

MarkEWaite:
[JENKINS-46054](https://issues.jenkins-ci.org/browse/JENKINS-46054) reports that submodules whose URL includes the text `.url` cannot be cloned or updated into a Jenkins workspace.

The `git config` output parsing regular expression incorrectly extended across a space character when a space character is not valid in the name of a git remote.

This pair of changes writes tests for the problem and fixes the problem.


jacob-keller:
This seems correct to me.

PR #285, opened 2017-11-09
Fixes issue with submodules url not found when starting  from fresh workspace clones

TheSkorm:
Since `getSubmodulePath` uses `-f .gitmodules` we should probably do the same on `getSubmoduleUrl`. This resolves the issue listed here https://issues.jenkins-ci.org/browse/JENKINS-38860

I'm not a java dev, or a Jenkins dev so this PR is provided AS IS and changes "Works for me".

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1181,7 +1181,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * Get submodule URL
      */
     public @CheckForNull String getSubmoduleUrl(String name) throws GitException, InterruptedException {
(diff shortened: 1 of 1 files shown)


TheSkorm:
https://issues.jenkins-ci.org/browse/JENKINS-41553 this should be implemented instead of this change, but this is a nice workaround

MarkEWaite:
Tests to confirm behavior has improved?

TheSkorm:
Hi @MarkEWaite,

I'm not experienced enough to even know where to start writing tests for this module nor could I seem to get tests to run locally on my machine.

TheSkorm:
Hi @MarkEWaite  This issues still seems to be catching people out (https://issues.jenkins-ci.org/browse/JENKINS-38860). Just wondering if there is a work around (apart from this branch?)

I still can't comment in JIRA for some reason.

MarkEWaite:
No, @TheSkorm there is not a workaround as far as I can tell.

PR #288, opened 2017-12-02
[JENKINS-45447] Add GitClient.getTags()

MarkEWaite:
[JENKINS-45447](https://issues.jenkins-ci.org/browse/JENKINS-45447) reports that checkout of a freestyle  job with a wildcard in the branch name (and many other cases which invoke a more advanced branch selection mechanism) with many tags in the repository is dramatically slower than earlier releases.

[31fedce9c](https://github.com/jenkinsci/git-plugin/commit/31fedce9c41c9006c886835e03c9fe825d49aba6) commit added tag checks to the evaluation loop for branch names.  Unfortunately, tag evaluation needs both the tag name and the SHA1 of the commit identified by the tag.  The implementation used revParse() to compute that SHA1.  With many tags in the repository (bug report example was 60,000 tags), the checkout time was significant.

I found that even the number of tags in the git plugin repository could add as much as 5 seconds for the computation of SHA1 hashes which were then discarded very soon after the evaluation.

This commit adds a GitClient.getTags() method which collects all tag/SHA1 pairs in the current repository using a single git call.  That removes the per-call overhead for many rev-parse calls and is hoped to be sufficient to dramatically improve performance.  The git plugin will call that new method instead of calling GitClient.getTagNames() and then calling GitClient.revParse() for each tag name.

This pull request also makes the fields of the GitObject final.  They were previously package protected and could only be assigned through the constructor.

This pull request also adds an equals() and hashCode() method to GitObject, Tag, and Branch objects so that they can be used in a Set without inserting objects into the set which are identified as different objects, even though they have the same SHA1 and same name.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/git/GitObject.java
+++ b/src/main/java/hudson/plugins/git/GitObject.java
@@ -5,6 +5,7 @@ import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
 import java.io.Serializable;
+import java.util.Objects;
 
 /**
  * An object in a git repository. Includes the SHA1 and name of the
@@ -15,8 +16,8 @@ public class GitObject implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    ObjectId sha1;
-    String name;
+    final ObjectId sha1;
+    final String name;
 
     /**
      * Constructor for GitObject, a named SHA1 (tag, branch, etc.).
@@ -56,5 +57,44 @@ public class GitObject implements Serializable {
     @Exported(name="SHA1")
     public String getSHA1String() {
         return sha1 != null ? sha1.name() : null;
+    }
+
+    /**
+     * Returns a hash code value for the object. Considers sha1 and name in the
+     * calculation.
+     *
+     * @return a hash code value for this object.
+     */
+    @Override
+    public int hashCode() {
+        int hash = 3;
+        hash = 97 * hash + Objects.hashCode(this.sha1);
+        hash = 97 * hash + Objects.hashCode(this.name);
+        return hash;
+    }
+
+    /**
+     * Indicates whether some other object is "equal to" this one. Includes sha1
+     * and name in the comparison. Objects of subclasses of this object are not
+     * equal to objects of this class, even if they add no fields.
+     *
+     * @param obj the reference object with which to compare.
+     * @return true if this object is the same as the obj argument; false
+     * otherwise
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final GitObject other = (GitObject) obj;
+        return Objects.equals(this.name, other.name)
+                && Objects.equals(this.sha1, other.sha1);
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -19,6 +19,7 @@ import hudson.model.TaskListener;
 import hudson.plugins.git.Branch;
 import hudson.plugins.git.GitException;
 import hudson.plugins.git.GitLockFailedException;
+import hudson.plugins.git.GitObject;
 import hudson.plugins.git.IGitAPI;
 import hudson.plugins.git.IndexEntry;
 import hudson.plugins.git.Revision;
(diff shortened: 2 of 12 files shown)


src/main/java/hudson/plugins/git/GitObject.java
```diff
@@ -15,8 +16,8 @@
 
     private static final long serialVersionUID = 1L;
 
-    ObjectId sha1;
-    String name;
+    final ObjectId sha1;
+    final String name;
```
MarkEWaite: @jglick, @stephenc or @svanoort, does the change from package protected non-final to package protected final cause any risks in migration from version to version?  The object is serializable, and this change did not alter the number, type, or names of the fields, only the access

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2995,4 +2996,56 @@ static private boolean setsidExists() {
...
+         */
+        String[] output = result.split("[\\n\\r]");
+        if (output.length == 0 || (output.length == 1 && output[0].isEmpty())) {
+            return new HashSet<>();
+        }
+        Pattern pattern = Pattern.compile("(\\p{XDigit}{40})\\s+refs/tags/([^\\^]*)(\\^\\{\\})?");
+        Map<String, ObjectId> tagMap = new HashMap<>();
+        for (String line : output) {
+            Matcher matcher = pattern.matcher(line);
+            if (!matcher.find()) {
```
avdv: Do you really want to use `find` instead of `matches`? According to man git-show-ref the format is ` <SHA-1 ID> <space> <reference name>` and if anything else appears in the line it is probably an error to accept it.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2995,4 +2996,56 @@ static private boolean setsidExists() {
...
+        /*
+        Output shows SHA1 and tag with (optional) marker for annotated tags
+        7ac27f7a051e1017da9f7c45ade8f091dbe6f99d refs/tags/git-3.6.4
+        7b5856ef2b4d35530a06d6482d0f4e972769d89b refs/tags/git-3.6.4^{}
+         */
+        String[] output = result.split("[\\n\\r]");
+        if (output.length == 0 || (output.length == 1 && output[0].isEmpty())) {
+            return new HashSet<>();
+        }
+        Pattern pattern = Pattern.compile("(\\p{XDigit}{40})\\s+refs/tags/([^\\^]*)(\\^\\{\\})?");
```
avdv: Actually, you don't need to escape the ^ inside the character class, it only has a special meaning as the first character, so `[^^]` means the same thing and is easier to read.

Also, accepting an empty tag name is a bit weird...

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2995,4 +2996,56 @@ static private boolean setsidExists() {
...
+        }
+
+        /*
+        Output shows SHA1 and tag with (optional) marker for annotated tags
+        7ac27f7a051e1017da9f7c45ade8f091dbe6f99d refs/tags/git-3.6.4
+        7b5856ef2b4d35530a06d6482d0f4e972769d89b refs/tags/git-3.6.4^{}
+         */
+        String[] output = result.split("[\\n\\r]");
+        if (output.length == 0 || (output.length == 1 && output[0].isEmpty())) {
+            return new HashSet<>();
```
avdv: You could also simply use `Collections.emptySet()`.

MarkEWaite:
Thanks for the feedback @avdv .  I've implemented the changes and pushed to the branch.  I'll perform some further interactive testing to assure all is well.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
MarkEWaite: I've already seen more cases of git output changing unexpectedly than I want.  I'm leaving it more lax intentionally.

jacob-keller: In general, if the output changes, you're likely using the wrong git command, and should be parsing one of the plumbing commands instead. Changes have happened in the past. However, being slightly lax here isn't really a problem.

Git show-ref is definitely plumbing and thus unlikely to change.

src/main/java/hudson/plugins/git/GitObject.java
stephenc: Should not affect downstream plugins (unless they are squatting in the same package name... in which case they deserve what they get ;-) )

PR #289, opened 2017-12-06
POM updates

jglick:
@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.36</version>
+    <version>2.37</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.36</version>
+    <version>2.37</version>
```
jglick: Possible to go to 3.0 while still using an ancient Jenkins 1.x baseline, but not officially supported.

Note that to test against 2.89.x in `Jenkinsfile` you _do_ need to go to 3.x.

MarkEWaite:
Thanks for the note that using pom 3.0 is not supported with old Jenkins versions.  I'll need to assure that neither the git plugin nor the git client plugin versions use pom 3.0 until they have switched to use at least Jenkins 2.60.

PR #290, opened 2017-12-21
Custom ClassFilter entries

jglick:
https://github.com/jenkinsci/jenkins/pull/3120 shows that this plugin (well, really `git`) does unusual stuff that required custom whitelist entries.

@reviewbybees

### DIFF
--- a/src/main/resources/META-INF/hudson.remoting.ClassFilter
+++ b/src/main/resources/META-INF/hudson.remoting.ClassFilter
@@ -1,0 +1,8 @@
+# TODO avoid transferring JGit objects over the Remoting channel, or using them in build.xml
+org.eclipse.jgit.lib.ObjectId
+org.eclipse.jgit.lib.ObjectIdOwnerMap$Entry
+org.eclipse.jgit.lib.PersonIdent
+org.eclipse.jgit.revwalk.RevCommit
+org.eclipse.jgit.revwalk.RevObject
+org.eclipse.jgit.revwalk.RevTree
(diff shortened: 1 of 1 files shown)


jglick:
> Are these classes proposed for the whitelist?

Yes, these are already in the core whitelist (to protect users who neglect to update `git-client`).

> if we don't transfer JGit objects over the remoting channel, how can the agent perform a checkout with JGit?

The agent can _use_ JGit freely. The master could even _send_ JGit classes to the agent (though that seems unlikely). This is only about the master _receiving_ JGit types from the agent—as opposed to strings, collections, etc., or types defined in the plugin itself.

If you care to see typical problems, run against a patched core but prefix every line in this new file in this plugin with `!`, e.g.

```
!org.eclipse.jgit.transport.URIish
```

and then run `git-plugin` tests against those snapshots. (`GitSCMTest.testCustomSCMName`, for the above example. `RemotingTest.testRemotability` in this plugin requires `PersonIdent`, which perhaps does not correspond to any production code path.)

Especially as of jenkinsci/git-plugin#497 the Git plugin does a lot of tricky things with JGit. That is true even if you have specified the CLI impl of Git for use in the build. In a case covered by `SCMStepTest.checkoutsRestored` (in `workflow-scm-step`), several entries were needed because `GitSCM.printCommitMessageToLog` asks the agent to return a `RevCommit`, and then the master calls `getShortMessage()`. It would be simpler, faster, and safer to do this processing on the agent and send back a `String`, but the deceptive ease of Remoting tempts developers to do the wrong thing.

PR #292, opened 2017-12-27
Use Jenkins 2.60.1 and JDK 8

MarkEWaite:
Depend on Jenkins LTS 2.60.1.

Test in Jenkinsfile with Jenkins 2.60.1 and 2.89.1.

* [x] unit tests passing
* [x] interactive tests passing
* [x] code review
* [ ] 2 week beta test
  
Request review from @jenkinsci/code-reviewers

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 #!groovy
 
 // build both versions, retry test failures
-buildPlugin(jenkinsVersions: [null, '2.60.3'],
+buildPlugin(jenkinsVersions: [null, '2.89.1'],
             findbugs: [run:true, archive:true, unstableTotalAll: '0'],
             failFast: false)
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 #!groovy
 
 // build both versions, retry test failures
-buildPlugin(jenkinsVersions: [null, '2.89.2'],
+buildPlugin(jenkinsVersions: [null, '2.89.3'],
             findbugs: [run:true, archive:true, unstableTotalAll: '0'],
             failFast: false)
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 #!groovy
 
 // build both versions, retry test failures
-buildPlugin(jenkinsVersions: [null, '2.89.1'],
+buildPlugin(jenkinsVersions: [null, '2.89.2'],
             findbugs: [run:true, archive:true, unstableTotalAll: '0'],
             failFast: false)
 
--- a/pom.xml
+++ b/pom.xml
@@ -228,7 +228,7 @@
         <artifactId>findbugs-maven-plugin</artifactId>
         <configuration>
           <skip>${skipFindbugs}</skip>
-          <failOnError>false</failOnError>
+          <failOnError>true</failOnError>
           <xmlOutput>true</xmlOutput>
           <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -241,25 +241,6 @@
         <extensions>true</extensions>
       </plugin>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>findbugs-maven-plugin</artifactId>
-        <configuration>
-          <skip>${skipFindbugs}</skip>
-          <failOnError>true</failOnError>
-          <xmlOutput>true</xmlOutput>
-          <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
-        </configuration>
-        <executions>
-          <execution>
-            <id>run-findbugs</id>
-            <phase>verify</phase>
-            <goals>
-              <goal>check</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
           <systemProperties>
--- a/pom.xml
+++ b/pom.xml
@@ -47,8 +47,8 @@
     <maven-project-info-reports-plugin.version>2.9</maven-project-info-reports-plugin.version>
     <jenkins.version>2.60.1</jenkins.version>
     <java.level>8</java.level>
-    <jgit.version>4.5.4.201711221230-r</jgit.version>
-    <jgit.javadoc.version>4.5.0.201609210915-r</jgit.javadoc.version>
+    <jgit.version>4.9.2.201712150930-r</jgit.version>
+    <jgit.javadoc.version>4.9.2.201712150930-r</jgit.javadoc.version>
   </properties>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,6 @@
     <jenkins.version>2.60.1</jenkins.version>
     <java.level>8</java.level>
     <jgit.version>4.10.0.201712302008-r</jgit.version>
-    <jgit.javadoc.version>4.10.0.201712302008-r</jgit.javadoc.version>
   </properties>
 
   <repositories>
@@ -231,7 +230,7 @@
           <failOnError>false</failOnError>
           <links>
             <link>http://docs.oracle.com/javase/8/docs/api/</link>
-            <link>http://download.eclipse.org/jgit/site/${jgit.javadoc.version}/org.eclipse.jgit/apidocs/</link>
+            <link>http://download.eclipse.org/jgit/site/${jgit.version}/org.eclipse.jgit/apidocs/</link>
           </links>
         </configuration>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -47,8 +47,8 @@
     <maven-project-info-reports-plugin.version>2.9</maven-project-info-reports-plugin.version>
     <jenkins.version>2.60.1</jenkins.version>
     <java.level>8</java.level>
-    <jgit.version>4.9.2.201712150930-r</jgit.version>
-    <jgit.javadoc.version>4.9.2.201712150930-r</jgit.javadoc.version>
+    <jgit.version>4.10.0.201712302008-r</jgit.version>
+    <jgit.javadoc.version>4.10.0.201712302008-r</jgit.javadoc.version>
   </properties>
 
   <repositories>
@@ -95,6 +95,18 @@
       <groupId>org.eclipse.jgit</groupId>
       <artifactId>org.eclipse.jgit.http.server</artifactId>
       <version>${jgit.version}</version>
+      <exclusions>
+        <exclusion>
+          <!-- Provided by JSch plugin -->
+          <groupId>com.jcraft</groupId>
+          <artifactId>jsch</artifactId>
+        </exclusion>
+        <exclusion>
+          <!-- Provided by apache-httpcomponents-client-4-api -->
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpclient</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <!-- Include jgit's Apache HTTP client so that we can avoid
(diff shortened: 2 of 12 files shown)


pom.xml
```diff
@@ -137,7 +161,7 @@
     <dependency>
       <groupId>org.jenkins-ci.modules</groupId>
       <artifactId>sshd</artifactId>
-      <version>1.11</version>
+      <version>2.2</version>
```
oleg-nenashev: I should really finalize https://issues.jenkins-ci.org/browse/JENKINS-48578

pom.xml
```diff
@@ -198,8 +222,8 @@
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
```
oleg-nenashev: You could just remove the entire section and use `java.level` property

pom.xml
```diff
@@ -228,7 +252,7 @@
         <artifactId>findbugs-maven-plugin</artifactId>
         <configuration>
           <skip>${skipFindbugs}</skip>
```
oleg-nenashev: Can be also replaced by the Parent POM implementation

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -623,6 +619,9 @@ public void execute() throws GitException, InterruptedException {
 
                     FetchCommand fetch = git.fetch();
                     fetch.setTagOpt(tags ? TagOpt.FETCH_TAGS : TagOpt.NO_TAGS);
+                    if (allRefSpecs.isEmpty() && tags) {
```
oleg-nenashev: probably it is a separate fix, but ok

src/main/java/org/jenkinsci/plugins/gitclient/LegacyCompatibleGitAPIImpl.java
```diff
@@ -195,7 +195,10 @@ public void clone(RemoteConfig rc, boolean useShallowClone) throws GitException,
     /** {@inheritDoc} */
     @Override
     protected Object writeReplace() {
-        return remoteProxyFor(Channel.current().export(IGitAPI.class, this));
+        Channel currentChannel = Channel.current();
+        if (currentChannel == null)
+            return null;
```
oleg-nenashev: I would throw https://docs.oracle.com/javase/7/docs/api/java/io/ObjectStreamException.html or so to block serialization over other channels at all and to propagate the error

Jenkinsfile
```diff
@@ -1,7 +1,7 @@
 #!groovy
 
 // build both versions, retry test failures
-buildPlugin(jenkinsVersions: [null, '2.60.3'],
+buildPlugin(jenkinsVersions: [null, '2.89.1'],
```
jglick: 2.89.2

pom.xml
```diff
@@ -45,10 +45,10 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <argLine>-Dfile.encoding=${project.build.sourceEncoding}</argLine>
     <maven-project-info-reports-plugin.version>2.9</maven-project-info-reports-plugin.version>
-    <jenkins.version>1.625.3</jenkins.version>
-    <java.level>7</java.level>
-    <jgit.version>4.5.4.201711221230-r</jgit.version>
-    <jgit.javadoc.version>4.5.0.201609210915-r</jgit.javadoc.version>
+    <jenkins.version>2.60.1</jenkins.version>
```
jglick: I tend to recommend 2.60.3. No one should be running .1/.2 and people running .3 are better served by a plugin which is actually tested against the fixes they have.

pom.xml
```diff
@@ -45,10 +45,10 @@
...
+    <java.level>8</java.level>
+    <jgit.version>4.10.0.201712302008-r</jgit.version>
+    <jgit.javadoc.version>4.10.0.201712302008-r</jgit.javadoc.version>
```
jglick: BTW why are these even distinct?

pom.xml
jglick: Yes, remove all this.

pom.xml
```diff
@@ -228,7 +252,7 @@
         <artifactId>findbugs-maven-plugin</artifactId>
         <configuration>
           <skip>${skipFindbugs}</skip>
-          <failOnError>false</failOnError>
+          <failOnError>true</failOnError>
```
jglick: Delete, this is the default.

src/main/java/hudson/plugins/git/Revision.java
```diff
@@ -118,12 +118,7 @@ public String toString() {
         StringBuilder s = new StringBuilder("Revision " + revisionName + " (");
         if (branches != null) {
             Joiner.on(", ").appendTo(s,
-                    Iterables.transform(branches, new Function<Branch, String>() {
-
-                        public String apply(Branch from) {
-                            return Util.fixNull(from.getName());
-                        }
-                    }));
+                    Iterables.transform(branches, (Branch from) -> Util.fixNull(from.getName())));
```
jglick: Or even simpler:

```java
Iterables.transform(branches, from -> Util.fixNull(from.getName())));
```

src/main/java/org/jenkinsci/plugins/gitclient/AbstractGitAPIImpl.java
```diff
@@ -92,7 +92,10 @@ public void merge(ObjectId rev) throws GitException, InterruptedException {
      * @return a {@link java.lang.Object} object.
      */
     protected Object writeReplace() {
-        return remoteProxyFor(Channel.current().export(GitClient.class, this));
+        Channel currentChannel = Channel.current();
+        if (currentChannel == null)
+            return null;
```
jglick: More likely you would want to throw an `IllegalStateException` here.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -2563,11 +2560,7 @@ public String describe(ObjectId tip) throws IOException {
                 throw new GitException("No tags can describe "+tip);
 
             // if all the nodes are dominated by all the tags, the walk stops
-            Collections.sort(candidates,new Comparator<Candidate>() {
-                public int compare(Candidate o1, Candidate o2) {
-                    return o1.depth-o2.depth;
-                }
-            });
+            Collections.sort(candidates, (Candidate o1, Candidate o2) -> o1.depth-o2.depth);
```
jglick: Tip: if `depth` is given a matching getter,

```java
Collections.sort(candidates, Comparator.comparingInt(Candidate::getDepth));
```

src/main/java/org/jenkinsci/plugins/gitclient/Netrc.java
```diff
@@ -122,20 +122,24 @@ synchronized private Netrc parse() {
                         break;
 
                     case REQ_KEY:
-                        if ("login".equals(match)) {
```
jglick: Related?

src/main/javadoc/overview.html
```diff
@@ -5,31 +5,26 @@
...
-general-purpose git operations on a local or remote repository. Its
-primary use is from the
-<a href="https://wiki.jenkins-ci.org/display/JENKINS/Git+Plugin" target="_blank" unselectable="on">git-plugin</a>;
-as such, it is also used by
-<a href="https://wiki.jenkins-ci.org/display/JENKINS/Gerrit+Plugin" target="_blank" unselectable="on">gerrit-plugin</a>,
-<a href="https://wiki.jenkins-ci.org/display/JENKINS/Git+Parameter+Plugin" target="_blank" unselectable="on">git-parameter-plugin</a>,
-workflow and cloudbees validated merge plugins.
+The Jenkins git client plugin provides an API to execute general-purpose git operations on a local or remote repository.
+Its primary use is from the
+<a href="https://wiki.jenkins-ci.org/display/JENKINS/Git+Plugin" target="_blank" unselectable="on">git plugin</a>;
```
jglick: might as well use canonical URLs now

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
MarkEWaite: I'm not sure I understand this comment, @oleg-nenashev .  That change is required because the behavior of JGit changed from JGit 4.5 (the version previously supported by the plugin) to JGit 4.9.  We could include the transition from JGit 4.5 to JGit 4.10 as a separate pull request, but it is included as a separate commit already (two commits, showing the transition from JGit 4.5 to JGit 4.9 and then from JGit 4.9 to JGit 4.10).

pom.xml
MarkEWaite: When the JGit project releases a new patch version (for example, 4.5.4), they sometimes fail to publish the javadoc for that specific minor version.  Instead, they rely on the javadoc for the matching minor version (for example: 4.5).

I could collapse the two properties into a single one now, and then only expand them again if the JGit project fails to publish javadoc for some future patch release.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
jglick: > That change is required because the behavior of JGit changed from JGit 4.5 \[…\] to JGit 4.9.

Then seems fine to me. As a general tip, I find it worthwhile to look over my own PR when filing it to check for unintended changes, like gratuitous reindentation. At the same time I try to look for changes which are desired here but which would not make immediate sense to reviewers because there is no obvious link to the PR description. For those cases, I just leave comments in
 the PR—a self-review which makes reviews by others more streamlined.

(If the code itself is confusing, then add comments to the code—I am talking about code any version of which would look unremarkable if seen in isolation but the diff of which could be unexpected.)

pom.xml
```diff
@@ -16,7 +16,7 @@
 
   <name>Jenkins Git client plugin</name>
   <description>Utility plugin for Git support in Jenkins</description>
-  <url>https://wiki.jenkins.io/display/JENKINS/Git+Client+Plugin</url>
+  <url>https://plugins.jenkins.io/git-client</url>
```
jglick: You want to link to the wiki; otherwise there will be a redirect loop on `plugins.jenkins.io` and there will be no way to find the wiki. By “canonical URL” I meant the `jenkins.io` version as opposed to `jenkins-ci.org`.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
MarkEWaite: Thanks for the advice.  I missed the reality that most code review will be done by looking at the differences in the code, without looking at the comments on the commit which introduced the code.  That applies for code review before the merge of the pull request, and code review when someone is maintaining the code later.

The effort placing good commit messages would have been better spent and more useful as comments in the code with the information related to the commit message.  The knowledge is currently recorded, but not in the best place for those who will most benefit from that knowledge.

src/main/javadoc/overview.html
```diff
@@ -7,10 +7,10 @@
 
 The Jenkins git client plugin provides an API to execute general-purpose git operations on a local or remote repository.
 Its primary use is from the
-<a href="https://plugins.jenkins.io/git" target="_blank" unselectable="on">git plugin</a>;
+<a href="https://wiki.jenkins.io/display/JENKINS/Git+Plugin" target="_blank" unselectable="on">git plugin</a>;
```
jglick: `plugins.jenkins.io` is actually fine for permalinks like this. It was just the use from `<url>` in the POM which was a problem, since that is precisely the metadata that `plugins.jenkins.io` uses to determine the content to embed!

pom.xml
MarkEWaite: done

pom.xml
MarkEWaite: done

Jenkinsfile
MarkEWaite: done

pom.xml
MarkEWaite: done

pom.xml
MarkEWaite: done

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
MarkEWaite: done

pom.xml
MarkEWaite: done

src/main/java/hudson/plugins/git/Revision.java
MarkEWaite: Done.

PR #296, opened 2018-01-20
Fix JENKINS-48818 and keep JENKINS-46054 fixed

MarkEWaite:
The initial fix for [JENKINS-46054](https://issues.jenkins-ci.org/browse/JENKINS-46054) - update fails if submodule remote name ends with '.url'. introduced [JENKINS-48818](https://issues.jenkins-ci.org/browse/JENKINS-48818), a more serious regression than the bug it fixed.  Submodule update now fails with remote names which contain a whitespace character.  In git client 3.6.4 and prior, submodule update worked, even with a remote name which contained a space character.

This change adds more tests and resolves the regression introduced in git client 3.7.0.  It was verified interactively with the [JENKINS-48818 test job](https://github.com/MarkEWaite/jenkins-bugs/blob/JENKINS-48818/Jenkinsfile) and the [JENKINS-46054 test job](https://github.com/MarkEWaite/jenkins-bugs/blob/JENKINS-46054/Jenkinsfile).  It is verified with automated unit tests in SubmodulePatternTest and with a functional test in GitClientTest.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
@@ -41,7 +41,6 @@ import org.eclipse.jgit.lib.Constants;
 
 import static org.hamcrest.Matchers.*;
 import org.junit.AfterClass;
-import org.junit.Ignore;
 import static org.junit.Assert.*;
 import static org.junit.Assume.*;
 import org.junit.Before;
@@ -1442,7 +1441,6 @@ public class GitClientTest {
 
     @Issue("JENKINS-46054")
     @Test
-    @Ignore
     public void testSubmoduleUrlEndsWithDotUrl() throws Exception {
         // Create a new repository that includes ".url" in directory name
         File baseDir = tempFolder.newFolder();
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -1441,6 +1442,7 @@ public void testSubmodulesUsedFromOtherBranches() throws Exception {
 
     @Issue("JENKINS-46054")
     @Test
+    @Ignore
```
varyvol: Is this one ignored because it can be considered tested by "SubmodulePatternStringTest"? Why not removing then?

fcojfernandez: At least a comment pointing out the reason why the test is ignored (if the test is not removed as @varyvol suggests) would be useful.

MarkEWaite: Thanks for detecting that mistake!

The test should *not* be ignored.  The submodule pattern string test is a nice unit test that confirms the regular expression based processing is working as expected.  This test is an integration test that performs an "end to end" verification of behavior.  During earlier attempts to fix the problem, I had cases where the unit tests of the pattern string were failing, and this integration test failed with a real problem that was not caught by the unit tests.

MarkEWaite: Fixed by the next commit...

PR #300, opened 2018-02-16
Fix JENKINS-24304

ericcitaire:
Log executed command in Jenkins log instead of build console.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -59,6 +59,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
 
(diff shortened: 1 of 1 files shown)


MarkEWaite:
I'm sorry, but I can't accept this pull request.

We depend heavily on the diagnostic output that is written to the build log in order to identify git related issues in the job.  It won't work to shift that output into the log without warning, without a transition plan, and without a switch to bring it back for diagnosis.

ericcitaire:
It was a bad idea to depend on this output in the first place... Who is "We" by the way? CloudBees? what about the community?

"We" depend heavily on a clean job output in order to identify job related issues.

I wonder what the job output would look like if every single plugin wrote some random debugging informations in the job output...

MarkEWaite:
"We" is the set of people in the community who regularly use the output to diagnose authentication errors, connectivity errors, and configuration errors related to git.  The community is the primary beneficiary of that diagnostic output.  This plugin writes that diagnostic output because it has been helpful to the users of the plugin.  Other plugins don't have the same challenges as the git plugin, so they don't need to provide that diagnostic output.

If a clean build log is crucial to your use, you're welcome to submit a pull request which makes it optional to write to the console log.  The default setting needs to continue to write to the console log, but users who don't want the extra output could enable the setting to write to the Jenkins log instead of the build log.

ericcitaire:
Ok, great. So there is no hurry in closing this PR. Let's discuss about it.

If I manage to add an option to disable this output, will you consider merging it? or at least leaving it open so the community can comment?

MarkEWaite:
Adding an option to disable the output is much more involved.  If you're willing to add that option, I'm willing to consider it.  As this pull request sits now, I'm not willing to consider it.

PR #301, opened 2018-02-16
Fix JENKINS-24304

ericcitaire:


ericcitaire:
Two possible implementations:
* showGitCommands enabled by default (first commit)
* showGitCommands disabled by default (second commit)

I personnally prefer the second implementation.

ericcitaire:
I'll be happy to rewrite history if needed.

MarkEWaite:
I'd prefer "HIDE_GIT_COMMANDS" rather than "SHOW_GIT_COMMANDS" because the default must be to show the commands, not hide them.

A property is harder to use because there is no way to adjust it at runtime.  If someone is hiding git commands but needs diagnostic help, they'll need to stop and restart their Jenkins instance to adjust the setting, then stop and restart again to restore the setting.

ericcitaire:
> I'd prefer "HIDE_GIT_COMMANDS" rather than "SHOW_GIT_COMMANDS" because the default must be to show the commands, not hide them.

I don't understand why debugging should be the default mode.
JVM has debugging options to print out Garbage Collection statistics. This is very useful to track memory leaks, but it is not the default behavior.
But, ok. I can live with that.

> A property is harder to use because there is no way to adjust it at runtime. If someone is hiding git commands but needs diagnostic help, they'll need to stop and restart their Jenkins instance to adjust the setting, then stop and restart again to restore the setting.

I agree, the use of a property is not ideal, because of the reasons you're mentioning.
I suggested to use Jenkins System Logs which can easily be configured at runtime by creating a Log Recorder (that was my previous PR). It seemed a reasonable alternative to me.
What do you suggest?

MarkEWaite:
I won't accept the pull request if the default disables logging to the console log.  That will remove information from build logs that the community uses to help others.

I realize that my attempts to explain why that is the default have not persuaded you.  That's unfortunate, but it doesn't change my opinion.

I won't accept the pull request if the information is not logged somewhere.  

Using the Jenkins system log seems fine as an alternative when the information is not logged to the build log.

I'm not ready to handle frustration from many, many users because investigations were made more difficult by your proposed change.  Extra diagnostic text in the build log is better than removing that text and requiring users to find that information in a different location.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -257,6 +254,24 @@ protected CliGitAPIImpl(String gitExe, File workspace,
...
 
         launcher = new LocalLauncher(IGitAPI.verbose?listener:TaskListener.NULL);
+
+        GitTool git = resolveGitTool();
+        if (git != null && git.isHideDebugCommands()) {
+            this.verbose = false;
+        }
+    }
+
+    public GitTool resolveGitTool() {
```
MarkEWaite: Would you consider adding the findbugs suppression for this line, as shown in the [evaluation branch](https://github.com/MarkEWaite/git-client-plugin/commit/36a8abe8d29d0dec272395bb34b336a79063ddd0) I was using?

I haven't found a way to avoid that findbugs suppression.  The tests are structured in a way that causes GitTool to be used in contexts where there is no jenkins instance.  I don't want to rework the tests, so it is simplest to suppress the findbugs warning on the unnecessary null check when running in a real Jenkins environment.

MarkEWaite:
The change looks very good, but my interactive testing is not showing any difference in output comparing with hiding enabled or disabled.  Is it behaving correctly for you?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
ericcitaire: Hi @MarkEWaite. Sorry, this implementation doesn't work properly. Actually, the Jenkins instance is null on a slave node. I need to move this code to make this work.
I'll let you know.

ericcitaire:
I just realized that the property `hudson.plugins.git.IGitAPI.verbose` is already meant to control the verbosity of the Git CLI client (see [CliGitAPIImpl.java](https://github.com/jenkinsci/git-client-plugin/blob/4241e818dbe27fb78093094f29ebcd2203166610/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L253)). Unfortunately, it only affects the `CliGitAPIImpl.launcher.listener` (by assigning `TaskListener.NULL` to it), not `CliGitAPIImpl.listener`.

In fact, if you set `hudson.plugins.git.IGitAPI.verbose` to `true`, it will print every commands in the job output, in a slightly different format, and without information about the timeout.

It seems very redundant to me. What do you think?

The problem is that this property is `false` by default. And it needs a restart to take effect.

MarkEWaite:
Unfortunately, I think that it is better to leave IGitAPI.verbose alone and continue on the path you're following.  A separate (and new) setting avoids any collision with existing users, and allows more precise definition of the new setting (along with tests to confirm the new setting is honored).

ericcitaire:
I give up.

System properties are all over the place:
* [`useSETSID`](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L88), [`promptForAuthentication`](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L115) and [`untrustedSSL`](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L128) in `org.jenkinsci.plugins.gitclient.CliGitAPIImpl`
* [`verbose`](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/GitClient.java#L39) and [`quietRemoteBranches`](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/GitClient.java#L43) in `org.jenkinsci.plugins.gitclient.GitClient`

(none of them are tested)

Inlined implementations of git commands are tested with assertions in [`tearDown` methods](https://github.com/jenkinsci/git-client-plugin/blob/master/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java#L445) (seriously!?) to check that logs contains some expected comments in command lines...

It's a mess. The code needs to be refactored in order to be properly tested.

Don't get me wrong, I don't mean to disrespect your work. I like Jenkins and most of its plugins, they are all of great help, especially this one. But there is too much work to fit your requirements and I don't have the time for this.

Sorry.

PR #302, opened 2018-02-27
JENKINS-43427 Add sparse checkout support to Git LFS

tiffloon:
Fixed JENKINS-43427 by adding a check to see if lfsremote is present and if so call git config lfs.fetchinclude with all the paths in the sparse checkout field.


darxriggs:
I will try to implement some tests. I already found some issues.

PR #307, opened 2018-04-04
support non-heads refs in remote polling

avalanche123:
when trying to build github pull requests and using `refs/pull/{number}/(head|merge)` refs for that purpose, git plugin doesn't detect changes using remote polling because it only fetches `refs/heads` refs


avalanche123:
#181 seems related

avalanche123:
also see janinko/ghprb#423 for a more complete description of the issue

avalanche123:
@MarkEWaite thanks for your feedback, I've updated the change to be backwards compatible and to only drop `-h` argument if refspec starts with `refs/pull`, would this be considered low enough impact?

PR #313, opened 2018-04-23
Jenkins 48258 busy ssh text file fix

presPetkov:
Jenkins 48258 busy ssh text file fix - As per JENKINS-48258 comments.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1965,6 +1965,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     private File createUnixGitSSH(File key, String user) throws IOException {
         File ssh = createTempFile("ssh", ".sh");
+        File ssh_copy = new File(ssh.toString() + "-copy");
+        boolean isCopied = false;
         try (PrintWriter w = new PrintWriter(ssh, Charset.defaultCharset().toString())) {
             w.println("#!/bin/sh");
             // ${SSH_ASKPASS} might be ignored if ${DISPLAY} is not set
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1975,7 +1977,31 @@ private File createUnixGitSSH(File key, String user) throws IOException {
...
         }
         ssh.setExecutable(true, true);
-        return ssh;
+        //JENKINS-48258 git client plugin occasionally fails with "text file busy" error
+        //The following creates a copy of the generated file and deletes the original
+        //In case of a failure return the original and delete the copy
+        String fromLocation = ssh.toString();
+        String toLocation = ssh_copy.toString();
+        //Copying ssh file
+	try {
```
MarkEWaite: Since you submitted an indentation repair commit, could you also make the indentation on this line match the preceding lines?

The contributing guidelines generally prefer that white space changes be avoided so that it is easier to read diffs.  However, since you've made a change to white space, please align with the rest of the text.

dscunkel:
Should this fix (write to temporary and copy with cp) also be applied in CliGitApiImpl.createUnixSshAskpass() and createUnixStandardAskpass(), which are also writing scripts that would get executed by git?  A script given with SSH_ASKPASS would seem to be vulnerable to the same issues as one given with GIT_SSH.  It just would only affect users giving an ssh password instead of those of us providing an ssh key.

MarkEWaite:
As command line git versions are upgraded in distributions, it is becoming more and more attractive to detect if the command line git version is 2.3 or greater, then use `GIT_SSH_COMMAND` to avoid executing any temporary script.  That will need to wait for a future version.  

Including this in git client plugin 3.0.0 seems like a reasonable alternative, especially considering that several different users have reported success with it.

PR #327, opened 2018-06-29
[JENKINS-23345] Implement CloneCommand using clone instead of init+fetch

mdoppler:
This pull request replaces **git init+fetch** in CliGitApiImpl CloneCommand by proper **git clone** for better performance on the initial clone of the repository. 

The speedup depends on the number of commits and tags in the repository. For a high number of tags, git clone ist much faster than git init+fetch.

Changes in comparison to init+fetch:
- the refspec **+refs/heads/\*:refs/remotes/origin/\*** is implicitly added by git clone
- for shallow clones of local repositories the git clone command requires a **file://** scheme prefix

The tests have been updated to reflect these changes. 

A testcase **test_getTags_after_clone** that checks getting tags after cloning has been added. This slightly improves coverage.

This fixes [JENKINS-23345](https://issues.jenkins-ci.org/browse/JENKINS-23345).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -166,6 +166,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     /* Package protected for testing */
     final static String SUBMODULE_REMOTE_PATTERN_STRING = SUBMODULE_REMOTE_PATTERN_CONFIG_KEY + "\\s+[^\\s]+$";
 
+    protected static final String DEFAULT_ORIGIN = "orgin";
+
     private void warnIfWindowsTemporaryDirNameHasSpaces() {
         if (!isWindows()) {
             return;
@@ -471,13 +473,15 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     public CloneCommand clone_() {
         return new CloneCommand() {
             String url;
-            String origin = "origin";
+            String origin = DEFAULT_ORIGIN;
             String reference;
-            boolean shallow,shared;
+            boolean shallow = false;
+            boolean shared = false;
             Integer timeout;
             boolean tags = true;
             List<RefSpec> refspecs;
             Integer depth = 1;
+            private boolean noCheckout = true;
 
             public CloneCommand url(String url) {
                 this.url = url;
@@ -510,7 +514,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             }
 
             public CloneCommand noCheckout() {
-                //this.noCheckout = true; Since the "clone" command has been replaced with init + fetch, the --no-checkout option is always satisfied
+                this.noCheckout = true;
                 return this;
             }
 
@@ -558,11 +562,6 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                     throw new GitException("Failed to delete workspace", e);
                 }
 
-                // we don't run a 'git clone' command but git init + git fetch
-                // this allows launchCommandWithCredentials() to pass credentials via a local gitconfig
-
-                init_().workspace(workspace.getAbsolutePath()).execute();
-
                 if (shared) {
                     if (reference == null || reference.isEmpty()) {
                         // we use origin as reference
@@ -572,6 +571,50 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                     }
                 }
 
+                launchCloneCommand(urIish);
+
+                addReferenceToAlternates();
+                if (refspecs != null) {
+                    for (RefSpec refSpec : refspecs) {
+                        launchCommand("config", "--add", "remote." + origin + ".fetch", refSpec.toString());
+                    }
+                }
+            }
+
+            private void launchCloneCommand(URIish urIish) throws InterruptedException {
+                ArgumentListBuilder args = new ArgumentListBuilder();
+                args.add("clone");
+                if( noCheckout ) {
+                    args.add("--no-checkout");
+                }
+                if( shared ) {
+                    args.add("--shared");
+                }
+                if( shallow ) {
+                    if( new File( url ).exists() ) { // shallow clones of local repos need file urls
+                        urIish = urIish.setScheme( "file" );
+                    }
+                    args.add("--depth");
+                    args.add( depth );
+                }
+                if( !origin.equals( DEFAULT_ORIGIN ) ) {
+                    args.add( "--origin" );
+                    args.add( origin );
+                }
+
+                if (isAtLeastVersion(1,7,1,0)) {
+                    args.add("--progress");
+                }
+                args.add(urIish.toString());
+                args.add(workspace.getAbsolutePath());
+
+                StandardCredentials cred = credentials.get(urIish.toPrivateString());
+                if (cred == null) cred = defaultCredentials;
+
+                launchCommandWithCredentials(args, workspace.getParentFile(), cred, urIish, timeout);
+            }
+
+            private void addReferenceToAlternates() {
                 if (reference != null && !reference.isEmpty()) {
                     File referencePath = new File(reference);
                     if (!referencePath.exists())
@@ -585,9 +628,9 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                             // reference path is bare repo
                             objectsPath = new File(referencePath, "objects");
                         }
-                        if (!objectsPath.isDirectory())
+                        if (!objectsPath.isDirectory()) {
                             listener.error("Reference path does not contain an objects directory (no git repo?): " + objectsPath);
-                        else {
+                        } else {
                             File alternates = new File(workspace, ".git/objects/info/alternates");
                             try (PrintWriter w = new PrintWriter(alternates, Charset.defaultCharset().toString())) {
                                 String absoluteReference = objectsPath.getAbsolutePath().replace('\\', '/');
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -572,6 +571,50 @@ public void execute() throws GitException, InterruptedException {
                     }
                 }
 
+                launchCloneCommand(urIish);
```
MarkEWaite: This change will dramatically increase disc use and time to clone for large repositories which were using a reference repository previously.  It negates almost entirely the benefits of a reference repository by not adding the reference repository until after the clone has been performed.

The reference repository **must** be honored in the initial clone.  

If it is not honored, then large repositories (like the LInux kernel with its 2 GB repo or many other repos) will copy the entire repository on first clone.

MarkEWaite: This change also removes all the benefits that were previously gained for Pipeline and multi-branch Pipeline jobs by honoring the refspec on first clone.  Honoring the refspec on first clone results in a significant reduction in clone time and disc use.

If a narrow refspec has been specified, it **must** be honored in the initial clone.

MarkEWaite: This change also ignores the "no tags" setting on initial clone.  Honoring "no-tags" on initial clone has not been as dramatic a savings as honoring a narrow refspec or using a reference repository, but that would also be a regression for those jobs which have been created that do not clone tags initially.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -691,15 +700,22 @@ public void test_clone_refspecs() throws Exception {
       w.git.clone_().url(localMirror()).refspecs(refspecs).repositoryName("origin").execute();
       w.git.withRepository((Repository repo, VirtualChannel channel) -> {
           String[] fetchRefSpecs = repo.getConfig().getStringList(ConfigConstants.CONFIG_REMOTE_SECTION, Constants.DEFAULT_REMOTE_NAME, "fetch");
-          assertEquals("Expected 2 refspecs", 2, fetchRefSpecs.length);
-          assertEquals("Incorrect refspec 1", "+refs/heads/master:refs/remotes/origin/master", fetchRefSpecs[0]);
-          assertEquals("Incorrect refspec 2", "+refs/heads/1.4.x:refs/remotes/origin/1.4.x", fetchRefSpecs[1]);
+          int refSpecIndex = 0;
+          if( fetchRefSpecs.length == 3 ) {
```
MarkEWaite: This will cause a major increase in repository size for those repositories which previously relied on the refspec being honored on first fetch.

By adding the "full width" refspec, the other two refspecs which follow are extraneous.  They are already covered by the full width refspec.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -4085,7 +4119,7 @@ public void test_autocreate_fails_on_multiple_matching_origins() throws Exceptio
         w.git.fetch_().from(remote, refspecs).execute();
 
         try {
-            w.git.checkout().ref(Constants.MASTER).execute();
+            w.git.checkout().ref("1.4.x").execute();
```
MarkEWaite: This change hints that something has changed in an incompatible way.  I suspect that switching to "clone" instead of "init + fetch" means that a branch matching the remote default branch name (usually "master") is always created.  That will be a functional regression for users that depend on the initial checkout not creating a named branch unless specifically requested to do so.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
mdoppler: As specified in the pull request description above, git clone implicitly adds the full width refspec when doing the initial git clone. According to the git documentation there is no possibility to limit the refspec during git clone: https://git-scm.com/docs/git-clone

According to our experience cloning a repository with the full refspec but without checkout (--no-checkout) and then checking out the requested ref is much faster than fetching only the required refspec initiallly. 

I understand your concerns regarding the repository size for projects that used a specific refspec to save space while sacrificing it for the slower performance of git fetch. This prohibits changing the default behaivour of the clone command. Would it be acceptable to have this behaviour configurable and off by default?

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
mdoppler: Yes, the default branch of the origin or the branch explicitly set using the -b option is implicitly available, when cloning a repository.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
mdoppler: From the discussion in https://issues.jenkins-ci.org/browse/JENKINS-23345 there seem to be at least some other people that have similar performance issues with git init+fetch in comparison to git clone.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
mdoppler: Are there cases where git init+fetch is faster than git clone? Is this reproducible using repositories like the linux kernel with specific refspecs?

MarkEWaite: Yes, there are cases where init+fetch is much faster than clone.  My jenkins-bugs repo is only 50 MB (small in the world of large git repositories) and init+fetch is much faster than clone when using a narrow refspec.

Git init+fetch: less than 1 second for a single branch with a narrow refspec

```
$ mkdir jenkins-bugs
$ cd jenkins-bugs
$ git init
$ time git fetch git://github.com/MarkEWaite/jenkins-bugs.git +refs/heads/JENKINS-43198:refs/remotes/bare/JENKINS-43198
real    0m0.643s
user    0m0.016s
sys     0m0.016s
```

Git clone: 8 seconds

```
$ time git clone git://github.com/MarkEWaite/jenkins-bugs.git
real    0m8.748s
user    0m19.824s
sys     0m0.844s
```

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
MarkEWaite: You asked:

> Would it be acceptable to have this behaviour configurable and off by default?

I think you're asking if it would be acceptable to allow the caller of the git client API to optionally choose if it should use clone rather than init and fetch.  Yes, that would be acceptable, so long as the default remained to use init and fetch.  There are more than 130000 Jenkins installations that depend on the git plugin.  Many of those installations use it in ways that are critical to their production environments.

We very rarely change default behaviors of the git client plugin or the git plugin because so many installations depend on those default behaviors.

MarkEWaite: You mentioned JENKINS-23345 but didn't mention the performance comparisons I reported as part of that bug report.  Can you share the commands you're using and the `time` output of those commands so that I have some idea why `clone` is significantly faster than `init+fetch` in your case?

I understand from the comments by Jason Salaz of GitHub that there are cases where clone will be significantly faster than init + fetch.  Unfortunately, I was unable to duplicate those cases.  I was able to duplicate that `init+fetch` is slightly slower.  I don't see any way to retain the existing git client plugin behaviors and still switch to `clone`.  

If the implementation switches to `git clone` then narrow refspecs are out (since git clone doesn't take a refspec as an argument).

It should be possible in the implementation you provided to use a reference repository from the clone command.

I think many of the narrow refspec cases might be supported by newer command line git implementations if the `--single-branch --branch BRANCH_NAME` arguments are passed.  Not all of them, but many of the cases that I've encountered could be handled by that.

If a parsing of the user provided refspec (if one is set) shows that it could only match a single branch name, then `git clone --single-branch --branch BRANCH_NAME` has the same performance as `init+fetch` .  In my example, they both complete in less than 1 second when the clone with full width refspec requires 8 or more seconds

MarkEWaite:
@mdoppler if you are determined to press forward with adding `git clone` as an alternative to `git init + git fetch` you might consider creating a different implementation.  

Today there are 3 implementations, "default" (command line git), "jgit" (using pure java code from the JGit project), and "jgitapache" (using pure Java code from the JGit project and Apache HTTP authentication).

You might add a fourth implementation, as "git-clone" (or "git-using-clone").  Then the behavioral differences between the implementations are left as a choice to those who choose a specific implementation.  When a specific job chooses an implementation, it states its acceptance of the limitations of that implementation.  Thus, a job which uses "jgit" accepts that it doesn't have git large file support and doesn't have full support for submodules.

mdoppler:
Thanks for your time to review this pull request and provide examples of faster **git init+fetch** invocations than **git clone** for smaller repositories. The performance of init+fetch seems to  get slower than clone after a certain amount of tags or commits. 

I agree that because the change would make the initial checkout slower and larger for smaller git repositories, this rules out changing the default behaviour to use cloning.

> You might add a fourth implementation, as "git-clone" (or "git-using-clone"). Then the behavioral differences between the implementations are left as a choice to those who choose a specific implementation. When a specific job chooses an implementation, it states its acceptance of the limitations of that implementation. Thus, a job which uses "jgit" accepts that it doesn't have git large file support and doesn't have full support for submodules.

I will consider implementing it this way. The real clone implementation could derive CGitApiImpl while only changing the CloneCommand to do real cloning instead of git+fetch.

I am closing this request and will open a new one when i get to implement the clone based strategy.

PR #334, opened 2018-07-14
Use jgit 5.0.1

MarkEWaite:
Use latest JGit release and update JGitAPIImpl and GitAPITestCase for the change from getRef(String) to exactRef(String) and findRef(String).

Also updates the equalsverifier test dependency to use the latest release.

@darxriggs since you've been looking at tests recently on the plugin, I'd love to have you review this pull request.

I've used this change in my test configuration and confirmed that JGit 5.0.1 clones repositories from:

* Locally hosted git with ssh
* Assembla with ssh and https
* Beanstalk with ssh and https
* Bitbucket with ssh and https
* GitHub with ssh and https
* Gitlab with ssh and https
* Visual Studio Online with ssh and https

None of the regression tests have detected any issues that cause JGit 5.0.1 to behave any differently than JGit 4.11.0 (included in git client plugin 3.0.0-beta3) or JGit 4.5.4 (included in git client plugin 2.7.2).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
@@ -766,7 +766,6 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 	}
 	try (Repository repo = getRepository()) {
 	    List<Ref> refList = repo.getRefDatabase().getRefsByPrefix(refPrefix);
-	    // The key set for refList will have refPrefix removed, so to recover it we just grab the full name.
 	    Set<String> refs = new HashSet<>(refList.size());
 	    for (Ref ref : refList) {
 		refs.add(ref.getName());


src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -536,7 +536,7 @@ public void tag(String name, String message) throws GitException {
     @Override
     public boolean tagExists(String tagName) throws GitException {
         try (Repository repo = getRepository()) {
-            Ref tag =  repo.getRefDatabase().getRef(R_TAGS + tagName);
+            Ref tag =  repo.getRefDatabase().exactRef(R_TAGS + tagName);
```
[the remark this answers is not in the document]
MarkEWaite: Thanks for catching those!

I've made changes for the cases you detected, rebased onto master, and pushed the commits.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -765,10 +765,10 @@ public void deleteRef(String refName) throws GitException, InterruptedException
 	    refPrefix = refPrefix.replace(' ', '_');
 	}
 	try (Repository repo = getRepository()) {
-	    Map<String, Ref> refList = repo.getRefDatabase().getRefs(refPrefix);
+	    List<Ref> refList = repo.getRefDatabase().getRefsByPrefix(refPrefix);
```
darxriggs: The comment below is out-of-sync, still about the "key set".

MarkEWaite: Thanks.  That comment is incorrect now.  That code is now just converting a `List<Ref>` into a `Set<String>`.  I'll delete the incorrect comment.

darxriggs:
Having a look at the release notes of JGit 5, first I was wondering why the big version increase.
I think it's due to the support for protocol v2 and removing that many deprecated methods.
Beside these major changes (which should not impact the plugin that much) there are not that many other ones. Because of this I think no regressions have been detected so far.

But anyway I think this should be released as another beta version so others can test it. (I don't know how many users are testing the beta versions.)

MarkEWaite:
Agreed that it should release as part of the next beta version.

PR #335, opened 2018-07-14
Report better exception if filename not supported

MarkEWaite:
When `LANG=C mvn -Dtest=CliGitAPIImplTest#test_clean test` fails, give a better message than the exception from "git add".

This is not foolproof, just an aid for diagnosing test failures when the locale is unable to support the filename in the test.

@darxriggs since you've been looking at tests recently on the plugin, I'd love to have you review this pull request.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -865,7 +865,14 @@ public abstract class GitAPITestCase extends TestCase {
          */
         String fileName = "\uD835\uDD65-\u5c4f\u5e55\u622a\u56fe-\u0041\u030a-\u00c5-\u212b-fileName.xml";
         w.touch(fileName, "content " + fileName);
-        w.git.add(fileName);
+        try {
+            w.git.add(fileName);
+        } catch (GitException ge) {
+            // Exception should contain actual file name
+            // If mangled name is seen instead, throw a clear exception to indicate root cause
+            assertTrue("System locale does not support filename '" + fileName + "'", ge.getMessage().contains("?-????-A?-?-?-fileName.xml"));
+            throw ge; // Fail the test on exception even if the preceding assertion did not fail
+        }
         w.git.commit(fileName);
 
         /* JENKINS-27910 reported that certain cyrillic file names


darxriggs:
Yes, the output is improved a bit, which should help debuging.
Based on the execution environment, the reporting (encoding) can be different - see below.

**CLI**
`System locale does not support filename '?-????-A?-?-?-fileName.xml'`

**IDE**
`System locale does not support filename '𝕥-屏幕截图-Å-Å-Å-fileName.xml'`

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -865,7 +865,14 @@ public void test_clean() throws Exception {
          */
         String fileName = "\uD835\uDD65-\u5c4f\u5e55\u622a\u56fe-\u0041\u030a-\u00c5-\u212b-fileName.xml";
         w.touch(fileName, "content " + fileName);
-        w.git.add(fileName);
+        try {
```
darxriggs: The problem with many tests in this project is, that in a single test method many assertions for different cases are performed. This is the case here too.

This improved error reporting should be performed for all assertions in this test method - lines [869](https://github.com/MarkEWaite/git-client-plugin/blob/de9da9a7f942e113b919b200b1811332acdb6aeb/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java#L869), [883](https://github.com/MarkEWaite/git-client-plugin/blob/de9da9a7f942e113b919b200b1811332acdb6aeb/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java#L883), [888](https://github.com/MarkEWaite/git-client-plugin/blob/de9da9a7f942e113b919b200b1811332acdb6aeb/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java#L888).

I suggest to implement a helper method which wraps the improved reporting. With the help of Java 8 lambdas this even reads quite well.

line 869
```
withSystemLocaleReporting(fileName, () -> {
    w.git.add(fileName);
    w.git.commit(fileName);
});
```

line 883
```
withSystemLocaleReporting(fileNameSwim, () -> {
    w.git.add(fileNameSwim);
    w.git.commit(fileNameSwim);
});
```

line 888
```
withSystemLocaleReporting(fileNameFace, () -> {
    w.git.add(fileNameFace);
    w.git.commit(fileNameFace);
});
```

helper method
```
private void withSystemLocaleReporting(String fileName, TestedCode code) throws Exception {
    try {
        code.run();
    } catch (GitException ge) {
        // Exception message should contain the actual file name.
        // It may just contain ? for characters that are not encoded correctly due to the system locale.
        // If such a mangled file name is seen instead, throw a clear exception to indicate the root cause.
        assertTrue("System locale does not support filename '" + fileName + "'", ge.getMessage().contains("?"));
        // Rethrow exception for all other issues.
        throw ge;
    }
}

@FunctionalInterface
interface TestedCode {
    void run() throws Exception;
}
```

darxriggs:
Oh you already merged while I was writing the review.

PR #338, opened 2018-07-23
Fix JENKINS-50573 on master branch

MarkEWaite:
Merges changes from the stable-2.7 branch onto the master branch and makes minor changes as needed to adapt to the master branch.

@reviewbybees please and @richbg as well.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/CredentialsTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/CredentialsTest.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.StringJoiner;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/CredentialsTest.java
```diff
@@ -295,21 +306,88 @@ public static Collection gitRepoUrls() throws MalformedURLException, FileNotFoun
...
+            cmd.shallow(true).depth(1);
+        }
+        cmd.execute();
     }
 
-    private void addCredential(String username, String password, File privateKey) throws IOException {
-        if (random.nextBoolean()) {
+    private String listDir(File dir) {
+        File[] files = repo.listFiles();
+        StringBuilder fileList = new StringBuilder();
```
jvz: Could be simpler to use [StringJoiner](https://docs.oracle.com/javase/8/docs/api/java/util/StringJoiner.html).

MarkEWaite: Yes, since this branch uses Java 8, that would be better.

The changes in #337 can't use StringJoiner because that version can only use Java 7 language features.

PR #342, opened 2018-07-25
Improve tests that are system locale dependent

darxriggs:
This is a follow-up to de9da9a7f942e113b919b200b1811332acdb6aeb.


darxriggs:
@MarkEWaite If you don't like it that's ok for me. I just wanted to give it a last try.

PR #343, opened 2018-07-26
Add test that verifies invocation of ToolInstallers on master

oleg-nenashev:
Just an extra test, which verifies that Core Tool management logic is fine for Jenkins master node. 

https://github.com/jenkinsci/git-plugin/pull/605 is a related fix in the downstream plugin

@MarkEWaite @btk3000 @jenkinsci/code-reviewers

### DIFF
--- a/src/test/java/hudson/plugins/git/GitToolTest.java
+++ b/src/test/java/hudson/plugins/git/GitToolTest.java
@@ -4,14 +4,20 @@ import hudson.EnvVars;
 import hudson.model.Node;
 import hudson.model.TaskListener;
 import hudson.slaves.DumbSlave;
+import hudson.tools.CommandInstaller;
+import hudson.tools.InstallSourceProperty;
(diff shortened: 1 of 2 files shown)


src/test/java/hudson/plugins/git/GitToolTest.java
```diff
@@ -43,6 +49,21 @@ public void testForNode() throws Exception {
         assertEquals(gitTool.getGitExe(), newTool.getGitExe());
     }
 
+    @Test
+    public void shouldResolveToolsOnMaster() throws Exception {
+        Assume.assumeTrue(!hudson.remoting.Launcher.isWindows());
```
MarkEWaite: Explanatory comment why this is skipped on Windows?

MarkEWaite: I think I found the reason why it is skipped on Windows.  I've prepared a proposed change to the test so that it can also be run on Windows.

PR #348, opened 2018-08-02
[JENKINS-44720] Add option for updating submodules in parallel

nre-ableton:
This change uses an ExecutorService to optionally update submodules in
parallel. This can dramatically improving update times for
repositories with lots of submodules. Since the CliGitAPIImpl class
doesn't launch a single "git submodule update" command, but instead
launches one command per submodule, we can't use the "--jobs" CLI
argument for git, available since git 2.9.0.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -58,6 +58,9 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
@@ -1072,6 +1075,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             Map<String, String> submodBranch   = new HashMap<>();
             public Integer timeout;
             Integer depth = 1;
+            Integer threads = 1;
 
             public SubmoduleUpdateCommand recursive(boolean recursive) {
                 this.recursive = recursive;
@@ -1110,6 +1114,11 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
             public SubmoduleUpdateCommand depth(Integer depth) {
                 this.depth = depth;
+                return this;
+            }
+
+            public SubmoduleUpdateCommand threads(Integer threads) {
+                this.threads = threads;
                 return this;
             }
 
@@ -1176,6 +1185,14 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 // path.
                 Pattern pattern = Pattern.compile(SUBMODULE_REMOTE_PATTERN_STRING, Pattern.MULTILINE);
                 Matcher matcher = pattern.matcher(cfgOutput);
+
+                ExecutorService executorService;
+                if (threads > 1) {
+                    executorService = Executors.newFixedThreadPool(threads);
+                } else {
+                    executorService = Executors.newSingleThreadExecutor();
+                }
+
                 while (matcher.find()) {
                     ArgumentListBuilder perModuleArgs = args.clone();
                     String sModuleName = matcher.group(1);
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1209,8 +1226,19 @@ else if (!referencePath.isDirectory())
                     String sModulePath = getSubmodulePath(sModuleName);
 
                     perModuleArgs.add(sModulePath);
-                    launchCommandWithCredentials(perModuleArgs, workspace, cred, urIish, timeout);
+                    StandardCredentials finalCred = cred;
+                    URIish finalUrIish = urIish;
+                    executorService.submit(() -> {
```
darxriggs: This change has a big issue - now exceptions are swallowed. This should really be fixed.

nre-ableton: We're only catching `InterruptedException`, and even then, we re-throw it as `GitException`. Perhaps I'm misunderstanding something, but I don't believe that this change is swallowing all exceptions.

darxriggs: You can verify the described behaviour by writing a test that performs a git submodule call with an invalid parameter. No output is logged and no exception is thrown.

Have a look at the [Future<?> submit(Runnable task)](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#submit-java.lang.Runnable-). The returned `Future` has a [V get()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html#get--) method which throws an `ExecutionException` if the computation threw an exception.

Every `Future` should be checked for an exception.

Some useful references:
* https://www.baeldung.com/java-executor-service-tutorial
* https://www.baeldung.com/java-runnable-callable
* https://www.baeldung.com/java-executor-wait-for-threads

nre-ableton: @darxriggs Thanks for the additional information. I'll try to put together a patch in a new PR (along with a test) to fix this behavior.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1176,6 +1185,14 @@ else if (!referencePath.isDirectory())
                 // path.
                 Pattern pattern = Pattern.compile(SUBMODULE_REMOTE_PATTERN_STRING, Pattern.MULTILINE);
                 Matcher matcher = pattern.matcher(cfgOutput);
+
+                ExecutorService executorService;
```
darxriggs: Naming the threads for debugging purposes in production would be a good idea.

Some useful resources
* see https://dzone.com/articles/executorservice-10-tips-and
* https://github.com/jenkinsci/jenkins/blob/jenkins-2.137/core/src/main/java/hudson/Proc.java#L141

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1209,8 +1226,19 @@ else if (!referencePath.isDirectory())
...
+                        try {
+                            launchCommandWithCredentials(perModuleArgs, workspace, finalCred, finalUrIish, timeout);
+                        } catch (InterruptedException e) {
+                            throw new GitException("Interrupted while updating submodule for " + sModuleName);
+                        }
+                    });
                 }
+
+                executorService.shutdown();
+                executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```
darxriggs: There are more flaws in here:

1) The timeout value is really big. Why not use the configurable `timeout` value?
2) The method [boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#awaitTermination-long-java.util.concurrent.TimeUnit-) just asks the tasks to complete. If they have not been completed within the specified timeout this is returned in the result. [List<Runnable> shutdownNow()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#shutdownNow--) should then be used.
3) `InterruptedException`s are not handled.

A good example to properly use the `ExectorService` is given in its [Javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html) in the beginning.

nre-ableton: 1. As I understand it (based on the previous implementation), the `timeout` value is for each submodule update, not the duration of all updates. If we used `timeout` for the value in `awaitTermination`, it would then apply to the duration of all submodules.
2. Do you mean to say that we should check the return value of `shutdown()` and then call `shutdownNow()` if it returned false?
3. I didn't think that this was necessary because the method declares that it throws `InterruptedException`. Should we rather `try`/`catch` the `awaitTermination` statement and repackage the `InterruptedException` as a `GitException`?

nre-ableton:
@darxriggs So I've come up with an initial solution, though I haven't made a PR yet:

https://github.com/jenkinsci/git-client-plugin/compare/master...nre-ableton:nre/master/future-check-exceptions?expand=1

There is a WIP commit in the middle of that branch, marked as such since I can't figure out why the test fails even when `thrown.expect(GitException.class)` is there. Also, other than abusing the `depth` option, I had a difficult time figuring out how to get the unit tests to construct an invalid `git submodule` command. Am I looking at this problem the wrong way? I'd really like to have a unit test to go along with this PR; I think it's fairly important to avoid this type of bug in the future.

nre-ableton:
@darxriggs Ah, I just realized that my branch has some errors, namely that `Future.get()` is of course a blocking operation. I will fix this, but in the meantime, my question about the unit tests still stands.

PR #350, opened 2018-08-05
[JENKINS-52891] Always log the timeout for a CliGit command

darxriggs:
[JENKINS-52891](https://issues.jenkins-ci.org/browse/JENKINS-52891)

A timeout is always used for a cli git command and therefore should always be logged.

Options a timeout may be set
- timeout parameter for some commands
- system property org.jenkinsci.plugins.gitclient.Git.timeOut
- default of 10 minutes

So far the timeout was only logged for commands where
a) it cannot be customized
b) it can and actually was customized

Category a) some commands with correct default timeout logging so far
    > git --version # timeout=10
    > git init ... # timeout=10
    > git config ... # timeout=10
    > git submodule init # timeout=10

Category b) commands with missing default timeout logging so far
    > git fetch ...
    > git checkout ...
    > git submodule update ...

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2102,11 +2102,12 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 /* GIT_SSH won't call the passphrase prompt script unless detached from controlling terminal */
                 args.prepend("setsid");
             }
-            listener.getLogger().println(" > " + command + (timeout != null ? TIMEOUT_LOG_PREFIX + timeout : ""));
+            int usedTimeout = timeout == null ? TIMEOUT : timeout;
+            listener.getLogger().println(" > " + command + TIMEOUT_LOG_PREFIX + usedTimeout);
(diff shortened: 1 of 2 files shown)


darxriggs:
Hi @MarkEWaite, could you have a look at this PR?

I am also not sure how the linked ticket has to be handled?
* Has a link from the ticket to the PR be created manually? (Sometimes I have seen some automatic linking)
* Who is changing the state from "In Progress" to "Review" and who should get assigned?
* Who is changing state and version information when it is released?
* How is issue added to the release notes?

MarkEWaite:
Comments will be automatically attached to the ticket if the commit message includes the Jira ID in the message.

I've not seen any automatic linking, though that may also be happening.

When a new version is released, I update the wiki page with the changes since the last release.  I've added a placeholder for the next version sometimes without adding a release date.  For example, in this case, a placeholder might have been added as "3.0.0-beta6" with "XX-XXX-2018" as the release date.

When a new version of the plugin is released, I click each of the bug hyperlinks from the release notes and set the state of the bug to match the release if it is not already correctly set.

When the pull request is merged to the master branch, the pull request author is assumed that they will change the state from "In Review" to "Fixed (not released)".  If they don't, then I'll update the Jira state when the plugin is released.

You're welcome to insert release note information into the wiki page prior to a release.  That would be a nice help.

PR #351, opened 2018-08-11
Use JGit's implementation of prune during fetch

MarkEWaite:
JGit 5.0.2 fixes the behavior of prune during fetch so that the alternative implementation is no longer required.  Unfortunately, JGit 5 also logs warnings that were not logged previously when it attempts to remove a directory that is not empty.

@darxriggs I'd love to have your review of the proposed change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
     <argLine>-Dfile.encoding=${project.build.sourceEncoding}</argLine>
     <jenkins.version>2.60.3</jenkins.version>
     <java.level>8</java.level>
-    <jgit.version>5.1.0.201808281540-m3</jgit.version>
+    <jgit.version>5.0.3.201809091024-r</jgit.version>
   </properties>
 
   <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
     <argLine>-Dfile.encoding=${project.build.sourceEncoding}</argLine>
     <jenkins.version>2.60.3</jenkins.version>
     <java.level>8</java.level>
-    <jgit.version>5.0.2.201807311906-r</jgit.version>
+    <jgit.version>5.1.0.201808281540-m3</jgit.version>
   </properties>
 
   <dependencies>
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -1323,13 +1323,13 @@ public abstract class GitAPITestCase extends TestCase {
     }
 
     /**
-     * JGit 3.3.0 thru 3.6.0 "prune during fetch" prunes more remote
-     * branches than command line git prunes during fetch.  This test
-     * should be used to evaluate future versions of JGit to see if
-     * pruning behavior more closely emulates command line git.
-     *
-     * This has been fixed using a workaround.
+     * JGit 3.3.0 thru 4.5.4 "prune during fetch" prunes more remote
+     * branches than command line git prunes during fetch.  JGit 5.0.2
+     * fixes the problem.
+     * Refer to https://bugs.eclipse.org/bugs/show_bug.cgi?id=533549
+     * Refer to https://bugs.eclipse.org/bugs/show_bug.cgi?id=533806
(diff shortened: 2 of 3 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -464,15 +464,23 @@ private void check_remote_url(final String repositoryName) throws InterruptedExc
...
 
+    private Collection<String> getBranchNames(Set<Branch> branches) {
+        return Collections2.transform(branches, GitObject::getName);
+    }
+
+    private void assertBranchCount(Set<Branch> branches, int expectedCount) {
+        assertEquals("Unexpected branch count in " + getBranchNames(branches), expectedCount, branches.size());
+    }
+
     private void assertBranchesExist(Set<Branch> branches, String ... names) throws InterruptedException {
```
darxriggs: This is like Hamcrest's [hasItems(...)](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#hasItems(T...)).
I suggest to use it internally due to the fact that it reports all missing branches (not only the first one missing).

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -464,15 +464,23 @@ private void check_remote_url(final String repositoryName) throws InterruptedExc
...
-        Collection<String> branchNames = Collections2.transform(branches, GitObject::getName);
+        Collection<String> branchNames = getBranchNames(branches);
         for (String name : names) {
             assertTrue(name + " branch not found in " + branchNames, branchNames.contains(name));
         }
     }
 
     private void assertBranchesNotExist(Set<Branch> branches, String ... names) throws InterruptedException {
```
darxriggs: This is similar to [`assertBranchesExist(...)`](https://github.com/jenkinsci/git-client-plugin/pull/351/files#diff-963c705aa3f7bb6ac484408977597efbR475) with Hamcrest's [not(...)](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#not(org.hamcrest.Matcher)) sneaked in.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -1327,39 +1332,33 @@ public void test_fetch_with_prune() throws Exception {
         ObjectId newAreaHead = newArea.head();
         Set<Branch> remoteBranches = newArea.git.getRemoteBranches();
         assertBranchesExist(remoteBranches, "origin/master", "origin/branch1", "origin/branch2", "origin/HEAD");
-        assertEquals("Wrong count in " + remoteBranches, 4, remoteBranches.size());
+        assertBranchCount(remoteBranches, 4);
```
darxriggs: I have seen this pattern (check for branch names + branch count) here and in some other tests.
Currently it always requires 2 assertions where just 1 fact should be asserted - that the branches match exactly independent of their order.
Hamcrest's [containsInAnyOrder(...)](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#containsInAnyOrder(T...)) is precisely for this purpose.

pom.xml
```diff
@@ -48,7 +48,7 @@
     <argLine>-Dfile.encoding=${project.build.sourceEncoding}</argLine>
     <jenkins.version>2.60.3</jenkins.version>
     <java.level>8</java.level>
-    <jgit.version>5.0.2.201807311906-r</jgit.version>
+    <jgit.version>5.1.0.201808281540-m3</jgit.version>
```
darxriggs: Are we fine with only a jgit milestone release?

MarkEWaite: Not for a merge to the master branch.  I wanted to confirm that the fix in JGit really fixed the problem I had seen.  It did that.  

This will request will wait until the release of either JGit 5.1 or JGit 5.0.3.

darxriggs: JGit v5.0.3.201809091024-r got released.

PR #344, opened 2018-07-29
shallow submodule update

darxriggs:
This implements https://issues.jenkins-ci.org/browse/JENKINS-21248.

The original implementation is by @fujii - see pull request #303.
I rebased his commit 89ba2d3ff38ec5bc870441ec06b73fbfcab6743e on master (without changes) and added improvements via new commits.
The tests are working fine now with a locally created git repository containing a submodule.

PR #352, opened 2018-08-15
Check for exceptions when updating submodules

nre-ableton:
This PR fixes an error introduced in https://github.com/jenkinsci/git-client-plugin/pull/348 where exceptions arising in git submodule update commands would be mistakenly suppressed. A test case has been added to verify the new behavior is correct.

Also, another fix is included which provides proper names to the threads in the `ExecutorService` pool via the `NamingThreadFactory` class.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1,6 +1,4 @@
-
 package org.jenkinsci.plugins.gitclient;
-
 
 import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
@@ -38,6 +36,7 @@ import org.eclipse.jgit.lib.RepositoryBuilder;
 import org.eclipse.jgit.transport.RefSpec;
 import org.eclipse.jgit.transport.RemoteConfig;
 import org.eclipse.jgit.transport.URIish;
+import org.jenkinsci.plugins.gitclient.cgit.GitCommandsExecutor;
 import org.kohsuke.stapler.framework.io.WriterOutputStream;
 
 import java.io.*;
@@ -67,8 +66,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
@@ -1264,12 +1262,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 Pattern pattern = Pattern.compile(SUBMODULE_REMOTE_PATTERN_STRING, Pattern.MULTILINE);
                 Matcher matcher = pattern.matcher(cfgOutput);
 
-                ExecutorService executorService;
-                if (threads > 1) {
-                    executorService = Executors.newFixedThreadPool(threads);
-                } else {
-                    executorService = Executors.newSingleThreadExecutor();
-                }
+                List<Callable<String>> commands = new ArrayList<>();
 
                 while (matcher.find()) {
                     ArgumentListBuilder perModuleArgs = args.clone();
@@ -1306,17 +1299,11 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                     perModuleArgs.add(sModulePath);
                     StandardCredentials finalCred = cred;
                     URIish finalUrIish = urIish;
-                    executorService.submit(() -> {
-                        try {
-                            launchCommandWithCredentials(perModuleArgs, workspace, finalCred, finalUrIish, timeout);
-                        } catch (InterruptedException e) {
-                            throw new GitException("Interrupted while updating submodule for " + sModuleName);
-                        }
-                    });
+
+                    commands.add(() -> launchCommandWithCredentials(perModuleArgs, workspace, finalCred, finalUrIish, timeout));
                 }
 
-                executorService.shutdown();
-                executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
+                new GitCommandsExecutor(threads, listener).invokeAll(commands);
             }
         };
     }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -283,8 +283,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         this.listener = listener;
         this.gitExe = gitExe;
         this.environment = environment;
-        
-        if( isZos() && System.getProperty("ibm.system.encoding") != null ) { 
+
+        if( isZos() && System.getProperty("ibm.system.encoding") != null ) {
             this.encoding = Charset.forName(System.getProperty("ibm.system.encoding")).toString();
         } else {
             this.encoding = Charset.defaultCharset().toString();
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1143,7 +1143,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             private Map<String, String> submodBranch   = new HashMap<>();
             private Integer timeout;
             private Integer depth = 1;
-            private Integer threads = 1;
+            private int threads = 1;
 
             @Override
             public SubmoduleUpdateCommand recursive(boolean recursive) {
@@ -1194,7 +1194,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             }
 
             @Override
-            public SubmoduleUpdateCommand threads(Integer threads) {
+            public SubmoduleUpdateCommand threads(int threads) {
                 this.threads = threads;
                 return this;
             }
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -2631,6 +2631,21 @@ public abstract class GitAPITestCase extends TestCase {
         assertFalse("shallow file existence: " + shallow, w.exists(shallow));
     }
 
+    @NotImplementedInJGit
+    public void test_submodule_update_with_error() throws Exception {
+        w.git.clone_().url(localMirror()).execute();
+        w.git.checkout().ref("origin/tests/getSubmodules").execute();
+        w.rm("modules/ntp");
+        w.touch("modules/ntp", "file that interferes with ntp submodule folder");
+
+        try {
+            w.git.submoduleUpdate().execute();
+            fail("Did not throw expected submodule update exception");
+        } catch (GitException e) {
+            assertThat(e.getMessage(), containsString("Command \"git submodule update modules/ntp\" returned status code 1"));
+        }
+    }
+
     public void test_submodule_update_shallow() throws Exception {
         WorkingArea remote = setupRepositoryWithSubmodule();
         w.git.clone_().url("file://" + remote.file("dir-repository").getAbsolutePath()).repositoryName("origin").execute();
--- a/pom.xml
+++ b/pom.xml
@@ -163,6 +163,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.22.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>nl.jqno.equalsverifier</groupId>
       <artifactId>equalsverifier</artifactId>
       <version>3.0.2</version>
--- a/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
@@ -2201,7 +2201,7 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             }
 
             @Override
-            public org.jenkinsci.plugins.gitclient.SubmoduleUpdateCommand threads(Integer threads) {
+            public org.jenkinsci.plugins.gitclient.SubmoduleUpdateCommand threads(int threads) {
                 // TODO: I have no idea if JGit can update submodules in parallel
                 // It might work, or it might blow up horribly. This probably depends on
                 // whether JGit relies on any global/shared state. Since I have no
--- a/src/main/java/org/jenkinsci/plugins/gitclient/SubmoduleUpdateCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/SubmoduleUpdateCommand.java
@@ -83,5 +83,5 @@ public interface SubmoduleUpdateCommand extends GitCommand {
      * @param threads number of threads to use for updating submodules in parallel
      * @return a {@link org.jenkinsci.plugins.gitclient.SubmoduleUpdateCommand} object.
      */
-    SubmoduleUpdateCommand threads(Integer threads);
+    SubmoduleUpdateCommand threads(int threads);
 }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/SubmoduleUpdateCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/SubmoduleUpdateCommand.java
@@ -78,7 +78,8 @@ public interface SubmoduleUpdateCommand extends GitCommand {
     SubmoduleUpdateCommand depth(Integer depth);
 
     /**
-     * Update submodules in parallel with the given number of threads.
+     * Update submodules in parallel with the given number of threads. Note that this parallelism only applies to the
+     * top-level submodules of a repository. Any submodules of those submodules will be updated serially.
      *
      * @param threads number of threads to use for updating submodules in parallel
      * @return a {@link org.jenkinsci.plugins.gitclient.SubmoduleUpdateCommand} object.
--- a/src/main/java/org/jenkinsci/plugins/gitclient/cgit/GitCommandsExecutor.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/cgit/GitCommandsExecutor.java
@@ -1,0 +1,91 @@
+package org.jenkinsci.plugins.gitclient.cgit;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionService;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorCompletionService;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.util.concurrent.MoreExecutors;
+import hudson.model.TaskListener;
+import hudson.plugins.git.GitException;
+import hudson.util.DaemonThreadFactory;
+import hudson.util.ExceptionCatchingThreadFactory;
+import hudson.util.NamingThreadFactory;
+
+/**
+ * This executor can invoke multiple git commands in parallel using threads.
+ * <p>
+ * If threads = 1 the caller thread is used.
+ * If a git command fails, invocation of all running and not yet started commands is stopped.
+ */
+public class GitCommandsExecutor {
+
+    private final int threads;
+    private final TaskListener listener;
+
+    public GitCommandsExecutor(int threads, TaskListener listener) {
+        this.threads = Math.max(1, threads);
+        this.listener = listener;
+    }
+
+    public <T> void invokeAll(Collection<Callable<T>> commands) throws GitException, InterruptedException {
+        ExecutorService executorService = null;
+        try {
+            if (threads == 1) {
+                executorService = MoreExecutors.sameThreadExecutor();
+            } else {
+                ThreadFactory threadFactory = new ExceptionCatchingThreadFactory(new NamingThreadFactory(new DaemonThreadFactory(), GitCommandsExecutor.class.getSimpleName()));
+                executorService = Executors.newFixedThreadPool(threads, threadFactory);
+            }
+            invokeAll(executorService, commands);
+        } finally {
+            if (executorService != null) {
+                executorService.shutdownNow();
+                if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) {
+                    listener.getLogger().println("[WARNING] Threads did not terminate properly");
+                }
+            }
+        }
+    }
+
+    private <T> void invokeAll(ExecutorService executorService, Collection<Callable<T>> commands) throws InterruptedException {
+        CompletionService<T> completionService = new ExecutorCompletionService<>(executorService);
+        Iterator<Callable<T>> remainingCommands = commands.iterator();
+        int nCommands = commands.size();
+
+        for (int i = 0; i < threads && i < nCommands; i++) {
+            submitRemainingCommand(completionService, remainingCommands);
+        }
+
+        for (int i = 0; i < nCommands; i++) {
+            checkResult(completionService.take());
+            submitRemainingCommand(completionService, remainingCommands);
+        }
+    }
+
+    private <T> void submitRemainingCommand(CompletionService<T> completionService, Iterator<Callable<T>> remainingCommands) {
+        if (remainingCommands.hasNext()) {
+            completionService.submit(remainingCommands.next());
+        }
+    }
+
+    private <T> void checkResult(Future<T> result) throws InterruptedException {
+        try {
+            result.get();
+        } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof InterruptedException) {
+                throw (InterruptedException) new InterruptedException().initCause(cause);
+            } else {
+                throw new GitException(cause);
+            }
+        }
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/gitclient/cgit/package-info.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/cgit/package-info.java
@@ -1,0 +1,4 @@
+/**
+ * Git client API implementation with <a href="https://git-scm.com/">cgit</a>.
+ */
+package org.jenkinsci.plugins.gitclient.cgit;
--- a/src/test/java/org/jenkinsci/plugins/gitclient/cgit/GitCommandsExecutorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/cgit/GitCommandsExecutorTest.java
@@ -1,0 +1,268 @@
+package org.jenkinsci.plugins.gitclient.cgit;
+
+import static java.util.Arrays.asList;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.lessThan;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.theInstance;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.Mockito.when;
+
+import java.io.PrintStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import hudson.model.TaskListener;
+import hudson.plugins.git.GitException;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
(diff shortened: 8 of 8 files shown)


nre-ableton:
ping @darxriggs, @MarkEWaite, @samrocketman, thanks!

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -23,8 +23,7 @@
 import hudson.plugins.git.IGitAPI;
 import hudson.plugins.git.IndexEntry;
 import hudson.plugins.git.Revision;
-import hudson.util.ArgumentListBuilder;
-import hudson.util.Secret;
+import hudson.util.*;
```
MarkEWaite: Wildcard imports are generally discouraged, particularly for a package as broad as hudson.util.  I don't see any change that makes this change compelling.

Help me understand why it was switched to a wildcard import?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -58,10 +57,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
MarkEWaite: I think I understand this wildcard import, though even here I think it would be better to list all the imports rather than use a wildcard.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
samrocketman: I also dislike wildcard imports.  As someone who regularly digs through source code wildcard imports make any code digging expensive time-wise if there’s many wildcard imports and no known package without digging.

samrocketman:
Generally looks Ok to me other than the imports.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
nre-ableton: I'm using IntelliJ, and the default setting is to use a wildcard after >=5 imports from the same namespace have been used. So this refactoring happened automatically (though I did stage/commit it manually, of course). I did a quick code grep for `.*;` and found several other instances in the codebase where wildcard imports were being used, so I thought it would be ok.

Anyways, I don't have a strong opinion one way or another on the matter, I'd be happy to remove the wildcard and explicitly list the imports. I'll push a fixup for this.

nre-ableton:
@MarkEWaite @samrocketman Ok, I've fixed the imports (and also removed an unused import which snuck in there).

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1228,17 +1233,31 @@ else if (!referencePath.isDirectory())
...
-                            launchCommandWithCredentials(perModuleArgs, workspace, finalCred, finalUrIish, timeout);
+                            return launchCommandWithCredentials(perModuleArgs, workspace, finalCred, finalUrIish, timeout);
                         } catch (InterruptedException e) {
                             throw new GitException("Interrupted while updating submodule for " + sModuleName);
                         }
                     });
                 }
 
+                List<Future<String>> futures = executorService.invokeAll(launchCommands);
                 executorService.shutdown();
```
darxriggs: About the complete shutdown procedure I am not sure yet. It requires some more thought and testing.
Please have a look at ExecutorService [Javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html) under "Usage Examples" -> the two phase shutdown. The method `shutdownNow()` should actually be called to really try to stop running processes. The method `shutdown()` just prevents new ones to start running.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1228,17 +1233,31 @@ else if (!referencePath.isDirectory())
...
+                    executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
+                } catch (InterruptedException e) {
+                    throw new GitException(e);
+                }
+
+                futures.forEach((future) -> {
+                    try {
+                        future.get();
+                    } catch (Exception e) {
+                        throw new GitException(e);
```
darxriggs: Have a look at [ExecutionException](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutionException.html) how to get to original exception in case a thread threw one.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -2638,6 +2638,17 @@ public void test_submodule_update() throws Exception {
         assertFalse("shallow file existence: " + shallow, w.exists(shallow));
     }
 
+    @NotImplementedInJGit
+    public void test_submodule_update_with_error() throws Exception {
+        w.init();
+        try {
+            w.git.addSubmodule("non-existent-submodule-url", "submodule-dir");
```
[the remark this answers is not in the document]
nre-ableton: Ok, this works for me. Thanks for the help!

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1186,9 +1182,10 @@ else if (!referencePath.isDirectory())
                 Pattern pattern = Pattern.compile(SUBMODULE_REMOTE_PATTERN_STRING, Pattern.MULTILINE);
                 Matcher matcher = pattern.matcher(cfgOutput);
 
+                List<Callable<String>> launchCommands = new ArrayList<>();
                 ExecutorService executorService;
                 if (threads > 1) {
-                    executorService = Executors.newFixedThreadPool(threads);
+                    executorService = Executors.newFixedThreadPool(threads, new ExceptionCatchingThreadFactory(new NamingThreadFactory(new DaemonThreadFactory(), "GitSubmoduleUpdate")));
                 } else {
                     executorService = Executors.newSingleThreadExecutor();
```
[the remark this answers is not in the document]
nre-ableton: Ok, I'll make this refactoring. I thought that `newSingleThreadExecutor` would use the current thread, but reading the API documentation, it does indeed create a new thread (which would be unnamed).

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1228,17 +1233,31 @@ else if (!referencePath.isDirectory())
                     perModuleArgs.add(sModulePath);
                     StandardCredentials finalCred = cred;
                     URIish finalUrIish = urIish;
-                    executorService.submit(() -> {
+
+                    launchCommands.add(() -> {
                         try {
-                            launchCommandWithCredentials(perModuleArgs, workspace, finalCred, finalUrIish, timeout);
+                            return launchCommandWithCredentials(perModuleArgs, workspace, finalCred, finalUrIish, timeout);
                         } catch (InterruptedException e) {
```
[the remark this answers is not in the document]
nre-ableton: Good catch, it should not be required here anymore. I agree that the simpler syntax would be preferable.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
nre-ableton: I'm pretty sure that we want `shutdown`, and not `shutdownNow`. From the documentation:

> The shutdown() method will allow previously submitted tasks to execute before terminating, while the shutdownNow() method prevents waiting tasks from starting and attempts to stop currently executing tasks.

Our test case is quite simple, so let's imagine a larger example. Let's say the repository has 30 submodules, and the thread pool has 5 threads. So the first 5 get scheduled right away when calling `invokeAll`. If we call `shutdownNow`, these first 5 commands run to completion, but not the next 25, since they are in the waiting state. These commands will be cancelled, which is not what we want. Likewise, we don't want to stop any running commands either. Basically, we just want to block until everything finishes executing.

Therefore, we do indeed want `shutdown`, since the idea is we want all waiting tasks to finish executing. FWIW, I can verify that this behavior works in the above example, since I built HPI plugins for this plugin and `git-plugin` and tested them on a Jenkins instance with one of the big repositories at my company, which has >30 submodules. Updating submodules went much faster than serially, and all submodules were correctly updated in the checkout phase of the job.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
nre-ableton: Ok, let's collapse this to `catch (ExecutionException | InterruptedException e)`.

nre-ableton:
@darxriggs Thanks for the review! I've pushed a pair of fixups based on your suggestions.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1186,12 +1191,9 @@ else if (!referencePath.isDirectory())
...
 
-                ExecutorService executorService;
-                if (threads > 1) {
-                    executorService = Executors.newFixedThreadPool(threads);
-                } else {
-                    executorService = Executors.newSingleThreadExecutor();
-                }
+                List<Callable<String>> launchCommands = new ArrayList<>();
+                ExceptionCatchingThreadFactory threadFactory = new ExceptionCatchingThreadFactory(new NamingThreadFactory(new DaemonThreadFactory(), "GitSubmoduleUpdate"));
+                ExecutorService executorService = threads > 1 ? Executors.newFixedThreadPool(threads, threadFactory) : Executors.newSingleThreadExecutor();
```
darxriggs: I was still thinking more about the default case (threads = 1). I would really like to behave it like before - without any new threads beeing spawned, to reduce non-required resource costs. There may already be a simple solution to this - replace Java's `Executors.newSingleThreadExecutor()` with Guava's `MoreExecutors.sameThreadExecutor()`.

MarkEWaite: Adding guava as a git plugin dependency seems like a high price to pay for the case of a single submodule.  Have I missed that it is already an existing dependency somehow?

darxriggs: Guava (package `com.google.common`) is already used in some classes.
Its provided via the parent pom: `org.jenkins-ci.plugins:plugin:pom:3.19` -> `org.jenkins-ci.main:jenkins-core:jar:2.60.3` ->`com.google.guava:guava:jar:11.0.1`

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
darxriggs: To elaborate a bit more:

Method `shutdown()` should be called right after submitting all commands, to prevent new submissions. It is similar to making the `ExecutorService` immutable, as it rejects any new tasks afterwards. As the `ExecutorService` has just method scope here and "we know what we are doing", calling this method would not be required.

Method `shutdownNow()` should be called before leaving the `execute()` method, to try to stop any running task. This should be done after performing the `Future::get` calls, otherwise running tasks may be stopped. `Future::get` is blocking, so after retrieving the results all tasks should anyway have been finished, as a timeout is used inside. So calling `shutdownNow()` is mainly for handling error cases.

These points should be assured and tested:
* `Exceptions` are mapped properly from the `Future::get` to maintain existing behaviour
* `ExecutorService` is shut down with all its running threads in any case to prevent thread leakage
* `ExectorService` and its threads have to be used in a way that doesn't hinder Jenkins to shutdown

darxriggs: @nre-ableton BTW, are you aware of the shallow submodule pull request #348 that got merged some days ago. This should reduce execution time for submodules even more than parallel updates.

nre-ableton: @darxriggs I assume that you mean #344, not #348, right? Anyways, yes, I'm aware of it, and it's a great improvement to the plugin. For the repositories that I work with, parallelization is still more important due to the large number of submodules. This is mostly due to the fact that git performance on Windows is quite poor (we must build on Linux/Mac/Win in parallel). Shallow cloning is a bit faster on Windows, but not by much.

darxriggs: Right, #344 is what I meant.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
nre-ableton: Sounds good, I also think it makes sense to continue in the current thread in this case. I will remove the ternary one-liner refactoring as well.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
nre-ableton: @darxriggs Ok, I didn't realize that `invokeAll` would block until all threads completed (my Java is a bit rusty, apologies). Anyways I agree that in this case we should use `shutdownNow`.

nre-ableton:
@darxriggs @MarkEWaite Ok, I've rebased + squashed, since the fixups were getting a bit out of hand. Mind taking a look at the new patch? What else needs to be done for us to be able to merge this patch?

darxriggs:
Currently I am bit short of time. But there are still some changes required. I'll keep you posted.

darxriggs:
@nre-ableton do you work on the mentioned issues?

I also started some fixes and tests locally. Threading is hard. So it will take some time.

nre-ableton:
@darxriggs No, I haven't had time to go through your points yet, but I will do so soon. I have a few questions for you as well.

> 1. Not setting a value results in a NullPointerException in the git-client-plugin code.

I'm a bit confused here; which value are you referring to?

> 2. Manually aborting a job run (threads=1), just interrupts the current submodule update and it goes on with updating the next submodule.
> 3. Manually aborting a job run (threads=2) many times until all submodule updates are interrupted, does not abort the job run but runs into the plugin's retry mechanism (enabled by default), starting with a fresh workspace from scratch, updating all submodules.

How can we detect if a job run has been aborted? Will a certain type of exception be thrown that we can expect?

> 4. Even if one submodule update fails, all commands are run to the end. This means unneccessary waiting times in case of errors.

True; can fix.

> 5. This approach doesn't apply to recursive submodules.

IMHO this is fine, though we should probably document the behavior. However, I'm probably biased towards my specific use-case. In any case, supporting recursive submodules is likely to be a big headache. Are you fine with leaving it as-is and just documenting this behavior as such?

darxriggs:
**NullPointerException**
I updated my comment above. I was referring to the `threads` value.
As suggested in https://github.com/jenkinsci/git-plugin/pull/612#discussion_r217908351, this should be solved by changing the interface `public SubmoduleUpdateCommand threads(Integer threads)` to `public SubmoduleUpdateCommand threads(int threads)` and just using `int` internally.

**aborting jobs**
I would expect a `java.lang.InterruptedException`.

**recursive submodules**
I had a short look at it. With the current implementation and the feature to provide different credentials for each submodule it is quite hard to implement parallelity for recursive submodules.
So properly documenting the feature scope/resctriction is be fine.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1111,15 +1111,15 @@ public void submoduleSync() throws GitException, InterruptedException {
...
-            int threads = 1;
+            private boolean recursive                      = false;
+            private boolean remoteTracking                 = false;
+            private boolean parentCredentials              = false;
+            private boolean shallow                        = false;
+            private String  ref                            = null;
+            private Map<String, String> submodBranch   = new HashMap<>();
+            private Integer timeout;
+            private Integer depth = 1;
+            private Integer threads = 1;
```
darxriggs: @nre-ableton could you replace this merge commit with a rebase on master and correctly resolve the merge conflict?

MarkEWaite: @darxriggs I have no disagreement with a rebase.  

Sorry that I missed the type change from Integer to int for `threads` in the merge commit.

nre-ableton:
@darxriggs Thanks for your patch! I had a look at that commit on your branch and will cherry-pick it to this branch (and also rebase, while I'm at it).

darxriggs:
Please test it thoroughly.
It is still WIP - one of the unit tests is still a blinker, the commit comment is not finished, ...

nre-ableton:
@darxriggs Will do -- I have no problem cleaning up the commit message nor testing the patch. However, I'm unclear on which test is unstable. Is it in `GitCommandsLauncherTest`? I ran that class in IntelliJ with the "repeat until failure" mode and let it fly for several minutes and didn't get a single failure. Which test did you mean?

darxriggs:
This test (https://github.com/darxriggs/git-client-plugin/commit/9363f4dd0288355bc59a8893aeede84c370b626d#diff-f109f6ec426f8eedea9f0ee47ca8a0f1R102) was failing a lot of times.

I now updated my commit:
* performed small changes
* fixed the blinker test
* add more tests
* updated the commit comment

I also created a repository (https://github.com/darxriggs/repo_with_submodules) with some big random binary files, so fetching takes some time. Have a look at the branches "3-submodules" and "invalid-url". With this repository all testcases mentioned above can be performed.

From time to time (5-10% of manual tests) I see a git process still running after former git process failed, the job run into a timeout or was canceled by the user. This is still an issue and I don't understand it yet.

nre-ableton:
@MarkEWaite @darxriggs Sorry for the radio silence on my end -- I've been quite busy with other work and haven't had time to test these changes properly. Even so, I don't think that I'm familiar enough with the codebase to confidently say that my efforts in that regard would be dependable.

I'm really not sure how we can proceed and get this patch merged. Would @darxriggs be interested in "adopting" this PR by cherry-picking/rebasing the patches into his branch and submitting it? Or should I rather rebase from that branch and force push here?

Either way, I'm unsure of how to proceed. Where do we go from here?

MarkEWaite:
I think you should integrate the changes from @darxriggs into your branch so that the pull request can be evaluated further by others.  I'd really like to include this change in the git client plugin 3.0 plugin that I want to deliver before end of December 2018.

I have some testing ideas and testing infrastructure that I can use for further evaluation.  If that testing shows the same types of issues that @darxriggs detected, it may also provide some hints of things that should be investigated further.

nre-ableton:
@MarkEWaite @darxriggs Ok, I've rebased this branch on `master`, and also squashed and reordered the commits. The `HEAD` commit basically replaces my previous commits in this branch, so I squashed them together. Hopefully the diff came out correctly.

nre-ableton:
@darxriggs that's perfectly understandable. Thanks so much for all of your help and time on this issue!

nre-ableton:
re-ping @MarkEWaite

MarkEWaite:
@nre-ableton this change is included in the git client plugin 3.0.0-beta6 release and is part of the testing plan.  Parallel submodule updates either needs to be included completely in git client plugin 3.0.0 and git plugin 4.0.0 (with this change and or the git client plugin change that initially introduced parallel submodule updates needs to be removed before release.

I welcome your help testing git client plugin 3.0.0-beta6 and git plugin 4.0.0-beta4 in your environment.

nre-ableton:
@MarkEWaite I will happily to test these beta plugins and report back here. Is there anything else I can do to help get this change merged?

MarkEWaite:
@nre-ableton general testing for compatibility is helpful.

The most valuable would be to stress test and duplicate the problem reported by @darxriggs and propose a fix for it.

kshultzCB:
Which of darxriggs' issues is the one remaining to be recreated? I've got some infra and would like to take a shot at it as well.

MarkEWaite:
@kshultzCB the remaining open item is:

> From time to time (5-10% of manual tests) I see a git process still running after a former git process failed, the job run into a timeout or was canceled by the user. This is still an issue and I don't understand it yet.

I've created the [JENKINS-44720 branch](https://github.com/MarkEWaite/jenkins-bugs/tree/JENKINS-44720) in my jenkins-bugs repository.  It includes multiple submodules referencing different branches from the same repository.  If not using a reference repository (for example on Windows) and not using a shallow clone, the clone of each submodule is pulling ~40 MB of repository data.

Refer to the [global pipeline library](https://github.com/MarkEWaite/jenkins-pipeline-utils) which provides the shared library functions used in that repository.  The folder and jobs that reference that library are in the [lts-with-plugins branch](https://github.com/MarkEWaite/docker-lfs/tree/lts-with-plugins/ref) of my [docker-lfs repository](https://github.com/MarkEWaite/docker-lfs/).

kshultzCB:
Awesome, thanks @MarkEWaite . That's very helpful. That's what I figured, just wanted to be sure I was looking for the right thing.

nre-ableton:
So I did some basic testing of this feature using my company's repository and git setup. Specifically, we are testing on the following platforms:

- Ubuntu 16.04
- Mac OS X 10.12
- Windows 10

Our repository is rather large (hence the motivation to add this feature). It's about 1Gb excluding submodules, of which there are around 30 at the top-level and about 40 including all children, recursively.

The basic functionality of this feature works fine on all OS's that we test on. For the curious, the performance boost is quite nice. We see a 3x speed boost on Linux, 2.6x on Mac, and 2x on Windows. I did some manual cancellation tests on Linux & Mac and didn't get any orphaned processes. Following this, I made a Jenkinsfile which looked roughly like the following:

```groovy
while(true) {
  try {
    parallel(failFast: true,
      checkout: {
        node('linux') {
          // This job is configured to use 8 threads for cloning submodules.
          // We are using reference repositories for the main repo, but not
          // for submodules.
          checkout scm
        }
      },
      die: {
        // This should interrupt the above checkout scm step while it is
        // cloning submodules. The initial repo clone and checkout takes
        // around 20 seconds, while the entire step takes roughly 1 minute
        // to complete on our Linux build nodes.
        sleep 40
        throw new Exception('YOU DIE NOW')
      },
    )
  } catch (Exception ignored) {
    echo 'STILL ALIVE'
  }
}
```

I ran a build on this branch for about an hour and a half, which yielded over 120 iterations, and I did not see any hanging `git` processes on the build node. That said, my tests should be taken with a grain of salt given the small sample size. I will continue to run the job overnight and see if I observe anything irregular. I will also try to make similar stress tests on Mac and Windows and report back with any interesting results.

MarkEWaite:
That is great @nre-ableton .  You can create a single "mega-reference" repository which includes content from your parent repository and all of your submodule repositories.  The technique is to perform a bare clone to create the reference repository from the parent, then add a remote for each of the submodules, update the definition of the parent repository in the config file to match the new remotes, and do a "git fetch --all" into that mega-repo.

kshultzCB:
@MarkEWaite - I'm pretty sure I've found an issue. The steps to recreate are pretty straightforward - I just need to make sure I haven't got something unintentionally screwy in the two repositories I'm using for this test.

kshultzCB:
Here's what I currently have running. 

1. I've forked Mark's jenkins-bugs repository [here](https://github.com/kshultzCB/jenkins-bugs), and have created three branches based on [his JENKINS-44720](https://github.com/MarkEWaite/jenkins-bugs/tree/JENKINS-44720) branch as referenced [above](https://github.com/jenkinsci/git-client-plugin/pull/352#issuecomment-452282354).

2. Those three branches have Jenkinsfiles which target each of: the Jenkins master, a remote Linux node with 8GB of RAM, and a remote Windows node with 4GB of RAM. [Sample Jenkinsfile here](https://github.com/kshultzCB/jenkins-bugs/blob/windows-karl/Jenkinsfile) - the only difference is the [node label](https://github.com/kshultzCB/jenkins-bugs/blob/windows-karl/Jenkinsfile#L13).

3. The jobs are running every 10 minutes.

So far, I have seen no leftover `git` processes running between jobs.

darxriggs:
Let me try to provide the steps to reproduce the problem in more detail.

**Steps**
* checkout https://github.com/jenkinsci/git-plugin/pull/612
* update pom.xml
  * org.jenkins-ci.plugins:plugin:3.32 (because in 3.29 `hpi:run` is broken)
  * org.jenkins-ci.plugins:git-client:3.0.0-beta6-rc1831.3506eebd3e1f (the incrementals build of https://github.com/jenkinsci/git-client-plugin/pull/352)
* `mvn hpi:run`
* set 50 executors for master
* create Freestyle job
  * Execute concurrent builds if necessary
  * git
    * Repository URL: https://github.com/darxriggs/repo_with_submodules
    * Branch Specifier: */3-submodules
    * Advanced sub-module behaviours
    * Wipe out repository & force clone
* start this job 50 times (all concurrent)
* cancel all concurrent 50 runs
* `ps -ef | grep git`

**Result**
There are still 1-3 running git instances.

As described in my [comment](https://github.com/jenkinsci/git-client-plugin/pull/352#issuecomment-427460686), the problem occurs if the job terminated abnormally - e.g.
a) a git process failed
b) the job run into a timeout
c) the job was canceled by the user
If the git command returned successfully, everything is fine.

I hope this provides further insight.

darxriggs:
Hi @kshultzCB, I updated my original description slightly.

Just add "Advanced sub-module behaviours" to enable submodules.
It doesn't matter if the field "Number of threads to use when updating submodules" is filled or not.

What hasn't been verified yet: **Is the reported issue really about this pull request?**

I checked this with other beta versions, that don't yet use the parallel git-client plugin code (4.0.0-beta1,
4.0.0-beta2, 4.0.0-beta3).

Answer: **No.** The issue is already present in code prior to this pull request.

I will provide some further analysis soon.

darxriggs:
# Further Analysis

The issue (remaining `git` processes) manifests most of the time but not always.
Therefore multiple iterations may be required to reproduce it.

Some scripts have been created to check the processes - see [gist](https://gist.github.com/darxriggs/9fbd6eb35053148bc4ba4436f11e4b60).

* get-process-counts
* get-process-tree-git
* get-process-tree-jenkins

## Process Counts

`watch ./get-process-counts` has been used to get an overview.

Output _before starting_ the 50 builds:

```
# processes #
git submodule update: 0
git:                  0
jenkins:              1
user:                 72
```

Output _after starting_ the 50 builds:
(It takes about 30s after starting the builds until all processes are started.)

```
# processes #
git submodule update: 50
git:                  500
jenkins:              501
user:                 572
```

Output _after stopping_ the 50 builds:
(It takes about 5s after stopping the builds until the processes are stopped.)
The `git` process count differs slightly from iteration to iteration but seems to always be a multiple of 4.

```
# processes #
git submodule update: 0
git:                  4
jenkins:              1
user:                 76
```

## Process Details

A detailed analysis revealed:
The git processes, that are running _after_ the builds have been stopped, are different ones than the ones that are running _before_ the builds have been stopped.

A custom git directory that contains the `BUILD_ID` has been used to track in which build a git process was started. The script `get-process-tree-git` has been used to list the git processes.

The below outputs have been stripped down a bit manually to only contain the relevant information.
They contain all the git processes that are assumed to belong to a single `git clone` operation which is still going on _after_ all builds have been stopped.

The 1st process tree shows that the `git clone` process is correctly a child of the `git submodule update` process tree. The 2nd process tree shows that the `git clone` process has PPID 1. This means that something is *not correct*.

git processes _before_ builds have been stopped:

```
  UID   PID  PPID  PGID TPGID CMD
 1000 57634 29477 29476 29476 git submodule update sub1
 1000 57640 57634 29476 29476  \_ /bin/sh /usr/lib/git-core/git-submodule update sub1
 1000 57930 57640 29476 29476      \_ /bin/sh /usr/lib/git-core/git-submodule update sub1
 1000 57932 57930 29476 29476      |   \_ git submodule--helper update-clone sub1
 1000 57939 57932 29476 29476      |       \_ /usr/lib/git-core/git submodule--helper clone --path sub1 --name sub1 --url https://github.com/darxriggs/repo
 1000 57948 57939 29476 29476      |           \_ /usr/lib/git-core/git clone --no-checkout --separate-git-dir <REDACTED>/workspace/project-with-checkout@26/subfolder-21982/.git/modules/sub1 -- https://github.com/darxriggs/repo <REDACTED>/workspace/project-with-checkout@26/subfolder-21982/sub1
 1000 57955 57948 29476 29476      |               \_ /usr/lib/git-core/git-remote-https origin https://github.com/darxriggs/repo
 1000 60017 57955 29476 29476      |                   \_ /usr/lib/git-core/git fetch-pack --stateless-rpc --stdin --lock-pack --thin --check-self-contained-and-connected --cloning --no-progress https://github.com/darxriggs/repo/
 1000 60254 60017 29476 29476      |                       \_ /usr/lib/git-core/git unpack-objects -q --pack_header=2,33
 1000 57931 57640 29476 29476      \_ /bin/sh /usr/lib/git-core/git-submodule update sub1
```

git processes _after_ builds have been stopped:

```
  UID   PID  PPID  PGID TPGID CMD
 1000 61218     1 29476 29476 /usr/lib/git-core/git clone --no-checkout --separate-git-dir <REDACTED>/workspace/project-with-checkout@26/subfolder-21982/.git/modules/sub1 -- https://github.com/darxriggs/repo <REDACTED>/workspace/project-with-checkout@26/subfolder-21982/sub1
 1000 61219 61218 29476 29476  \_ /usr/lib/git-core/git-remote-https origin https://github.com/darxriggs/repo
 1000 61242 61219 29476 29476      \_ /usr/lib/git-core/git fetch-pack --stateless-rpc --stdin --lock-pack --thin --check-self-contained-and-connected --cloning --no-progress https://github.com/darxriggs/repo/
 1000 61244 61242 29476 29476          \_ /usr/lib/git-core/git unpack-objects -q --pack_header=2,33
```

## Conclusion

There is a problem with disconnected `git` child processes which is not understood.

nre-ableton:
While I'm not entirely comfortable with the idea of shipping known, buggy code, it seems that we can establish that this is a prior issue and one not caused by this specific PR. (BTW @darxriggs thanks for the very detailed investigation!) Therefore, is there any reason *not* to go ahead and merge this PR (and the corresponding https://github.com/jenkinsci/git-plugin/pull/612)?

The hanging `git` processes issue is troubling, but since we now know that it isn't caused by this particular change, I wonder if we should instead create a JIRA issue and migrate the respective information from this PR to that.

PR #353, opened 2018-08-18
delete @tmp directory in workspace

rhubert:
Since 75ea3fe05 a @tmp directoy is created in the workspace but not cleaned up.
We need to clean this up for buildsystems hashing the workspace to find
binary artefacts.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1621,7 +1621,16 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         if (workspace == null) {
             return createTempFileInSystemDir(prefix, suffix);
         }
-        File workspaceTmp = new File(workspace.getAbsolutePath() + "@tmp");
+        // Not using workspace variable to create the @tmp directory
+        // cause this is not the job's workspace when cloning into
+        // subdirectories.
+        String workspaceDir;
+        if (environment != null) {
+            workspaceDir = environment.get("WORKSPACE", workspace.getAbsolutePath());
+        } else {
+            workspaceDir = workspace.getAbsolutePath();
+        }
+        File workspaceTmp = new File(workspaceDir + "@tmp");
         if (!workspaceTmp.isDirectory() && !workspaceTmp.mkdirs()) {
             if (!workspaceTmp.isDirectory()) {
                 return createTempFileInSystemDir(prefix, suffix);


MarkEWaite:
Thanks for the pull request.  Can you further describe the problem you're seeing?

The code in 75ea3fe05650fc6ca09046a72493e2b3f066fb98 is intended to create the temporary directory in the same **parent directory** as the workspace.  It is not intended to create the temporary directory in the workspace itself.

Are you seeing a directory named '@Tmp' in the workspace directory?  If so, then the logic which is being used to create that temporary directory needs to be changed.  If the workspace directory name is '/var/lib/jenkins/workspace/my-job' then the temporary directory should be named 
'/var/lib/jenkins/workspace/my-job@Tmp', not '/var/lib/jenkins/workspace/my-job/@Tmp'.

If the temporary directory is not placed in the correct parent directory, let's fix that problem, rather than attempting to delete a known temporary directory that is often used by other parts of the system.

rhubert:
We're using the `Checkout to subdirectory` option to checkout one ore more repo into the same workspace. So before 75ea3fe the workspace content was:

```
/var/lib/jenkins/workspace/my-job/foo
/var/lib/jenkins/workspace/my-job/bar
```

but now it's

```
/var/lib/jenkins/workspace/my-job/foo
/var/lib/jenkins/workspace/my-job/foo@tmp
/var/lib/jenkins/workspace/my-job/bar
/var/lib/jenkins/workspace/my-job/bar@tmp
```

Now our [buildtool](https://github.com/BobBuildTool/bob) hashes the workspace to calculate a buildID which afterwards is used to download binary artifacts already build by jenkins. As we do not have the @tmp for local builds binary artifacts are not working any more.

If I got you right the @tmp  should be next to the workspace? So even with `Checkout to subdirectory` option the workspace should look like:

```
/var/lib/jenkins/workspace/my-job/foo
/var/lib/jenkins/workspace/my-job/bar
/var/lib/jenkins/workspace/my-job@tmp
```
? This would also work as we only hash `/var/lib/jenkins/workspace/my-job/`. I'll try to update the PR soon..

MarkEWaite:
Thanks for the clarification.  Yes, that says that the temporary directory is being created inside the Jenkins workspace if checkout to subdirectory is used.

I suspect the same behavior happens if you use a Pipeline and perform the checkout inside a `dir('xxx') { }' block.  It is the incorrect behavior there as well.

I also suspect that there is a standard Jenkins function somewhere which should be called to get the temporary folder associated with the workspace of the current Jenkins job.

rhubert:
AFAICS the `workspace` used by the git-client-plugin is computed by the plugin using the git-client-plugin? So in the git-client-plugin it's not possible to know if the checkout is done using the job's workspace or a subdirectory?

The git-client-plugin itself has no possibility to get the jobs workspace using the JenkinsAPI as it do not have the `AbstractBuild` object. My only idea is to use the WORKSPACE environment variable. And I haven't found a Jenkins function to get a temporary folder.

MarkEWaite:
I created an [automated test case](https://github.com/MarkEWaite/jenkins-bugs/blob/1b55bc67aae0925c921bfc0c8ddef8c46b4835c0/Jenkinsfile#L62) to verify the problem.  I believe the test case is correct.  The test case performs parallel checkouts of three repositories into subdirectories of the current workspace while a "monitor' is watching for the unexpected existence of the temporary directories **inside** the workspace.  It finds those temporary directories in most executions, even with the 131ms sleep between checks

When I install a build of the plugin with this proposed change, it still fails that test.

It could be that the automated test is wrong, or it could be that there is some reason that the changes you're proposing are not having the expected result in my environment.

Could you investigate further to see if you can duplicate the types of failures that I'm seeing in my job?

MarkEWaite:
Closing for now, included in the "Later" milestone so that it can be reconsidered in the future.

mahaase:
Why this issues is closed? The problem still exists.

MarkEWaite:
It was closed and assigned the label `Later` so that it can be reopened in the future when I have time to review the pull request.  You are welcome (and encouraged) to apply the changes to a private copy of the plugin and use it.  Then you can report the results of your use to this pull request so that I and the other maintainers are aware of your results.

rhubert:
I tried to reproduce the issues described in https://github.com/jenkinsci/git-client-plugin/pull/353#issuecomment-431596471 but were not able to run the Jenkinsfile on my jenkins. I don't know where the `globalPipelineLibraryMarkEWaite` came from and also other thinks like https://github.com/MarkEWaite/jenkins-bugs/blob/1b55bc67aae0925c921bfc0c8ddef8c46b4835c0/Jenkinsfile#L28 are looking somehow strange. That's why I gave up as it was working with my own copy of this plugin.

@MarkEWaite  Maybe you can bring in some light to this enabling me to reproduce the issues on my system and to follow up on this topic?

MarkEWaite:
Sure!  

My Jenkins test setup is composed of:
1. [Public docker image lts-with-plugins branch](https://github.com/MarkEWaite/docker-lfs/tree/lts-with-plugins) that defines the Jenkins installation including plugins, pipeline shared libraries, folders, etc.  That image defines [`globalPipelineLibraryMarkEWaite`](https://github.com/MarkEWaite/docker-lfs/blob/79945d3361acff5e26c541b89cb1b08ba686055a/ref/jenkins.yaml#L193).  The image can be built with the command `./docker_build.py` from that repository.  It can be run with the command `./docker_run.py --clean`
2. [Private docker image lts-with-plugins-add-credentials-and-nodes branch](https://github.com/MarkEWaite/docker-private-lfs/tree/lts-with-plugins-add-credentials-and-nodes) that defines agents, credentials, more jobs, and other private information based on the contents of the public docker image and its lts-with-plugins branch
3. [Public Jenkins bugs repository branch JENKINS-47874](https://github.com/MarkEWaite/jenkins-bugs/blob/JENKINS-47874/Jenkinsfile) that attempts to explore the problem.  It relies on the public Docker image contents to create a branch for each bug that the repository is testing

rhubert:
Tanks! The reason for the failing test after installing the git-client-plugin with the suggested fix is another plugin called [durable-task-plugin](https://github.com/jenkinsci/durable-task-plugin/blob/master/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java#L373). This gets involved when using pipelines which is the reason I have not seen this before. As there might be other plugins creating the same `@tmp` directory and I'm not motivated to fix them all we'll update our tool to just ignore them.

PR #358, opened 2018-08-20
Command improvements

darxriggs:
* Make all instance variables private in all Command implementations
* Use @Override on all implemented methods in all Command implementations

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -341,41 +341,49 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             private boolean tags = true;
             private Integer depth = 1;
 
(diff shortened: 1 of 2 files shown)


darxriggs:
I guess this is due to historical reasons.

I also vote for non-anonymous classes. This would reduce the huge LOC number of the main implementation class and also increase testability.

MarkEWaite:
@Wadeck I believe the anonymous classes implementation was a choice that Nicolas de Loof and Kohsuke Kawaguchi made during initial implementation.  It was a reasonable match at that time to the use of anonymous classes in Jenkins as a whole.

Current biases in Jenkins are pushing strongly towards not using anonymous classes, especially in those cases that involve Jenkins remoting.  I'm not sure how large the change would be to switch away from anonymous inner classes, but I'm willing to consider it.

darxriggs:
Hi @MarkEWaite can this be merged?
I think there is not that much merge conflict potential with most of the latest still open pull request.

PR #361, opened 2018-09-03
z/OS Unix System services CLI git enablement

OldNavi:
Hi all,

I made git-client-plugin to work with z/OS Unix System Service with command line git implementation made by Rocket SW as it has addons to work correctly under z/OS unix. I had it running successfully almost for one year and decided to offer my changes to community.  Most code changes in this PR are related to the transcoding input/output  from git cli from/to IBM z/OS EBCDIC code page.  Kindly let me know if this PR can be accepted and what should i amend/changed to get this PR accepted.

PR #362, opened 2018-09-03
z/OS Unix System services CLI git enablement

OldNavi:
Hi all,

I made git-client-plugin to work with z/OS Unix System Service with command line git implementation made by Rocket SW as it has addons to work correctly under z/OS unix. I had it running successfully almost for one year and decided to offer my changes to community. Most code changes in this PR are related to the transcoding input/output from git cli from/to IBM z/OS EBCDIC code page. Kindly let me know if this PR can be accepted and what should i amend/changed to get this PR accepted.

PS: This PR is for master branch (3.0.0) - same as in PR #361

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -8,13 +8,10 @@ import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredenti
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.EnvVars;
-import hudson.FilePath;
-import hudson.Launcher;
+import hudson.*;
 import com.google.common.collect.Lists;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Launcher.LocalLauncher;
-import hudson.Util;
 import hudson.model.TaskListener;
 import hudson.plugins.git.Branch;
 import hudson.plugins.git.GitException;
@@ -169,6 +166,10 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     /* Package protected for testing */
     final static String SUBMODULE_REMOTE_PATTERN_STRING = SUBMODULE_REMOTE_PATTERN_CONFIG_KEY + "\\s+[^\\s]+$";
 
+
+    /* Encoding charset for z/OS, only on z/OS USS platform                 
+    */
+    private final String encoding;
     private void warnIfWindowsTemporaryDirNameHasSpaces() {
         if (!isWindows()) {
             return;
@@ -252,7 +253,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         this.listener = listener;
         this.gitExe = gitExe;
         this.environment = environment;
-
+        this.encoding = ZosCheck() ? "IBM1047" : Charset.defaultCharset().toString();
+        
         launcher = new LocalLauncher(IGitAPI.verbose?listener:TaskListener.NULL);
     }
 
@@ -804,15 +806,15 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     public void clean(boolean cleanSubmodule) throws GitException, InterruptedException {
         reset(true);
 	String cmd = "-fdx";
-	if (cleanSubmodule) cmd = "-ffdx";
-
+    if (cleanSubmodule) cmd = "-ffdx";
+    
 	launchCommand("clean", cmd);
-    }
+}
 
-    /**
-     * Remove untracked files and directories, including files listed
-     * in the ignore rules.
-     *
+/**
+ * Remove untracked files and directories, including files listed
+ * in the ignore rules.
+ *
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
      * @throws java.lang.InterruptedException if interrupted.
      */
@@ -1816,7 +1818,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     private File createSshKeyFile(SSHUserPrivateKey sshUser) throws IOException, InterruptedException {
         File key = createTempFile("ssh", ".key");
-        try (PrintWriter w = new PrintWriter(key, Charset.defaultCharset().toString())) {
+        try (PrintWriter w = new PrintWriter(key, this.encoding)) {
             List<String> privateKeys = sshUser.getPrivateKeys();
             for (String s : privateKeys) {
                 w.println(s);
@@ -1852,7 +1854,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     private File createWindowsSshAskpass(SSHUserPrivateKey sshUser) throws IOException {
         File ssh = createTempFile("pass", ".bat");
-        try (PrintWriter w = new PrintWriter(ssh, Charset.defaultCharset().toString())) {
+        try (PrintWriter w = new PrintWriter(ssh, this.encoding)) {
             // avoid echoing command as part of the password
             w.println("@echo off");
             // no surrounding double quotes on windows echo -- they are echoed too
@@ -1865,7 +1867,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
     private File createUnixSshAskpass(SSHUserPrivateKey sshUser) throws IOException {
         File ssh = createTempFile("pass", ".sh");
-        try (PrintWriter w = new PrintWriter(ssh, Charset.defaultCharset().toString())) {
+        try (PrintWriter w = new PrintWriter(ssh, this.encoding)) {
             w.println("#!/bin/sh");
             w.println("echo '" + quoteUnixCredentials(Secret.toString(sshUser.getPassphrase())) + "'");
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -8,13 +8,10 @@
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.EnvVars;
-import hudson.FilePath;
-import hudson.Launcher;
+import hudson.*;
```
MarkEWaite: Please don't switch to wildcard imports.  That change tends to result in unnecessary conflicts with other pull requests and forces work onto the authors of other pull requests.

I'll include a pointer to the SCM API guidelines in the CONTRIBUTING file so that others are less likely to have that problem

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -8,13 +8,10 @@
...
 import com.google.common.collect.Lists;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Launcher.LocalLauncher;
-import hudson.Util;
```
MarkEWaite: Don't make whitepace related changes without consultation with the maintainers.  There are enough open pull requests for the git client plugin that we can't make whitespace changes without creating conflicts in those other pull requests.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -804,15 +806,15 @@ public void execute() throws GitException, InterruptedException {
     public void clean(boolean cleanSubmodule) throws GitException, InterruptedException {
         reset(true);
 	String cmd = "-fdx";
-	if (cleanSubmodule) cmd = "-ffdx";
-
+    if (cleanSubmodule) cmd = "-ffdx";
```
MarkEWaite: Please don't make whitespace changes.  Even as ugly as the formatting is, we retain the formatting for existing code so that pending pull requests are not blocked by conflicts due to whitespace changes.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -804,15 +806,15 @@ public void execute() throws GitException, InterruptedException {
...
+
```
MarkEWaite: No trailing whitespace on lines, please.  The current whitespace setup is already bad enough without adding more cases of whitespace issues.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -804,15 +806,15 @@ public void execute() throws GitException, InterruptedException {
...
+    
 	launchCommand("clean", cmd);
-    }
+}
 
-    /**
-     * Remove untracked files and directories, including files listed
-     * in the ignore rules.
-     *
+/**
```
MarkEWaite: Remove this whitespace only change please.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2023,20 +2025,24 @@ private File createWindowsGitSSH(File key, String user) throws IOException {
         File ssh = createTempFile("ssh", ".bat");
 
         File sshexe = getSSHExecutable();
-
-        try (PrintWriter w = new PrintWriter(ssh, Charset.defaultCharset().toString())) {
+
```
MarkEWaite: No trailing white space on lines.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2046,7 +2052,7 @@ private File createUnixGitSSH(File key, String user) throws IOException {
             w.println("ssh -i \"" + key.getAbsolutePath() + "\" -l \"" + user + "\" -o StrictHostKeyChecking=no \"$@\"");
         }
         ssh.setExecutable(true, true);
-        //JENKINS-48258 git client plugin occasionally fails with "text file busy" error
+                //JENKINS-48258 git client plugin occasionally fails with "text file busy" error
```
MarkEWaite: Remove this whitespace change please

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2067,12 +2073,12 @@ private File createUnixGitSSH(File key, String user) throws IOException {
                 deleteTempFile(ssh_copy);
             }
             //Previous operation failed. Return original file
-            return ssh;
-        }
-		
-        return ssh_copy;
+        return ssh;
```
MarkEWaite: Remove this whitespace change please

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2067,12 +2073,12 @@ private File createUnixGitSSH(File key, String user) throws IOException {
...
     }
 
+    return ssh_copy;
```
MarkEWaite: Remove this whitespace change please

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2104,14 +2110,49 @@ private String launchCommandIn(ArgumentListBuilder args, File workDir, EnvVars e
...
+            if(ZosCheck() == true) {
+                // Another behavior on z/OS required due to the race condition
+                Launcher.ProcStarter p = launcher.launch().cmds(args.toCommandArray()).
+                        envs(freshEnv);
+                if (workDir != null) p.pwd(workDir);
+                p.readStdout().readStderr();
+                Proc prc = p.start();
+
+                status = prc.joinWithTimeout(usedTimeout, TimeUnit.MINUTES, listener);
+                BufferedReader brStdout = new BufferedReader(new InputStreamReader(prc.getStdout(),Charset.forName("IBM1047")));
```
MarkEWaite: Please use the Java 'try with resources' syntax so that we can reduce the risk of file descriptor leaks and remove the `close` calls at the end of the block.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2104,14 +2110,49 @@ private String launchCommandIn(ArgumentListBuilder args, File workDir, EnvVars e
...
-            if (workDir != null) p.pwd(workDir);
-            int status = p.start().joinWithTimeout(usedTimeout, TimeUnit.MINUTES, listener);
 
-            String result = fos.toString(Charset.defaultCharset().toString());
+            int status=0;
+            String result = "";
+            String errorString = "";
+            if(ZosCheck() == true) {
+                // Another behavior on z/OS required due to the race condition
+                Launcher.ProcStarter p = launcher.launch().cmds(args.toCommandArray()).
```
MarkEWaite: Please provide more details in this comment on the race condition that is requiring this difference of behavior.  For example:

* What is the race condition?
* Where is it described?
* What makes it unique to z/OS?
* How would we detect at some future time that the race condition no longer exists?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: Sorry, fixed now.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: ditto

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: Sorry, fixed now.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: ditto

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: ditto

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: ditto

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: ditto

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: ditto

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: ditto

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: Fixed.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
OldNavi: It is something internal with IBM Java on z/OS implementation - i didn't get much into details there. It seems Runtime.exec() behaves a little bit different when JVM does transcoding on z/OS. As result if i do us LocalProc implementation with StreamCopyThread - i'm getting garbage in a first few chars and i didn't found a way to overcome that easily.  Seems IBM Java implementation of Runtime.exec() is not that good when it have to deal with transcoding - that's why there is IBM z/OS Java own version of exec() in JZOS library - which do work great https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.zsecurity.api.80.doc/com.ibm.jzos/com/ibm/jzos/Exec.html  - but i don't want to add that kind of dependency in the project.

PR #368, opened 2018-09-26
Getting rid of static constant in z/OS encoding

OldNavi:


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -252,7 +252,8 @@ protected CliGitAPIImpl(String gitExe, File workspace, TaskListener listener, En
         this.listener = listener;
         this.gitExe = gitExe;
         this.environment = environment;
-        this.encoding = isZos() ? "IBM1047" : Charset.defaultCharset().toString();
+        this.encoding = isZos() && System.getProperty("ibm.system.encoding")!=null ?
```
MarkEWaite: I don't think that does what is needed in the rest of the file.

The rest of the file acts as though this.encoding is the String representation of a Charset name, yet this assignment will cause `this.encoding` to be assigned the value `true` on zOS and `false` on all other platforms (at least as far as I can tell).

I assume that methods which are using `encoding` continue to pass the unit tests because the called methods default to the system encoding when they detect an unrecognized encoding character set.

Please retain the prior meaning of this.encoding, since it is used in other places in the file.

OldNavi: Mmm... tertiary operations should take precedence if i'm not mistaken
image
image

OldNavi: but agree, better to rewrite for readability

MarkEWaite:
Ah, you're absolutely correct.  I misread the ternary operator use in my attempt to quickly review the code.

PR #370, opened 2018-09-30
Depend on Jenkins 2.107.3 and test 2.121.3

MarkEWaite:
Jenkins 2.121.3 is the most recent 2.121 release.

The Jenkins 2.138.1 release introduces a new requirement for instance-identity that needs more analysis before testing with 2.138.1

Jenkins 2.107 is a year newer than Jenkins 2.60.3 and is already superseded by two LTS releases.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -28,6 +28,7 @@ import hudson.util.Secret;
 import hudson.Proc;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.jgit.lib.Constants;
 import org.eclipse.jgit.lib.ObjectId;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2169,23 +2182,8 @@ private String launchCommandIn(ArgumentListBuilder args, File workDir, EnvVars e
...
-
-                StringBuilder stderrBuilder = new StringBuilder();
-                try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getStderr(), encoding))) {
-                    String line;
-                    while ((line = reader.readLine()) != null) {
-                        stderrBuilder.append(line);
-                    }
-                }
-                stderr = stderrBuilder.toString();
+                stdout = getZosStdio(process, encoding, false);
```
darxriggs: Why not `InputStream` as parameter?
```
stdout = getZosStdio(process.getStdout(), encoding);
stderr = getZosStdio(process.getStderr(), encoding);
```
The method is independent of ZOS. Therefore the name could be something like `readStreamIntoString()`.

There are even existing implementations for this - see https://stackoverflow.com/questions/309424/how-to-read-convert-an-inputstream-into-a-string-in-java
```
stdout = IOUtils.toString(process.getStdout(), encoding);
stderr = IOUtils.toString(process.getStderr(), encoding);
```

MarkEWaite: After more careful reading of your comments @darxriggs , I believe I've understood your concerns and implemented something that uses your suggestion.

PR #371, opened 2018-10-07
[JENKINS-53933] set file permissions to allow use of openssh on windows

casz:
See [JENKINS-53933](https://issues.jenkins-ci.org/browse/JENKINS-53933)

~Please say if I messed up the import statements~ I did 🤭

Yay for tests! 🙌

Hopefully we can get this in before 4.0 release!
We have had success with this and using OpenSSH from Microsoft 😎

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -46,13 +46,20 @@ import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.AclEntry;
+import java.nio.file.attribute.AclEntryPermission;
+import java.nio.file.attribute.AclEntryType;
+import java.nio.file.attribute.AclFileAttributeView;
 import java.nio.file.attribute.FileAttribute;
 import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.PosixFilePermissions;
+import java.nio.file.attribute.UserPrincipal;
+import java.nio.file.attribute.UserPrincipalLookupService;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -127,6 +134,29 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * ssh configurations).
      */
     private static final boolean CALL_SETSID;
+
+    /**
+     * Needed file permission for OpenSSH client that is made by Windows,
+     * this will remove unwanted users and inherited permissions
+     * which is required when the git client is using the SSH to clone
+     *
+     * The ssh client that the git client ships ignores file permission on Windows
+     * Which the PowerShell team at Microsoft decided to fix in their port of OpenSSH
+     */
+    static final EnumSet<AclEntryPermission> ACL_ENTRY_PERMISSIONS = EnumSet.of(
+        AclEntryPermission.READ_DATA,
+        AclEntryPermission.WRITE_DATA,
+        AclEntryPermission.APPEND_DATA,
+        AclEntryPermission.READ_NAMED_ATTRS,
+        AclEntryPermission.WRITE_NAMED_ATTRS,
+        AclEntryPermission.EXECUTE,
+        AclEntryPermission.READ_ATTRIBUTES,
+        AclEntryPermission.WRITE_ATTRIBUTES,
+        AclEntryPermission.DELETE,
+        AclEntryPermission.READ_ACL,
+        AclEntryPermission.SYNCHRONIZE
+    );
+
     static {
         acceptSelfSignedCertificates = Boolean.getBoolean(GitClient.class.getName() + ".untrustedSSL");
         CALL_SETSID = setsidExists() && USE_SETSID;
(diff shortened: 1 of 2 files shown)


casz:
Sadly tests are failing on CI with "UserPrincipalNotFoundException" I wonder how agents are setup... @rtyler might have the answer 👍

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -5,15 +5,15 @@
 import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
-
+import com.google.common.collect.Lists;
```
MarkEWaite: Please don't reorder the import statements.  It is a pure whitespace change that increases the risk that other pull requests will then conflict with this pull request.

In a perfect world, there would be very few pull requests pending and we could change white space freely.  Unfortunately, I'm far behind reviewing pull requests and need to limit white space changes.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -62,8 +79,8 @@
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.regex.Pattern;
 import java.util.regex.Matcher;
+import java.util.regex.Pattern;
```
MarkEWaite: Please move this import back to the prior position.  Unsorted import statements are not as bad as conflicts being generated in other pull requests due to white space differences

src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java
```diff
@@ -299,6 +307,68 @@ public void test_git_branch_with_line_breaks_and_long_strings() throws Exception
         assertEquals("\"git branch -a -v --no-abbrev\" output correctly parsed", 2, branches.size());
     }
 
+    public void test_windows_file_permission_is_set_correctly() throws Exception {
+        if (!isWindows()) {
```
MarkEWaite: Could you use the assumeThat!(isWindows()) rather than just doing the return?

src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java
```diff
@@ -299,6 +307,68 @@ public void test_git_branch_with_line_breaks_and_long_strings() throws Exception
...
+        assertEquals(1, fileAttributeView.getAcl().size());
+        AclEntry aclEntry = fileAttributeView.getAcl().get(0);
+        assertTrue(aclEntry.flags().isEmpty());
+        assertEquals(CliGitAPIImpl.ACL_ENTRY_PERMISSIONS, aclEntry.permissions());
+        assertEquals(userPrincipal, aclEntry.principal());
+        assertEquals(AclEntryType.ALLOW, aclEntry.type());
+    }
+
+    public void test_windows_file_permission_are_incorrect() throws Exception {
+        if (!isWindows()) {
```
MarkEWaite: Likewise here to use assume rather than just performing an early return?

src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java
```diff
@@ -299,6 +307,68 @@ public void test_git_branch_with_line_breaks_and_long_strings() throws Exception
...
+        assertNotSame(1, fileAttributeView.getAcl().size());
+        for (AclEntry entry : fileAttributeView.getAcl()) {
+            if (entry.principal().equals(userPrincipal)) {
+                assertNotSame(CliGitAPIImpl.ACL_ENTRY_PERMISSIONS, entry.permissions());
+            }
+        }
+    }
+
+    public void test_windows_username_lookup() throws Exception {
+        if (!isWindows()) {
```
MarkEWaite: Here as well for Assume.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
casz: Well the import ordering is all over the place 🤷‍♂️ so I have to do some handheld import handling 🤭

MarkEWaite: "All over the place" is a very kind way of saying "a total mess".  Yes, I agree.  Unfortunately, we need to minimize the impact of past mistakes on pending pull requests, so we keep whitespace changes to a minimum.  Sorry for the extra work (outside the IDE, etc.).

casz:
Atm trying to look for a better way to get the current user principal name

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1869,10 +1899,58 @@ private File createSshKeyFile(SSHUserPrivateKey sshUser) throws IOException, Int
...
+        } catch (IOException | UnsupportedOperationException e) {
+            throw new GitException("Error updating file permission", e);
+        }
+    }
+
+    /* package protected for testability */
+    String getWindowsUserName() {
+        if (launcher.isUnix()) return "";
+
+        String username = System.getenv("USERNAME");
```
casz: Maybe creating a file is the most appropriate way to get user principal instead of relying on env/property and lookup service for user principal?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
casz: Well then at least outside of any ordered list 😅

casz:
@MarkEWaite Wonder if I can do a docker fixture with openssh to test it?

MarkEWaite:
You're right that ci.jenkins.io only supports Linux docker containers.  

If you'd like to try a Windows based Docker container, I'm willing to ask the infrastructure team to experiment with Windows Docker containers.  I assume that will be a long and perilous road before we arrive at Windows Docker containers.

casz:
We already experimented 😅 The docker workflow plugin requires some love before it will work on Windows, we also had to use Windows Server 2019 because we couldn't get older things working. Git, msys2 and ssh would crash the container. Also Microsoft introduced a compatibility matrix https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1869,10 +1899,62 @@ private File createSshKeyFile(SSHUserPrivateKey sshUser) throws IOException, Int
...
+            UserPrincipalLookupService userPrincipalLookupService = file.getFileSystem().getUserPrincipalLookupService();
+            UserPrincipal userPrincipal = userPrincipalLookupService.lookupPrincipalByName(username);
+            AclEntry aclEntry = AclEntry.newBuilder()
+                .setType(AclEntryType.ALLOW)
+                .setPrincipal(userPrincipal)
+                .setPermissions(ACL_ENTRY_PERMISSIONS)
+                .build();
+            fileAttributeView.setAcl(Collections.singletonList(aclEntry));
+        } catch (IOException | UnsupportedOperationException e) {
+            throw new GitException("Error updating file permission", e);
```
MarkEWaite: Could the exception message include the name of the file that failed the permission setting operation?  

That will help us more quickly diagnose the failure in those cases where the caller shows the exception message rather than showing the entire exception stack trace.

casz: will remove the stack trace and print the full file path?

casz: @MarkEWaite done 🙌

casz:
Dno why CI decided to go nuts:

```
[windows-2.121.1] [INFO] Possible null pointer dereference in org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(ArgumentListBuilder, File, EnvVars, Integer) due to return value of called method [org.jenkinsci.plugins.gitclient.CliGitAPIImpl, org.jenkinsci.plugins.gitclient.CliGitAPIImpl] Method invoked at CliGitAPIImpl.java:[line 2255]Known null at CliGitAPIImpl.java:[line 2255] NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE

[windows-2.121.1] [INFO] Possible null pointer dereference in org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(ArgumentListBuilder, File, EnvVars, Integer) due to return value of called method [org.jenkinsci.plugins.gitclient.CliGitAPIImpl, org.jenkinsci.plugins.gitclient.CliGitAPIImpl] Method invoked at CliGitAPIImpl.java:[line 2264]Known null at CliGitAPIImpl.java:[line 2264] NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE
```

Did not even touch those lines 😇

casz:
What are the changes of getting a merge and a 3.0.0-beta6? 🙌

MarkEWaite:
Chances are quite good to include this in git client plugin 3.0.0,  

There is a set of pending pull requests ( #352 from @nre-ableton and a related change from @darxriggs) which I would like to have in the git client plugin before releasing beta6.  I would also like to include #370 in the release so that we base the plugin on Jenkins 2.107.3 instead of Jenkins 2.60.3.

I'd like beta6 to be a release candidate build.  I need to review Jenkins plugins for uses of the JGit Repository object and its getRef() method.  The transition from git client plugin 2.x to git client plugin 3.0.0 switches the bundled JGit version from JGit 4.5.4 to JGit 5.1.1.  That change drops a few API's from JGit and the usage of those dead API's needs to be reviewed in the plugins.

If you'd like to help accelerate the release of git client plugin 3.0.0, you could search the jenkinsci organization for use of the JGit API's that were dropped in the transition from 4.5.4 to 5.1.1.

casz:
Hi @MarkEWaite could this and #406 be patched into v2 git client plugin?

PR #386, opened 2018-12-20
[JENKINS-55284] CLI git tag collision change

MarkEWaite:
Adapt plugin to CLI git 2.20.0 bug fix.

The git fetch behavior prior to git 2.20.0 was to update existing tags without requiring a --force.  The newer versions require --force to update existing tags.

Retain existing behavior for earlier versions, only use the --force option to fetch for new versions of command line git.

Fix [JENKINS-55284](https://issues.jenkins-ci.org/browse/JENKINS-55284) CLI git tag collision change.

@jenkinsci/code-reviewers


MarkEWaite:
@oleg-nenashev I've added an opt-out option using the same technique as the plugin uses for other infrequent opt-out options.  It requires a command line property to opt-out.

PR #391, opened 2019-01-03
Add support for first-parent arg when fetching the list of changes from git

zambamingi:
This is the git-client-plugin side of the changes to implement https://issues.jenkins-ci.org/browse/JENKINS-50366.  There are changes in the git-plugin as well, I'll submitted a pull request there as well.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
@@ -85,14 +85,15 @@ public interface ChangelogCommand extends GitCommand {
     ChangelogCommand includes(ObjectId rev);
 
     /**
-     * Sets the firstParent private variable to true
+     * Limits the changelog to only follow the first parent commit.  If there
+     * are merge commits in the changelog, only follow the first parent.
      */
     ChangelogCommand firstParent();
 
     /**
-     * Sets the firstParent private variable to boolean param
+     * Enable or disable the firstParent option in the calculation of the changelog.
      *
-     * @param firstParent a {@link java.lang.boolean} object.
+     * @param firstParent
      */
     ChangelogCommand firstParent(boolean firstParent);
 
--- a/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
@@ -87,13 +87,15 @@ public interface ChangelogCommand extends GitCommand {
     /**
      * Limits the changelog to only follow the first parent commit.  If there
      * are merge commits in the changelog, only follow the first parent.
+     * @return a {@link org.jenkinsci.plugins.gitclient.ChangelogCommand} object.
      */
     ChangelogCommand firstParent();
 
     /**
      * Enable or disable the firstParent option in the calculation of the changelog.
      *
-     * @param firstParent
+     * @param firstParent a boolean object determining if the changelog is limited to first parent commits or not
+     * @return a {@link org.jenkinsci.plugins.gitclient.ChangelogCommand} object.
      */
     ChangelogCommand firstParent(boolean firstParent);
 
--- a/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
@@ -85,6 +85,18 @@ public interface ChangelogCommand extends GitCommand {
     ChangelogCommand includes(ObjectId rev);
 
     /**
+     * Sets the firstParent private variable to true
+     */
+    ChangelogCommand firstParent();
+
+    /**
+     * Sets the firstParent private variable to boolean param
+     *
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1058,6 +1059,15 @@ public ChangelogCommand includes(ObjectId rev) {
                 return includes(rev.name());
             }
 
+            public ChangelogCommand firstParent() {
```
darxriggs: Please add `@Override`.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1058,6 +1059,15 @@ public ChangelogCommand includes(ObjectId rev) {
...
+                return firstParent(true);
+            }
+
+            public ChangelogCommand firstParent(boolean firstParent) {
```
darxriggs: Please add `@Override`.

src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
```diff
@@ -84,6 +84,18 @@
      */
     ChangelogCommand includes(ObjectId rev);
 
+    /**
+     * Sets the firstParent private variable to true
```
darxriggs: I suggest to document the behaviour that is triggered by setting the variable value, not just that the variable is set. The former is more useful for the user of this API.

src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
```diff
@@ -84,6 +84,18 @@
...
+     */
+    ChangelogCommand firstParent();
+
+    /**
+     * Sets the firstParent private variable to boolean param
```
darxriggs: See the comment for `firstParent()`.

src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
```diff
@@ -84,6 +84,18 @@
...
+     *
+     * @param firstParent a {@link java.lang.boolean} object.
```
darxriggs: Please don't include the type in description again (even if there are some existing Javadocs like this} - `java.lang.boolean` is not even valid.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -1112,6 +1112,14 @@ public ChangelogCommand includes(ObjectId rev) {
                 }
             }
 
+            public ChangelogCommand firstParent() {
+                return firstParent(true);
+            }
+
+            public ChangelogCommand firstParent(boolean firstParent) {
+                return this;
```
darxriggs: Please either implement this functionality in JGit (preferred) or log a warning that this parameter is not supported in the JGit implementation. Otherwise the user has no idea when enabling this feature why it's not working as expected.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -4724,6 +4724,34 @@ public void test_changelog_with_merge_commit_and_max_log_history() throws Except
...
+        w.git.add("file-1");
+        w.git.commit("commit-1");
+        String commitSha1 = w.git.revParse("HEAD").name();
+
+        // Merge branch-1 into master
+        w.git.checkout("master");
+        String mergeMessage = "Merge message to be tested.";
+        w.git.merge().setMessage(mergeMessage).setGitPluginFastForwardMode(MergeCommand.GitPluginFastForwardMode.NO_FF).setRevisionToMerge(w.git.getHeadRev(w.repoPath(), "branch-1")).execute();
+
+        if (w.git instanceof CliGitAPIImpl) {
```
darxriggs: I suggest to use `@NotImplementedInJGit` on the test method instead of this `instanceof` check. The former is more descriptive.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
zambamingi: thanks for catching this, I addressed this in my latest commit

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
zambamingi: thanks for the suggestion, I addressed this in my latest commit

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
zambamingi: thanks for the suggestion, I addressed this in my latest commit

MarkEWaite:
Assigned label 'Changelog' and milestone 'Later' so that it can be revisited when reviewing changelog related pull requests

PR #393, opened 2019-01-14
WIP: test JTH 2.45 (with space in tmp dir path)

thomasgl-orange:
Following jenkinsci/plugin-pom#93, here is a test PR to see how jenkinsci/git-client-plugin unit tests behave when there is a space character in their temporary workdir path.

The first commit (9283c16) bumps `plugin-pom` to a SNAPSHOT version for testing, and the second commit (3cbdbe3) fixes numerous simple tests failures by replacing some `WorkingArea.cmd("git some command " + r.repoPath());` with `WorkingArea.launchCommand("git", "some", "command", r.repoPath());` (the former syntax assumes it is safe to split the command line by spaces, which is no more true).

After this second commit, at least on Linux and with a recent `git` version, the only remaining tests failures are related to submodules handling in the `CliGitAPIImpl`.  It seems to all boil down to the fix for [JENKINS-46504](https://issues.jenkins-ci.org/browse/JENKINS-46504) being based on the assumption that submodules URLs are properly escaped (as in RFC1738, with no space character - see comments for `CliGitAPIImpl.SUBMODULE_REMOTE_PATTERN_STRING`).  I don't think it is safe to assume so. Of course, I could modify test cases to make it true there and make the tests pass, but in reality, `git` seems to be perfectly happy with non-escaped URLs, and `git submodule add` does nothing by itself to enforce URL encoding, so I can imagine URLs with spaces are a pretty common case in the wild.
I'm not sure what the proper fix would be:
 - using `git config -f .gitmodules --get-regexp ... --name-only` would be the obvious solution for listing submodules, but this `--name-only` option has only been introduced in `git 2.6`.
 - playing with the parsing regexp themselves, I've been able to somehow improve the situation (I think), but there are still some fundamental ambiguities in the `git config --get-regexp` (as soon as you introduce the `".url "` substring, with a space, in either the submodule path or in its remote URL)
 - getting both `submodule.*.url` and `submodule.*.path`, and looking for pairs of keys, might be a solution to get an unambiguous list of submodules, but it sounds tedious.
 - maybe something with the `--null` option? (I've not tried it yet, but it looks like this option has been around longer than `--name-only`)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.32</version>
+    <version>3.33-20190110.085245-1</version>
     <relativePath />
   </parent>
 
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -11,6 +11,8 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
+
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Launcher.LocalLauncher;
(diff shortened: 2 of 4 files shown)


thomasgl-orange:
It looks like the test failures on Windows are the same that those I was expecting on Linux (that's the same 4 failed tests I've had here):
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgit-client-plugin/detail/PR-393/1/tests

thomasgl-orange:
Added 2 more commits (08663b3 and b0164da), with a fix for the previously discussed submodules listing issue. All unit tests are OK now:
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgit-client-plugin/detail/PR-393/2/tests

With `git 1.5.3+`, it uses the `--null` option to `git config`, which eases parsing the output.

With older `git` versions, it parses the standard `git config` output like before, but without assuming that there can be no space in the URL part.  Instead, it splits the config key / config value line on the right-most `".url "` substring.  So parsing will be correct with most remote URL containing spaces, but with those containing the `".url "` substring.  Which is, I think, still an improvement compared to the current implementation.

Note that I'm not happy with the code in b0164da though, I've introduced static methods in `CliGitAPIImpl` just to ease porting the `SubmodulePatternStringTest` unit test case, but it's not something I would really recommend merging as-is.

MarkEWaite:
Closing but marked with label 'Submodules' and milestone 'Later' for consideration in the future

PR #399, opened 2019-01-19
Reduce deprecated API usage

darxriggs:
This reduces the amount of deprecated API calls. It doesn't resolve all deprecated warnings.

### DIFF
--- a/src/main/java/hudson/plugins/git/GitAPI.java
+++ b/src/main/java/hudson/plugins/git/GitAPI.java
@@ -167,14 +167,14 @@ public class GitAPI extends CliGitAPIImpl {
     */
 
     /*
-    public void submoduleUpdate(boolean recursive) throws GitException {
-        if (Git.USE_CLI) super.submoduleUpdate(recursive); else  jgit.submoduleUpdate(recursive);
+    public void submoduleUpdate(boolean recursive) throws GitException, InterruptedException {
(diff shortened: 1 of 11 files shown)


darxriggs:
It looks like there is a problem with the Jenkins infrastructure.

MarkEWaite:
Sorry about the infrastructure problem, @darxriggs .

I've submitted a placeholder pull request https://github.com/MarkEWaite/git-client-plugin/pull/54 to let my CI infrastructure evaluate the pull request.  The server is behind my firewall, so you won't be able to see the details, but it should publish summary results on that placeholder pull request

MarkEWaite:
The Linux-based tests pass in my environment.  The Windows tests are still running.  

One test has complained in the usual way about parsing the results of the long-running credentials tests that I run in my environment.  I believe that is a result of the `buildPlugin()` PIpeline shared library that attempts to read tests results while the job is running.  The rest results file for the CredentialsTest in my environment remains open and receiving results for 60-120 seconds.  I assume that is why it often reports that the test results file can't be parsed in my environment and never reports it on ci.jenkins.io.

darxriggs:
Can it be merged then?

MarkEWaite:
I still need to review the code.  I need to confirm that there are not significant differences between the deprecated calls and the replacement calls.  Replacing deprecated calls can be a nice cleanup, but if it breaks use cases, it will be much worse than leaving the deprecated calls as they are.

darxriggs:
As quite some time passed and v3.0.0 is released now, I rebased the changes on the master branch.

PR #400, opened 2019-01-24
JENKINS-43427 Add Git LFS selective fetching support (sparse checkout)

darxriggs:
Automatically configure git-lfs selective fetching when sparse checkout is configured.

Negate patterns (leading "!") have to be transformed to `lfs.fetchexclude` settings.

Commas "," are not supported in `lfs.fetchinclude` and `lfs.fetchexclude`. See https://github.com/git-lfs/git-lfs/issues/2264. Therefore git-lfs selective fetching is not used if commas are detected in the sparse checkout configuration.

Resources
* https://git-scm.com/docs/gitignore/2.20.1
* https://git-scm.com/docs/git-config/2.20.1
* https://git-scm.com/docs/git-read-tree/2.20.1
* https://github.com/git-lfs/git-lfs/blob/v2.6.1/docs/man/git-lfs-fetch.1.ronn
* https://godoc.org/github.com/git-lfs/wildmatch

This supersedes pull request #302 and fixes all found issues. Thanks @tiffloon for the initial effort.


darxriggs:
Hi @MarkEWaite, could you have a look at this?

I have one open question: Should selective LFS fetching always be applied automaticly when sparse checkout is configured or should it have to be enabled/disabled explicitly via an extra option in `GitLFSPull`?

ilanbiala:
@darxriggs @MarkEWaite any update on this?

MarkEWaite:
@darxriggs I think it should be enabled if sparse checkout is enabled.  No need for a separate configuration setting.

I expect very few users will demand to have sparse checkout with LFS that does **not** also honor the sparse checkout with LFS.  This is one of those rare cases where I think we can risk compatibility because the previous behavior is inconsistent and the number of users expected to depend on the old inconsistent behavior is quite small.

ilanbiala:
@MarkEWaite I have not. I was checking on behalf of an infrastructure engineer, so I'm not certain to what extent this has been used/tested.

darxriggs:
As quite some time passed and v3.0.0 is released now, I rebased the changes on the master branch.

darxriggs:
@MarkEWaite Do you have a rough estimation when you will find the time to review this?

MarkEWaite:
I've added it to the `ShortTerm` list.  My current goal is to complete the review of the credentials during merge and submodule clean then deliver git client plugin 3.1 and git plugin 4.1 with submodule clean, credentials during merge, and JGit 5.6.0 before end of the calendar year.

darxriggs:
For me this one doesn't have a high priority. It's just that it's already open and ready for a very long time from my side.

PR #403, opened 2019-01-29
Add `rm` command to the client

drkunibar:
I needed the `rm` command. I have implemented this equivalently to the `add` command

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/GitClient.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/GitClient.java
@@ -158,6 +158,15 @@ public interface GitClient {
     void add(String filePattern) throws GitException, InterruptedException;
 
     /**
+     * rm.
+     *
+     * @param filePattern a {@link java.lang.String} object.
+     * @throws hudson.plugins.git.GitException if underlying git operation fails.
+     * @throws java.lang.InterruptedException if interrupted.
+     */
+    void remove(String filePattern) throws GitException, InterruptedException;
+
+    /**
      * commit.
      *
      * @param message a {@link java.lang.String} object.
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
@@ -754,6 +754,21 @@ public abstract class GitAPITestCase extends TestCase {
         assertFalse(w.git.isCommitInRepo(null)); // NPE safety check
     }
 
+    public void test_remove_file() throws Exception {
+        w.init();
+        assertFalse(w.git.isCommitInRepo(null)); // NPE safety check
+        w.touch("file1");
+        w.git.add("file1");
+        w.git.commit("commit1");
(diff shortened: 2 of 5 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/GitClient.java
```diff
@@ -157,6 +157,15 @@
      */
     void add(String filePattern) throws GitException, InterruptedException;
 
+    /**
+     * rm.
```
MarkEWaite: Could you add a sentence to describe the `git rm` command that is being invoked ?

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -754,6 +754,21 @@ public void test_detect_commit_in_repo() throws Exception {
...
+        w.git.commit("commit1");
+        w.git.remove("file1");
+        w.git.commit("commit2");
+        assertFalse("file1 should not exists", w.exists("file1"));
+        assertTrue("HEAD commit not found", w.git.isCommitInRepo(w.head()));
+        // this MAY fail if commit has this exact sha1, but please admit this would be unlucky
+        assertFalse(w.git.isCommitInRepo(ObjectId.fromString("1111111111111111111111111111111111111111")));
+        assertFalse(w.git.isCommitInRepo(null)); // NPE safety check
+    }
+
```
MarkEWaite: Could you add a test to confirm consistent behavior when calling:
* `remove("added-file-not-yet-committed")` - a file that has been added and not yet committed
* `remove("existing-file-not-yet-added")` - a file that exists and has not been added
* `remove("non-existing-file")` - a file that does not exist
* `remove("directory")` - should the remove be recursive by default or should it take an optional recursive argument?
* `remove(".git")` - a directory that is not allowed to be removed?
* `remove("pattern-matching-multiple-files*")` - a file pattern that matches multiple files

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestCase.java
```diff
@@ -754,6 +754,21 @@ public void test_detect_commit_in_repo() throws Exception {
...
+        assertFalse(w.git.isCommitInRepo(null)); // NPE safety check
+        w.touch("file1");
+        w.git.add("file1");
+        w.git.commit("commit1");
+        w.git.remove("file1");
+        w.git.commit("commit2");
+        assertFalse("file1 should not exists", w.exists("file1"));
+        assertTrue("HEAD commit not found", w.git.isCommitInRepo(w.head()));
+        // this MAY fail if commit has this exact sha1, but please admit this would be unlucky
+        assertFalse(w.git.isCommitInRepo(ObjectId.fromString("1111111111111111111111111111111111111111")));
```
MarkEWaite: Could you remove these two lines?  They are unrelated to remove and are already tested elsewhere in the file.

MarkEWaite:
Closing for now, assigned to the 'Later' milestone in case it should be reopened in the future

PR #404, opened 2019-01-30
[JENKINS-55875] Forcibly update git submodules

richardwhiuk:
This change resolves https://issues.jenkins-ci.org/browse/JENKINS-55875

Currently, submodules are not updated with `--force`. This means if there is a local change, then the checkout fails, causing the build to fail.

This can be worked round by cleaning the repository before every checkout, but this makes build jobs much slower, as they aren't able to be incrementally built, as the entire workspace has to be thrown away each time, or requires manual intervention.

Instead, this changes the checkout to be done forcefully, which matches how the checkout of the top level Git repository works.

This change should not have any harmful effects - it will only cause builds which were failing to now succeed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1234,7 +1234,9 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 submoduleInit();
 
                 ArgumentListBuilder args = new ArgumentListBuilder();
-                args.add("submodule", "update");
+
+                // Force the update, in the same manner we force checkouts
+                args.add("submodule", "update", "--force");
                 if (recursive) {
                     args.add("--init", "--recursive");
                 }


MarkEWaite:
Needs a test which shows the problem and that confirms the problem is fixed by the code change.

darxriggs:
@richardwhiuk can you provide a test that covers the new functionality. Maybe it's also sufficient to update an existing one (there is currently `test_submodule_update_with_error` that fails after your change).

darxriggs:
@richardwhiuk do you already know when you will work on this again?

darxriggs:
@richardwhiuk friendly reminder

NuclearCookie:
Hey @darxriggs I'll update the test as requested.
I've run into the same issue and would like to see this merged! 

I'll have to open a new PR because I don't have push rights on the OP's repo

NuclearCookie:
Done in #505

MarkEWaite:
Closing, work on the change will continue in #505

michaelakin:
Please merge this.  My company could use this right now.

PR #405, opened 2019-01-30
[JENKINS-55876] Recursively update submodules

richardwhiuk:
Currently `git submodule sync` is not run recursively, even when the Git Plugin is configured to be recursive for the SCM. This is because the `git submodule sync` command does not take into account the recursive flag.

This change exposes an interface from the git-client-plugin which allows the git-plugin to recursively synchronize Git URLs.


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1142,13 +1142,21 @@ public void addSubmodule(String remoteURL, String subdir) throws GitException, I
...
      * @throws java.lang.InterruptedException if interrupted.
      */
-    public void submoduleSync() throws GitException, InterruptedException {
+    public void submoduleSync(boolean recursive) throws GitException, InterruptedException {
         // Check if git submodule has sync support.
         // Only available in git 1.6.1 and above
-        launchCommand("submodule", "sync");
+        if (recursive)
+        {
+            launchCommand("submodule", "sync", "--recursive");
```
darxriggs: Could you please check since which version cli git supports this submodule sync option? Then we may have to discuss if this has to handled here explicitly (ignoring this `--recursive` and log an information about it).

MarkEWaite: Command line git 1.8.3 includes `git submodule sync --recursive`.  The oldest version of git officially supported by the git client plugin is 1.7.10.  All the command line git versions I've tested prior to 1.9 (1.7.1 in CentOS, 1.7.10 on Windows, 1.8.3 on Debian 7, etc.) are known to have various issues when working with the plugin. 

Many of those issues are described in the tests with `isAtLeastVersion` conditionals.

I would be interested to know if git as included in CentOS 6 (git 1.7.1) includes `git submodule sync --recursive`.  If it does, then I'd recommend we not bother with a conditional.  If it doesn't, then we should probably place a conditional around the recursive call in order to avoid breaking CentOS 6 users.  While their command line git version is not officially supported, it continues to work in many useful cases.

MarkEWaite:
See #205 for a similar request

darxriggs:
@richardwhiuk could you have a look at the review comments and also resolve the merge conflict?

MarkEWaite:
I reserved the merge conflict.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
MarkEWaite: Git version 1.7.1 (on CentOS 6) supports `git submodule sync` .  All command line git versions supported by the git plugin support it.

Pymann:
I have similar issue, imagine there are 2 branches: on both branches we have same submodule and this submodule has a submodule but the url for both is different on those branches. Anything, that can be done?!

MarkEWaite:
Closing this pull request.  I'm not allowed to push changes to it.  I've saved the branch in my repository and can open a new pull request to replace it.

PR #406, opened 2019-02-08
use current owner to avoid user lookup

casz:
We saw some issues when jenkins agents where running in windows docker containers and docker host was connected to an active directory

This works since the owner is correct because the process owner is the one who created the file.


casz:
Feels good to delete code 😅

casz:
@MarkEWaite hi would be good to get this merged before 3.0 release 👍

casz:
Just cross-referencing #371

PR #415, opened 2019-03-11
JENKINS-55948 - git parameter plugin calls fail with git client 3.x

MarkEWaite:
Fix JENKINS-55948 - git parameter plugin calls fail with git client 3.x

Need to identify the cause of the failure and fix it.


MarkEWaite:
This is not a regression.  The same problem has been reported by users running earlier git plugin versions and earlier git client plugin versions.  It does not need to block the release of git client plugin 3.0.0

PR #416, opened 2019-03-15
[JENKINS-56257] Allow embedded username/password in URL

ckrams:
Fix - prevent truncation of password in git url for git clone

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -453,7 +453,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 StandardCredentials cred = credentials.get(url.toPrivateString());
                 if (cred == null) cred = defaultCredentials;
 // JENKINS-56257 
-// prevent truncation of passward in the git url				
+// prevent truncation of passward in the git url for clone				
 				args.add(url.toPrivateString());
 // JENKINS-56257
                 if (refspecs != null)
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -452,8 +452,10 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
                 StandardCredentials cred = credentials.get(url.toPrivateString());
                 if (cred == null) cred = defaultCredentials;
-                args.add(url);
-
+// JENKINS-56257 
+// prevent truncation of passward in the git url				
+				args.add(url.toPrivateString());
+// JENKINS-56257
                 if (refspecs != null)
                     for (RefSpec rs: refspecs)
                         if (rs != null)
@@ -514,7 +516,9 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         String url = getRemoteUrl(remoteName);
         if (url == null)
             throw new GitException("remote." + remoteName + ".url not defined");
+
         args.add(url);
+        
         if (refspec != null && refspec.length > 0)
             for (RefSpec rs: refspec)
                 if (rs != null)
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -514,7 +516,9 @@ public void fetch(String remoteName, RefSpec... refspec) throws GitException, In
         String url = getRemoteUrl(remoteName);
         if (url == null)
             throw new GitException("remote." + remoteName + ".url not defined");
+
```
MarkEWaite: Please remove this whitespace change from the pull request.  There are many pull requests which need to be evaluated for the git client plugin.  White space changes tend to cause unnecessary merge conflicts with other pull requests.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -514,7 +516,9 @@ public void fetch(String remoteName, RefSpec... refspec) throws GitException, In
...
         args.add(url);
+
```
MarkEWaite: Please remove this whitespace change from the pull request.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -452,8 +452,10 @@ public void execute() throws GitException, InterruptedException {
 
                 StandardCredentials cred = credentials.get(url.toPrivateString());
                 if (cred == null) cred = defaultCredentials;
-                args.add(url);
-
+// JENKINS-56257 
+// prevent truncation of passward in the git url for clone				
+				args.add(url.toPrivateString());
```
MarkEWaite: Please extend the CredentialsTest in this repository so that it can be used to test this case.  The CredentialsTest uses a json file that is local to your file system.  It reads plain text credentials from that json file and uses them to test authenticated git operations.

The CredentialsTest class is currently structured to require either a private key or a username / password.  You could extend that test to also accept an optional field like `credentialsEmbeddedInURL`.  When that field has the value `true`, then the URL should be passed through the test without applying a credential.

That testing technique allows the developer (you) and the maintainer (me) to evaluate credentials without sharing our credentials in the repository.

venkateshilango7:
Hi @MarkEWaite ,

We have added the test case in CredentialsTest.
Can you please review changes and merge PR.

Thanks and Regards, Venkatesh

MarkEWaite:
@venkateshilango7 I had intended that you would modify the existing test(s) in CredentialsTest rather than create a new test in CredentialsTest.  By creating a new test, that will cause the new test to be executed for each of the 50+ credential data samples that I have.  That will likely increase the test execution time significantly and decrease the number of credential tests that can be run in the limited time that JenkinsRule allows a test to run.

I want the existing tests to be extended to adapt to the credentials being embedded in the URL rather than adding a new test.  Are you willing to revisit the test you wrote and rather attempt to modify the existing tests so that they honor the new, optional `credentialsEmbeddedInURL` data in the repos.json file?

venkateshilango7:
Hi Mark,

Yes I will revisit the test cases and update the PR.

Thanks and Regards,Venkatesh

venkateshilango7:
Hi @MarkEWaite ,

Can you please review PR changes.

Thanks and Regards, Venkatesh

MarkEWaite:
Thanks @venkateshilango7 .  Sorry that it took two weeks before I could return to the review of this pull request.

Thanks for extending CredentialsTest to accept a credential embedded in the URL.  That allowed me to run the automated tests and extend my private data set to test this case.

I've submitted several proposed changes in [my feature branch](https://github.com/MarkEWaite/git-client-plugin/tree/allow-embedded-credentials-in-url-JENKINS-56257).  After making the changes that I've proposed, the automated tests show that the JGitAPIImpl support for credentials embedded in the URL has not been done.  That will need to be done before this can be merged.

I also realized that there may be a better way to handle the embedded credentials in CredentialsTest.  I think we can avoid changing the data format of the repos.json file.  If the data generator function detects a username/password, then it can modify the URL to include the username and password and return that as another parameterized case.  In that situation, a single entry in repos.json with username/password would result in two tests, one with the username/password embedded in the URL and the other with the username and password passed as parameters.

If time allows today, I'll provide a first attempt at that technique, just to see if it could work.

MarkEWaite:
Assigning to the `Later` milestone and labeled as an enhancement and a work in progress so that it can be reconsidered later.  Closing it for now

MarkEWaite:
Change included in git client plugin 2.8.5 as part of fix for SECURITY-1534.

PR #421, opened 2019-04-14
[JENKINS-55655] Use CLI git in Blue Ocean

MarkEWaite:
## [JENKINS-55655](https://issues.jenkins-ci.org/browse/JENKINS-55655) - Use CLI git in Blue Ocean

JENKINS-55655 reports that exe == null did not honor USE_CLI value.  Corrects code to match javadoc and to match most user expectations that command line git is the default implementation.

This change will require command line git be installed on a master that runs the Blue Ocean user interface when Blue Ocean performs git operations.  Previously, Blue Ocean would accidentally use JGit and did not require a command line git installation on the master.  

This will also allow Blue Ocean to access git repositories through a proxy as described in JENKINS-55655.

Blue Ocean users can return to the previous behavior with the property:

```
org.jenkinsci.plugins.gitclient.Git.USE_CLI=false
```

## Checklist

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Further comments

Intentionally chose to honor the setting of USE_CLI in fixing this bug rather than retaining bug compatibility with the previous implementation.
Assumes that almost all Jenkins installations that use git have command line git installed on the master.

Those Jenkins installations that do not have command line git installed on the master and cannot install command line git on the master will need to make that decision explicit by setting the property:

```
org.jenkinsci.plugins.gitclient.Git.USE_CLI=false
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/Git.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/Git.java
@@ -5,6 +5,7 @@ import hudson.FilePath;
 import hudson.Main;
 import hudson.model.TaskListener;
 import hudson.plugins.git.GitAPI;
+import hudson.plugins.git.GitTool;
 import hudson.remoting.VirtualChannel;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.gitclient.jgit.PreemptiveAuthHttpClientConnectionFactory;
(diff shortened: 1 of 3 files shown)


MarkEWaite:
This is more complicated than I thought.  I'm closing it for now.  I may return to it in the future, after I've understood the additional tests that need to be written.

PR #426, opened 2019-05-17
[JENKINS-57326] make GitTool abstract to follow JCasC conventions

ndeloof:
## [JENKINS-57326](https://issues.jenkins-ci.org/browse/JENKINS-57326) - support Configuration as Code plugin

GitTool is made abstract, so that JCasC will correctly detect a plugable extension and can also support JGit|JGitApache
Please note this changes the yaml structure used today to configure git tools installations.

## Checklist

- [ ] Any dependent changes have been merged and published in upstream modules (like git-client-plugin)

## Types of changes

- [x] Bug fix
- [x] Breaking change

## Further comments

Needs deep testing for upgrade compatibility, run time compatibility, and compile time compatibility with git client plugin consumers.

### DIFF
--- a/src/main/java/hudson/plugins/git/GitTool.java
+++ b/src/main/java/hudson/plugins/git/GitTool.java
@@ -1,31 +1,17 @@
 package hudson.plugins.git;
 
-import hudson.EnvVars;
-import hudson.Extension;
-import hudson.init.Initializer;
 import hudson.model.EnvironmentSpecific;
-import hudson.model.Node;
-import hudson.model.TaskListener;
+import hudson.model.Items;
 import hudson.slaves.NodeSpecific;
 import hudson.tools.ToolDescriptor;
 import hudson.tools.ToolInstallation;
 import hudson.tools.ToolProperty;
-import hudson.util.FormValidation;
 import jenkins.model.Jenkins;
-import net.sf.json.JSONObject;
-import org.jenkinsci.Symbol;
+import org.jenkinsci.plugins.gitclient.CLIGitTool;
 import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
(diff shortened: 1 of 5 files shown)


MarkEWaite:
@ndeloof thanks for the pull request.  Considering the number of failing tests and how completely inadequate I feel in this area of Jenkins, would it be worth considering other alternatives?

For example, the GitTool class is annotated with the symbol `git` and that symbol seems to be used by JCasC to determine the `git` subsection of the `tools` section.  Could we adjust JCasC to honor the `jgit` symbol that is annotated on the JGitTool class, thus creating a new subsection in JCasC at the same level as the git section?  

If that would work, it would allow us to retain the current syntax for the `git` subsection of JCasC by adding two new subsections, one called `jgit` and the other called `jgitapache`.

I don't know what would be inside those subsections, since there really isn't anything to configure for jgit or jgitapache.  Maybe a key `enabled` with allowed values `true` and the default value `false`?

ndeloof:
@MarkEWaite sorry I forgot to mark this PR as work in progress. As tests hang on my laptop (even on master) I wanted to get them ran in a neutral environment with this PR. This is work in progress.

To answer your note, the issue here is not for JCasC to _not_ honor the symbols, but to know that `GitTool` when used as a field in another component (like `installations` here) can have derived implementation. In most Jenkins components we have a higher level abstract class to define the "extension point" and this one is used as type. When a concrete class is used as a reference, it _use_ to mean we just have a direct reference, without intent to let user plug alternative implementation. GitTool is pretty uncommon from this point of view.

MarkEWaite:
I don't think that this is likely to be a successful path to add configuration as code support for the jgit and jgit-apache implementations in the git client plugin.  I'm closing it for now.  If further interest is shown in this pull request, we can reopen it in the future.

PR #440, opened 2019-07-29
GitExceptionTest failed when run as root

jglick:
Discovered as part of https://github.com/jenkinsci/bom/pull/36, which runs this plugin’s tests under PCT in a Docker container running as root. (Would also see this if switching `linux` to `maven` in the `buildPlugin` config.)

Verified with

```sh
docker run -v ~/.m2:/var/maven/.m2 -ti --rm --name mvn -e MAVEN_CONFIG=/var/maven/.m2 -v "$PWD":/usr/src/mymaven -w /usr/src/mymaven maven:3.6.1-jdk-8 mvn -Duser.home=/var/maven surefire:test -Dtest=GitExceptionTest
```

### DIFF
--- a/src/test/java/hudson/plugins/git/GitExceptionTest.java
+++ b/src/test/java/hudson/plugins/git/GitExceptionTest.java
@@ -18,6 +18,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import static org.junit.Assert.*;
+import static org.junit.Assume.*;
 import org.junit.rules.TemporaryFolder;
 import static org.hamcrest.CoreMatchers.is;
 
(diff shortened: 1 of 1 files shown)


src/test/java/hudson/plugins/git/GitExceptionTest.java
```diff
@@ -54,24 +55,13 @@ public void throwsGitExceptionExpectedMessageWithCause() {
         throw new GitException(message, new IOException("Custom IOException message"));
     }
 
-    @Test
-    public void initDefaultImplThrowsGitException() throws GitAPIException, IOException, InterruptedException {
-        File badDirectory = new File("/this/is/a/bad/dir");
-        if (isWindows()) {
-            badDirectory = new File("\\\\badserver\\badshare\\bad\\dir");
-        }
-        GitClient defaultClient = Git.with(TaskListener.NULL, new EnvVars()).in(badDirectory).using("Default").getClient();
```
jglick: This test was nonsense—it _did_ throw an exception (running as `root` or not), but only because it was trying to running a command named `Default`, which of course failed.

jglick:
By the way I am also having problems with `FilePermissionsTest`, though I cannot track down why.

* trunk, local: fails with `JGitInternalException: missing credentials provider` and `Could not checkout master with start point origin/master`
* trunk, Docker: passes
* PCT, local: passes
* PCT, Docker: fails with `git-740.txt execute perm mismatch expected:<true> but was:<false>`

MarkEWaite:
> By the way I am also having problems with `FilePermissionsTest`, though I cannot track down why.
> 
> * trunk, local: fails with `JGitInternalException: missing credentials provider` and `Could not checkout master with start point origin/master`
> * trunk, Docker: passes
> * PCT, local: passes
> * PCT, Docker: fails with `git-740.txt execute perm mismatch expected:<true> but was:<false>`

Since file permissions may be modified by the `umask` setting of the runtime environment on Unix variants, it may be that there is a difference in the default file permissions used when creating the file initially.

If the test is too much of a distraction, it can be deleted.  It taught me what I needed to learn about writing a test that would run with platform specific code on multiple platforms, but has not detected a problem in a very long time.

jglick:
> Since file permissions may be modified by the `umask` setting of the runtime environment on Unix variants, it may be that there is a difference in the default file permissions used when creating the file initially.

In fact my local umask is 0002 whereas that inside the container is 0022. The trunk test still fails locally (in the same way) with a umask of 0022, though, and the PCT test still fails in Docker (in the same way) with a umask of 0002. So this does not seem to be the source of variance.

For the moment I am just suppressing the failure, since I do not understand the test well enough to judge whether or not the failure is indicative of a real problem.

PR #450, opened 2019-08-23
Fix test failure when running as root user

MarkEWaite:
## Fix test failure when running as root user

Test assumes it is running as an unprivileged user and relies on that for a failure case.  Fix that and release a new version so that the BOM can depend on a released version which includes passing tests.

## Checklist

## Types of changes

### DIFF
--- a/src/test/java/hudson/plugins/git/GitExceptionTest.java
+++ b/src/test/java/hudson/plugins/git/GitExceptionTest.java
@@ -20,6 +20,7 @@ import org.junit.rules.ExpectedException;
 import static org.junit.Assert.*;
 import org.junit.rules.TemporaryFolder;
 import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assume.*;
 
 public class GitExceptionTest {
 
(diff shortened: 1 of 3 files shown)


jglick:
Discovered via https://github.com/jenkinsci/bom/pull/71, FTR.

jglick:
Fix would likely look something like #440: just check whether the current user is actually barred from writing to the nonexistent folder.

jglick:
The downside of running with `useAci: true` is of course is that these tests would start to be skipped on CI, since this image runs as `root`.

Now we _could_ patch that image to run as a regular user, but that could be a breaking change. In fact sometimes it is handy to have the build run as `root`, so that you can install system packages to get tools not currently included in the image. (At least on an emergency basis—in the long term you should be adding tools to the image or creating new images.)

MarkEWaite:
I don't want to run the tests as root.  I think that running tests as root will hide more problems than it will discover.

However, I think that the tests need to run successfully in many different environments, including those environments (like a docker image) where the test may be running as root.

Once the most recent changes have confirmed that the tests are now "root safe", I will rebase this pull request, remove the `useAci: true` commit (https://github.com/jenkinsci/git-client-plugin/pull/450/commits/d298ad97381949ada951629b033887a270e452de) and force push.

PR #459, opened 2019-09-15
Run tests in parallel to reduce overall execution time

darxriggs:
The timeout setting in `CredentialsTest` has to be removed because due to parallel running tests individual methods may run slower.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
     <jenkins.version>2.121.1</jenkins.version>
     <java.level>8</java.level>
     <jgit.version>5.5.0.201909110433-r</jgit.version>
+    <forkCount>1C</forkCount>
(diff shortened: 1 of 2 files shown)


MarkEWaite:
I'm surprised at removing the credentials test timeout.  The credentials test runs at most a few tests on systems with the default configuration.  It requires a special file (~/.ssh/auth-data/repos.json) before it runs more than a minimal set of tests.

Did you see failures from that timeout or is the change more a safety measure?

The 17 second timeout is much longer than most credentials tests require, except in the failure case.  When they fail, that timeout allows other tests to run.

darxriggs:
When enabling parallel tests `testFetchWithCredentials` always runs into the timeout.

PR #465, opened 2019-09-27
BOM 3

jglick:
https://github.com/jenkinsci/bom/releases/tag/bom-3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,8 +260,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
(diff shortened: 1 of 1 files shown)


jglick:
Note that I am relying entirely on CI to tell whether PRs like these are OK! And the CI server broke yesterday. So YMMV

MarkEWaite:
Thanks!  I'm also relying on CI, though in my case it is both CI that I'm hosting myself.

This passed all tests on my CI setup yesterday (multiple times) and is now deployed into my CI setup as the running version of the git client plugin.  Same for the git plugin.

PR #468, opened 2019-10-02
[JENKINS-57326] JCasC support for GitTool and subtypes

rsandell:
## [JENKINS-57326](https://issues.jenkins-ci.org/browse/JENKINS-57326) - JCasC support for Git tool installers.

Different approach to #426 using a custom Configurator instead of trying to redesign the entire class structure.

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.md) doc
- [X] I have referenced the Jira issue related to my changes in one or more commit messages
- [X] I have added tests that verify my changes
- [X] Unit tests pass locally with my changes
- [X] I have added documentation as necessary
- [X] Any dependent changes have been merged and published in upstream modules (like git-client-plugin)

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
     <java.level>8</java.level>
     <jgit.version>4.5.7.201904151645-r</jgit.version>
     <jgit.javadoc.version>4.5.0.201609210915-r</jgit.javadoc.version>
-    <jcasc.version>1.29</jcasc.version>
+    <jcasc.version>1.30</jcasc.version>
   </properties>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
     <java.level>8</java.level>
     <jgit.version>4.5.7.201904151645-r</jgit.version>
     <jgit.javadoc.version>4.5.0.201609210915-r</jgit.javadoc.version>
+    <jcasc.version>1.29</jcasc.version>
   </properties>
 
   <repositories>
@@ -174,6 +175,20 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials</artifactId>
+    </dependency>
+    <!-- JCasC compatibility -->
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
+      <version>${jcasc.version}</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
+      <version>${jcasc.version}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
@@ -14,7 +14,6 @@ import io.jenkins.plugins.casc.model.CNode;
 import io.jenkins.plugins.casc.model.Mapping;
 import io.jenkins.plugins.casc.model.Sequence;
 
-import javax.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -70,7 +69,10 @@ public class GitToolConfigurator extends BaseConfigurator<GitTool> {
     }
 
     @Override
-    protected GitTool instance(Mapping mapping, ConfigurationContext context) throws ConfiguratorException {
+    protected GitTool instance(Mapping mapping, @NonNull ConfigurationContext context) throws ConfiguratorException {
+        if (mapping == null) {
+            return new GitTool("Default", "", instantiateProperties(null, context));
+        }
         final CNode mproperties = mapping.remove("properties");
         final String name = mapping.getScalarValue("name");
         if (JGitTool.MAGIC_EXENAME.equals(name)) {
@@ -92,13 +94,13 @@ public class GitToolConfigurator extends BaseConfigurator<GitTool> {
         }
     }
 
-    @Nonnull
-    private List<ToolProperty<?>> instantiateProperties(@CheckForNull CNode props, @Nonnull ConfigurationContext context) throws ConfiguratorException {
-        final Configurator<ToolProperty> configurator = context.lookupOrFail(ToolProperty.class);
+    @NonNull
+    private List<ToolProperty<?>> instantiateProperties(@CheckForNull CNode props, @NonNull ConfigurationContext context) throws ConfiguratorException {
         List<ToolProperty<?>> toolProperties = new ArrayList<>();
         if (props == null) {
             return toolProperties;
         }
+        final Configurator<ToolProperty> configurator = context.lookupOrFail(ToolProperty.class);
         if (props instanceof Sequence) {
             Sequence s = (Sequence) props;
             for (CNode cNode : s) {
@@ -110,6 +112,7 @@ public class GitToolConfigurator extends BaseConfigurator<GitTool> {
         return toolProperties;
     }
 
+    @CheckForNull
     @Override
     public CNode describe(GitTool instance, ConfigurationContext context) throws Exception {
         Mapping mapping = new Mapping();
@@ -117,11 +120,11 @@ public class GitToolConfigurator extends BaseConfigurator<GitTool> {
             mapping.put("name", JGitTool.MAGIC_EXENAME);
         } else if (instance instanceof JGitApacheTool) {
             mapping.put("name", JGitApacheTool.MAGIC_EXENAME);
-        } else {
+        } else if (instance != null) {
             mapping.put("name", instance.getName());
             mapping.put("home", instance.getHome());
         }
-        if (instance.getProperties() != null && !instance.getProperties().isEmpty()) {
+        if (context != null && instance != null && instance.getProperties() != null && !instance.getProperties().isEmpty()) {
             final Configurator<ToolProperty> configurator = context.lookupOrFail(ToolProperty.class);
             Sequence s = new Sequence(instance.getProperties().size());
             for (ToolProperty<?> property : instance.getProperties()) {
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
```diff
@@ -0,0 +1,127 @@
...
+            return toolProperties;
+        }
+        if (props instanceof Sequence) {
+            Sequence s = (Sequence) props;
+            for (CNode cNode : s) {
+                toolProperties.add(configurator.configure(cNode, context));
+            }
+        } else {
+            toolProperties.add(configurator.configure(props, context));
+        }
```
fcojfernandez: This if is a bit unclear for me. IIUC, the `if` clause is when there are more than one element in the list, so you're looping over them and using each element as input. So far, it's clear for me. But in the `else` clause, you're using the `props` element as input. Shouldn't it be the content of that object? Something equivalent to `list.get(0)`

src/test/java/org/jenkinsci/plugins/gitclient/JcascTest.java
```diff
@@ -0,0 +1,81 @@
...
+                        hasProperty("name", equalTo("Default")),
+                        hasProperty("home", equalTo("git"))
+                ),
+                allOf(
+                        instanceOf(GitTool.class),
+                        not(instanceOf(JGitTool.class)),
+                        not(instanceOf(JGitApacheTool.class)),
+                        hasProperty("name", equalTo("optional")),
+                        hasProperty("home", equalTo("/opt/git/git"))
+                ))
```
varyvol: How are these two different?

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
```diff
@@ -0,0 +1,127 @@
...
+        p.multiple(true);
+        return Arrays.asList(name, home, p);
+    }
+
+    @Override
+    protected GitTool instance(Mapping mapping, ConfigurationContext context) throws ConfiguratorException {
+        final CNode mproperties = mapping.remove("properties");
+        final String name = mapping.getScalarValue("name");
+        if (JGitTool.MAGIC_EXENAME.equals(name)) {
+            mapping.remove("home"); //Ignored but could be added, so removing to not fail handleUnknown
```
varyvol: What does "ignored but could be added" mean? That it is not taken into account right now?

fcojfernandez: My understanding is that `home` is a field for other GitTool objects but not for this type. In case the yaml file contains that field it's ignored so we don't get an error. If I'm wrong, please correct me @rsandell !

varyvol: Then I think we should raise an error? Otherwise people might think the configuration is actually used, when it's not.

fcojfernandez: Since other GitTools are using the field, it might be expected by users to be used, so putting the field might a recurrent mistake that always would cause an error. On the other hand, it's true that is a bad configuration. @rsandell how is the resulting exported yaml file? Maybe it's easier to make the decision with this input.
Beside, I think the maintainer can shed some light on this decision. CC @MarkEWaite

alecharp: Maybe not a error, but a warning. Doesn't seems to break anything.

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Collections;
+import java.util.List;
+
+@Extension(optional = true)
+public class GitToolConfigurator extends BaseConfigurator<GitTool> {
+
+    @NonNull
+    @Override
+    public String getName() {
+        return "git";
```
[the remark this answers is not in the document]
rsandell: https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/hudson/plugins/git/GitTool.java#L134

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
rsandell: Yes, home is hard coded in jgit and jgitapache, but if someone exported the configuration without this configurator, like an old version of git-tool then the `home` parameter would be there since it is there in all ToolInstallers. So I didn't want to break the config in that case.

src/test/java/org/jenkinsci/plugins/gitclient/JcascTest.java
rsandell: The first one is the default one added by GitTool during startup if there is no configuration. The second one has been manually added and also has tool properties configured.

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Git";
```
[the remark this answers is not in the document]
rsandell: Using the display name of the `@Symbol`

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
fcojfernandez: makes sense. What about a warning as @alecharp proposed?

varyvol: It makes sense not failing then, but I think a warning is not a bad idea.

rsandell: `BaseConfigurator` will run a check after creating the instance to see, among other things, if there is anything left. Since I allow `home` to be set in general but ignored for the two other types I need to remove any potential left overs here so that check doesn't fail. I was toying with the idea of adding a warning log or something. But since I got the `describe` method working I didn't.

src/test/java/org/jenkinsci/plugins/gitclient/JcascTest.java
varyvol: Yes, I got that but the "filtering" is the same: instancess of GitTool and not of JGitTool or JGitApacheTool, so I don't see why in one case one is returned and another one in the other case.

rsandell:
```
hasProperty("name", equalTo("Default")),
hasProperty("home", equalTo("git"))
```
vs.
```
hasProperty("name", equalTo("optional")),
hasProperty("home", equalTo("/opt/git/git"))
```

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
rsandell: Warn where? The system log?

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
fcojfernandez: I think it shouldn't since the configuration is a yaml file, but I son't think it hurts, so fine for me

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
fcojfernandez: I was thinking about the typical logger, like https://github.com/jenkinsci/git-client-plugin/blob/55b186b5ad6c362caeb27d077ff153a6263fba22/src/main/java/hudson/plugins/git/GitTool.java#L197

src/test/java/org/jenkinsci/plugins/gitclient/JcascTest.java
varyvol: Ahhhhh the assert is checking the order in the array. Fine, I got it now, thanks!

pom.xml
```diff
@@ -49,6 +49,7 @@
     <java.level>8</java.level>
     <jgit.version>4.5.7.201904151645-r</jgit.version>
     <jgit.javadoc.version>4.5.0.201609210915-r</jgit.javadoc.version>
+    <jcasc.version>1.29</jcasc.version>
```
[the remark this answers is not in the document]
rsandell: 1.30 has a jenkins dependency of `2.164.1` but git-client's is `2.138.4`

src/test/java/org/jenkinsci/plugins/gitclient/JcascTest.java
rsandell: It's checking that everything defined is in the array, but ignores the specific order or the items :)

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
rsandell: I found a `context.warning` that nicely puts the warning into an administrative monitor. But unfortunately there are other listeners to that warning that interprets it as an error :disappointed:

pom.xml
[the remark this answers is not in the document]
timja: https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/pom.xml#L24
2.60.3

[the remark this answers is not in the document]
casz: you could also switch to BOM and do not worry about version? 😄

casz:
The reason for the custom configurator is to avoid implementing readResolve call in JCasC?

rsandell:
@casz Yes, and also to make the yaml a little bit cleaner.

pom.xml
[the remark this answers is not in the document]
rsandell: @timja correct, I wonder where I got that other version from :thinking:

src/main/java/org/jenkinsci/plugins/gitclient/GitToolConfigurator.java
rsandell: added a jul logger

pom.xml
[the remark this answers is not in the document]
timja: Our integration tests use that version, probably there

PR #475, opened 2019-11-02
Add missing @Override annotations to CliGitAPIImpl

darxriggs:
This makes it consistent with `JGitAPIImpl`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -311,6 +311,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
     /** {@inheritDoc} */
+    @Override
     public GitClient subGit(String subdir) {
         return new CliGitAPIImpl(gitExe, new File(workspace, subdir), listener, environment);
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2789,8 +2834,8 @@ public void deleteBranch(String name) throws GitException, InterruptedException
 
     }
 
-
```
MarkEWaite: Please don't cleanup whitespace as part of a pull request.  It creates many more merge conflicts for pending pull requests (including those closed and labeled with 'Later')

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2912,15 +2959,14 @@ public void execute() throws GitException, InterruptedException {
         };
     }
 
-
```
MarkEWaite: Please don't cleanup whitespace as part of a pull request.  Merge conflicts increase dramatically with white space changes

darxriggs: fixed

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
darxriggs: fixed

PR #476, opened 2019-11-15
Check LFS version for smudge filter

yamhai:
As discussed in the issue: https://issues.jenkins-ci.org/browse/JENKINS-59139 , the smudge filter is no longer active for git >= 2.15 and git-lfs >= 2.3.0 .

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -287,6 +287,29 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         return gitVersion >= requestedVersion;
     }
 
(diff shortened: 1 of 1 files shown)


MarkEWaite:
Needs tests to confirm correct behavior when CLI git is not new enough and when it is new enough.

slietzau:
If the LFS backend requires authentication this is a breaking change because existing jobs would fail and require one of the workarounds below.

The current workaround I'm using is to not use the additional "Git LFS Pull after checkout" step and run `git lfs pull` manually in a controlled environment after after the checkout.

With this merged I'd have to either disable git lfs hooks globally on all slaves or set `GIT_LFS_SKIP_SMUDGE=1` manually in my job.

MarkEWaite:
I don;t see any activity on this pull request from the submitter.  I'm closing it.  We can reopen it if further activity happens.

PR #478, opened 2019-11-19
Switch from README.md to README.adoc in tests

MarkEWaite:
## Switch from README.md to README.adoc in tests

Tests intentionally reuse data from the git client plugin master branch. Now that the README is in adoc format, the file name changes in the tests to match what is stored on the file system.

## Checklist

## Types of changes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,7 +7,7 @@ If it fixes a bug or resolves a feature request, include a link to the issue.
 
 _Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. This is simply a reminder of what we are going to look for before merging your code._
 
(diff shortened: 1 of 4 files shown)


jglick:
> Tests intentionally reuse data from the git client plugin master branch.

@MarkEWaite this setup is unwise IMO. The change in `master` [broke PCT after Nov 15](https://ci.jenkins.io/job/Tools/job/bom/job/master/). I see that you backported this as #479, which is good but should not have been necessary (and is currently unreleased so not eligible for use in PCT). Test status should, to the extent humanly possible, be a deterministic function of the contents of the repository checkout and not depend on external factors (other than well-versioned dependencies). If you need test data, [package it up](https://git-scm.com/book/en/v2/Git-Tools-Bundling) and add it to `src/test/resources/`.

MarkEWaite:
Thanks @jglick. You make a very good point. I don't know how soon I'll be able to resolve that, but I agree that the test should be fixed to not depend on external repositories.

PR #484, opened 2019-12-19
Check git url as well as credential type before going down SSH route.

mhaley-miovision:
## [JENKINS-60329](https://issues.jenkins-ci.org/browse/JENKINS-60329) - Help make deciding to use SSH or not more explicit

The AWS credentials manager implements both SSHUserPrivateKey and StandardUsernamePasswordCredentials.  When that credential source is used with the git client plugin it has the unfortunate side effect of only checking what instance type the credential object is.  We have more information available to help make our decision clearer (mainly the github url).

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc) doc

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1943,7 +1943,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             }
         }
         try {
-            if (credentials instanceof SSHUserPrivateKey) {
+            if (credentials instanceof SSHUserPrivateKey && !("http".equalsIgnoreCase(url.getScheme()) || "https".equalsIgnoreCase(url.getScheme()))) {
                 SSHUserPrivateKey sshUser = (SSHUserPrivateKey) credentials;
                 listener.getLogger().println("using GIT_SSH to set credentials " + sshUser.getDescription());
 


MarkEWaite:
Please provide a test which shows the failure you're trying to prevent.  Code changes without tests are much less likely to be accepted.

mhaley-miovision:
Hi @MarkEWaite, thanks for replying so quickly.  I really just wanted to get a quick build going to test this out and see if it would even work.  Here is what I discovered:

We are using the AWS secrets manager credentials provider plugin.  It has an `AwsCredentials` object that implements both `StandardUsernamePasswordCredentials` and `SSHUserPrivateKey` interfaces.  This breaks some assumptions that the `git-plugin` and the `github-branch-source` have made.  Both of these plugins (and possibly others) check to see if the credentials they found are of instance type `SSHUserPrivateKey` and if so go down the route of doing some special SSH stuff and ultimately break when they try to use the secret as an ssh key when it really is an api token.  

Two such examples of this are:
* [GitHubSCMBuilder in the github-branch-source plugin](https://github.com/jenkinsci/github-branch-source-plugin/blob/3d1702dbd290ae9cf872ee65b47b16b105ae9032/src/main/java/org/jenkinsci/plugins/github_branch_source/GitHubSCMBuilder.java#L247)
* [CliGitAPIImpl in the git-client plugin](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L1946)

Seeing now that this is not an isolated case in this plugin, I am going to close this PR in favour of a PR for the aws credentials management plugin to create different types of credentials based on some optional configuration via tags.

PR #486, opened 2019-12-29
[JENKINS-60316] Fix escaping single quotes in Unix file paths

darxriggs:
## [JENKINS-60316](https://issues.jenkins-ci.org/browse/JENKINS-60316)

Points addressed
- a single quote is not allowed between single quotes, even if escaped
- a backslash requires escaping if to be used literally in a reqular expression

This escaping has previously already been used in `quoteUnixCredentials` which was added in commit 62872bccaf3f0e02c663acfdd85be2b38bddf1e8 - see [here](https://github.com/jenkinsci/git-client-plugin/commit/62872bccaf3f0e02c663acfdd85be2b38bddf1e8#diff-4a13310f7c84a9d6e133def48c0aed5aR1449-R1453).

Also see https://www.gnu.org/software/bash/manual/html_node/Single-Quotes.html.

I tested it manually on Linux, with a SSH key and a username/password credential, with a job inside a folder that contains spaces and single quotes.

There is no test infrastructure available yet to test such things automatically - right?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2078,7 +2078,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      */
     private String windowsArgEncodeFileName(String filename) {
         if (filename.contains("\"")) {
-            filename = filename.replaceAll("\"", "^\"");
+            filename = filename.replace("\"", "^\"");
         }
         return "\"" + filename + "\"";
     }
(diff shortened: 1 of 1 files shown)


darxriggs:
@MarkEWaite do you want to consider this for the next release?

MarkEWaite:
@darxriggs I made a first attempt to duplicate the problem from a Jenkins installation last week before I released 3.1.0-beta.  I was unable to duplicate the problem.

Since I could not duplicate the problem, I was unwilling to include a fix that I could not confirm actually fixed the issue.

This fix will need to wait for a release of git client plugin after 3.1.0.

I would love to have an automated test that shows the problem or an acceptance test on a branch in my [jenkins-bugs repository](https://github.com/MarkEWaite/jenkins-bugs) or an acceptance test in my [docker-lfs repository](https://github.com/MarkEWaite/docker-lfs).

darxriggs:
> I tested it manually on Linux, with a SSH key and a username/password credential, with a job inside a folder that contains spaces and single quotes.

@MarkEWaite This info was not helpful to reproduce it? It could reproduce it that way.

darxriggs:
I put the automated test on my todo list. I am just not sure when I will have a look at it.

MarkEWaite:
On my TODO list as well.  The interactive testing exercise exposed configuration errors in my CI setup.  Thanks for fixing the bug and for the positive impact that had on my ability to reliably test credentials issues in my environment.

PR #498, opened 2020-01-28
Checking tests on CI =/

timja:


MarkEWaite:
Thanks for that pull request.  I think I understand the problem and how to solve it.  The test depends on the names of the branches on the authoritative repository.  That was fine when the branches on the authoritative repository were solely managed by the maintainers.  Now that dependabot is enabled, the branch names are more fluid than the test will accept.  Should be an easy fix to that test.

PR #501, opened 2020-02-05
JGit timeout support (extended)

darxriggs:
This change adds a timeout to all JGit commands that not yet supported it. Only the `CloneCommand` so far used a timeout.

A slight difference between CliGit and JGit remains
* CliGit: timeout for the whole command
* JGit: timeout only for transport step (therefore missing for `CheckoutCommand`)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
@@ -553,6 +553,7 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             private List<RefSpec> refspecs;
             private boolean shouldPrune = false;
             private boolean tags = true;
+            private Integer timeout;
 
             @Override
             public org.jenkinsci.plugins.gitclient.FetchCommand from(URIish remote, List<RefSpec> refspecs) {
@@ -582,7 +583,7 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
             @Override
             public org.jenkinsci.plugins.gitclient.FetchCommand timeout(Integer timeout) {
-                // noop in jgit
+                this.timeout = timeout;
                 return this;
             }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -1473,7 +1477,9 @@ else if (!referencePath.isDirectory())
                             .setCredentialsProvider(getProvider())
                             .setTagOpt(tags ? TagOpt.FETCH_TAGS : TagOpt.NO_TAGS)
                             .setRefSpecs(refspecs);
-                    if (timeout != null) fetch.setTimeout(timeout);
+                    if (timeout != null) {
+                        fetch.setTimeout(timeout * 60);
```
MarkEWaite: Unfortunately, this won't work in a way that is consistent with the timeout implementation in the command line git implementation.

Command line git timeout sets the maximum duration of a single git command.

JGit FetchCommand.setTimeout() sets the maximum period during which there was not communication on the JGit transport.  Refer to comments on [my failed attempt](https://github.com/MarkEWaite/git-client-plugin/commit/ecaf1b2bad99fcfb6c4b44e0374513fd27d279f2) to add timeout using this technique.

darxriggs: I remembered a comment from you a while back about this but was not aware of the referenced commit.

darxriggs: I had this small change locally already prepared some months ago as you can see by the commit dates and just wanted to get it out the door. Unfortunately I would not want to spent much more time on this. So it would also be ok for me that we close the pull request unmerged.

MarkEWaite: Yes, I was so happy to discover the setTimeout() API on JGit.  It was only when writing several tests that should have failed that I discovered that the JGit concept of timeout is well suited to an API while the git plugin concept of timeout is well suited to an end user.

MarkEWaite: @darxriggs I think we should close this pull request unless you plan to implement the JGit timeout with separate threads or some other technique outside the JGit timeout method.

What do you think?

PR #502, opened 2020-02-05
[JENKINS-60940] Porting CliGitAPIImplTest to GitClientTest

rishabhBudhouliya:
## [JENKINS-60940](https://issues.jenkins-ci.org/browse/JENKINS-60940)

Tests from CliGitAPIImplTest have been ported to GitClientTest

## Checklist


src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -90,6 +91,9 @@
     /* Instance of object under test */
     private GitClient gitClient = null;
 
+    /* Instance of object of another test class*/
+    private CliGitAPIImplTest cliGitAPIImplTest = new CliGitAPIImplTest();
```
MarkEWaite: I would prefer to move the definition of the VersionTest inner class from the CliGitAPIImplTest class to this class rather than reference it here.  Eventually we hope to remove the CliGitAPIImplTest class completely so that it spends no time executing tests in that file.

rishabhBudhouliya:
@MarkEWaite Hi, thanks for taking out the time to review the PR. 

After your suggestions, I have done two things:

- Shifted all the necessary functionalities of CliGitAPIImplTest class to GitClientTest (that includes VersionTest)

- Deleted CliGitAPIImplTest because there was no need for it.

One thing necessary to mention here is the need to port GitAPITestCase because it contains most of the test cases extended by CliGit and JGit tests. Also, apart from `@before`, `@after`(Junit 3 -> 4) changes in GitAPITestCase, there is a need to make the tests more coherent to GitClientTest test implementations.

MarkEWaite:
Thanks very much.  I don't think we are ready to delete CliGitAPIImplTest yet.  The tests from GitAPITestCase are only executed for the command line git implementation because CliGitAPIImplTest extends GitAPITestCase.  The GitClientTest is a parameterized test that replaces the inheritance technique with a parameter that has values "git", "jgit", and "jgit-apache".

I'll review the specific changes as well, but wanted to reply on your specific deletion of CliGitAPIImplTest.

While thinking about VersionTest and how it interacts with the parameterized GitClientTest, I wonder if my advice to copy it into GitClientTest should be replaced with placing it in a completely independent test.  Placing the VersionTest into GitClientTest will cause it to be executed 3 times, once for each value of the parameter.  Only 1 of those 3 executions is needed. The others are a waste of time and output.

What do you think?

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -2026,6 +2029,278 @@ public void testFetchOverwritesExistingTags() throws Exception {
...
+    public void test_git_version_ubuntu_14_04_ppa_2_3_5() {
+        VersionTest[] versions = {
+                new VersionTest(true,  2, 3, 5, 0),
+                new VersionTest(true,  2, 2, 9, 9),
+                new VersionTest(false, 2, 3, 5, 1),
+                new VersionTest(false, 2, 4, 0, 0),
+                new VersionTest(false, 3, 0, 0, 0)
+        };
+        doTest("git version 2.3.5", versions);
+    }
```
MarkEWaite:
```suggestion
    }

```

rishabhBudhouliya:
Hi, in light of your suggestion, I tried finding a solution to where we could separate test cases which needed only a selective parameter and I found the [Runner Class](http://junit.org/javadoc/latest/org/junit/experimental/runners/Enclosed.html ) but upon some inspection, this works only when we have some parameterized and non-parameterized test cases in the same class.

Initially I didn't understand the need of CliGitAPIImplTest anymore, but after what you have said and the need to place classes like VersionTest in a separate test class makes it more obvious.

So we can either maintain VersionTest class in CliGitAPIImplTest and derive instances of these in GitClientTest (I'll revert my changes), or we could create a separate utility class. The role of the utility class would become clearer when other test classes have been ported as well.

What would you suggest?

MarkEWaite:
Good insights. Let's return to your original implementation that references VersionTest from inside CliGitAPIImplTest so that we can bring this pull request to completion.  As you continue working on the changes, I think you'll find which works best based on your experiences with the code.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -2026,6 +2030,294 @@ public void testFetchOverwritesExistingTags() throws Exception {
...
+                cliGitAPIImplTest.new VersionTest(false, 3, 0, 0, 0)
+        };
+        cliGitAPIImplTest.doTest("git version 2.3.5", versions);
+    }
+
+    @Test
+    public void test_git_ssh_executable_found_on_windows() {
+        if (!gitImplName.equals("git")) {
+            return;
+        }
```
MarkEWaite: JUnit offers a more elegant way to skip a test when some condition is incorrect.

```
        assumeThat(gitImplName, is("git"));
```

It is used elsewhere in the file and is understood by test reporting frameworks better than the early return technique.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -2026,6 +2030,294 @@ public void testFetchOverwritesExistingTags() throws Exception {
...
+
+        CliGitAPIImpl git = new CliGitAPIImpl("git", new File("."), cliGitAPIImplTest.listener, cliGitAPIImplTest.env);
+        assertTrue("ssh.exe not found", git.getSSHExecutable().exists());
+    }
+
+    @Test
+    public void test_git_branch_with_line_breaks_and_long_strings() {
+        if (!gitImplName.equals("git")) {
+            return;
+        }
```
MarkEWaite: Same comment as earlier block - use assumeThat rather than an early exit.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
rishabhBudhouliya: I just saw AssumeThat, changed it. Thanks!

PR #504, opened 2020-02-10
Some GitAPITestCase tests were replaced in GitClientTest

Loghijiaha:
## Checklist

## Types of changes

## Further comments

Junit 3 tests can be sorted according to alphabet order in GitAPITestCase and compared with GItClientTestCases. It would be an efficient idea to change the tests to Junit 4.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
@@ -44,6 +44,7 @@ import org.eclipse.jgit.transport.RefSpec;
 import org.eclipse.jgit.transport.URIish;
 import org.eclipse.jgit.lib.Constants;
 
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.*;
 import org.junit.AfterClass;
 import static org.junit.Assert.*;
@@ -599,7 +600,7 @@ public class GitClientTest {
         assertEquals(upstreamRepoURL, gitClient.getRemoteUrl("upstream"));
     }
 
-    @Test
+    @Test (expected = GitException.class)
     public void testAutocreateFailsOnMultipleMatchingOrigins() throws Exception {
         File repoRootTemp = tempFolder.newFolder();
         GitClient gitClientTemp = Git.with(TaskListener.NULL, new EnvVars()).in(repoRootTemp).using(gitImplName).getClient();
@@ -608,11 +609,11 @@ public class GitClientTest {
         FilePath gitClientTempFile = gitClientFilePath.createTextTempFile("aPre", ".txt", "file contents");
         gitClientTemp.add(".");
         gitClientTemp.commit("Added " + gitClientTempFile.toURI().toString());
-        gitClientTemp.clone_().url("file://" + repoRootTemp.getPath()).execute();
+        gitClient.clone_().url("file://" + repoRootTemp.getPath()).execute();
         final URIish remote = new URIish(Constants.DEFAULT_REMOTE_NAME);
 
         try ( // add second remote
-              FileRepository repo = new FileRepository(repoRootTemp)) {
+              FileRepository repo = new FileRepository(new File(repoRoot, ".git"))) {
             StoredConfig config = repo.getConfig();
             config.setString("remote", "upstream", "url", "file://" + repoRootTemp.getPath());
             config.setString("remote", "upstream", "fetch", "+refs/heads/*:refs/remotes/upstream/*");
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -597,6 +599,38 @@ public void testAddRemoteUrl() throws Exception {
...
+        gitClientTemp.fetch_().from(remote, refspecs).execute();
+        refspecs = Collections.singletonList(new RefSpec(
+                "refs/heads/*:refs/remotes/upstream/*"));
+        gitClientTemp.fetch_().from(remote, refspecs).execute();
+
+        try {
+            gitClientTemp.checkout().ref(Constants.MASTER).execute();
+            fail("GitException expected");
+        } catch (GitException e) {
+            // expected
```
MarkEWaite: I fear that the exception in this case is probably different than the exception in the test that has been deleted.  Unfortunately, the test is not asserting enough about the exception.  Would you be willing to improve the original test by asserting the contents of the exception?

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -597,6 +599,38 @@ public void testAddRemoteUrl() throws Exception {
         assertEquals(upstreamRepoURL, gitClient.getRemoteUrl("upstream"));
     }
 
+    @Test
+    public void testAutocreateFailsOnMultipleMatchingOrigins() throws Exception {
+        File repoRootTemp = tempFolder.newFolder();
+        GitClient gitClientTemp = Git.with(TaskListener.NULL, new EnvVars()).in(repoRootTemp).using(gitImplName).getClient();
+        gitClientTemp.init();
```
MarkEWaite: Seems like it is missing the initial commit that was added in the GitAPITestCase version of the test.  Maybe this would work (didn't compile it, just offering something based on javadoc exploration)?
```suggestion
        gitClientTemp.init();
        FilePath gitClientFilePath = gitClientTemp.getWorkTree()
        FilePath gitClientTempFile = gitClientFilePath.createTextTempFile("aPre", ".txt", "file contents");
        gitClientTemp.add(".");
        gitClientTemp.commit("Added " + gitClientTempFile.toURI().toString());
```

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -597,6 +599,38 @@ public void testAddRemoteUrl() throws Exception {
...
+        gitClientTemp.clone_().url("file://" + repoRootTemp.getPath()).execute();
```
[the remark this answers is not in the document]
Loghijiaha: Sorry for the late reply. In the original test( GitAPITestCase ), a new working area has created and made an initial empty commit. The repo was cloned to the current working folder and tested the multiple matching of origins. When I debugged, here it creates a cycle by pointing itself as you mentioned.

Hence, I will change the `gitClientTemp` to `gitClient` when I clone the new repo from `repoRootTemp` and following fetches as well.I think that could make more sense than this.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
Loghijiaha: Yes, I am happy to improve by writing some assertion about the exception.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
Loghijiaha: Yes, you are right @MarkEWaite, I was wondering to add an empty commit which created an failure in CliGit but JGit and JGitApache were successful on that. This would be the right solution. Thank you.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -597,6 +600,41 @@ public void testAddRemoteUrl() throws Exception {
...
+        // fill both remote branches
+        List<RefSpec> refspecs = Collections.singletonList(new RefSpec(
+                "refs/heads/*:refs/remotes/origin/*"));
+        gitClient.fetch_().from(remote, refspecs).execute();
+        refspecs = Collections.singletonList(new RefSpec(
+                "refs/heads/*:refs/remotes/upstream/*"));
+        gitClient.fetch_().from(remote, refspecs).execute();
+
+        // checkout will fail
+        gitClient.checkout().ref(Constants.MASTER).execute();
```
MarkEWaite: I assume this is the statement that throws the assertion.  If so, then the next two statements won't be executed.  In that case, it may be better to use the try / catch that you had before, but then insert an assertion into the catch block which checks some specific value of the GitException.
```suggestion
        gitClient.checkout().ref(Constants.MASTER).execute();
```

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -597,6 +600,41 @@ public void testAddRemoteUrl() throws Exception {
         assertEquals(upstreamRepoURL, gitClient.getRemoteUrl("upstream"));
     }
 
+    @Test (expected = GitException.class)
```
MarkEWaite: I think this is too broad.  Any GitException thrown anywhere in the test method will cause the test to be marked as successful,  even if the exception was in an unexpected location.

The `(expected = GitException.class)` technique is best used for very short tests with very few conditions which could fail or with distinct exceptions rather than overused and generalized exceptions like GitException.
```suggestion
    @Test
```

Loghijiaha: I will make it out. Thanks.

MarkEWaite:
This looks great!  Thanks very much @Loghijiaha

PR #507, opened 2020-02-14
[JENKINS-60940] Add WorkspaceWithRepoRule and two test classes

rishabhBudhouliya:
## [JENKINS-60940](https://issues.jenkins-ci.org/browse/JENKINS-60940) - Add  *WorkspaceWithRepoRule* as a test repository

**WorkspaceWithRepoRule**
- This is *not* an abstract class, it can create a working test repository using the `GitClientSampleRepoRule` class.
- It covers methods like initializing bare repos, cloning a particular repo, providing a local mirror repository and many more utility functions.

**GitClientCloneTest**
- This is a parameterized *test* class which uses the WorkspaceWithRepoRule to create test repos and test functions.
- It is a logical grouping of tests related to git_clone and timeout tests

**GitClientFetchTest**
- This is also a parameterized test class which contains related to git_fetch

**Important Note**: GitSampleRepoRule has been ported from git-plugin to git-client-plugin as GitClientSampleRepoRule and a dependency for scm-api has been added in pom.xml

## Checklist


src/test/java/org/jenkinsci/plugins/gitclient/WorkspaceWithRepoRule.java
```diff
@@ -0,0 +1,145 @@
+package org.jenkinsci.plugins.gitclient;
+
+import hudson.EnvVars;
+import hudson.Launcher;
+import hudson.model.TaskListener;
+import hudson.plugins.git.GitException;
+import javafx.concurrent.Task;
```
MarkEWaite: This is not the Task you want.
```suggestion
```

MarkEWaite:
@rishabhBudhouliya I was expecting there would be some deletions in the GitAPITestCase as a result of these tests being created.  Would you be willing to remove the tests from GitAPITestCase that are now covered in these tests?

src/test/java/org/jenkinsci/plugins/gitclient/WorkspaceWithRepoRule.java
```diff
@@ -0,0 +1,144 @@
...
+        return Git.with(listener, new EnvVars()).in(gitFileDir).using(gitImplName).getClient();
+    }
+
+    protected String localMirror() throws IOException, InterruptedException {
+        File base = new File(".").getAbsoluteFile();
+        for (File f=base; f!=null; f=f.getParentFile()) {
+            if (new File(f,"target").exists()) {
+                File clone = new File(f, "target/clone.git");
+                if (!clone.exists()) {  // TODO: perhaps some kind of quick timestamp-based up-to-date check?
+                    cliGitCommand.run("git", "clone", "--mirror", repoURL, clone.getAbsolutePath());
```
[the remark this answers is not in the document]
rishabhBudhouliya: I was actually confused between the use of CliGitCommand or launchCommand (which uses the Launcher to run cmds). 
Since I am using the RepoRule, I thought of using CliGitCommand only to notice that it returns `String[]` instead of a string (which is done by launchCommand). 
To replace launchCommand with gitCommand would have required considerable re-work so I used launchCommand. 

And yes, it was working on my MacBook. Thank you for pointing out the mistake!

rishabhBudhouliya:
> @rishabhBudhouliya I was expecting there would be some deletions in the GitAPITestCase as a result of these tests being created. Would you be willing to remove the tests from GitAPITestCase that are now covered in these tests?

Thanks for this comprehensive review Mark! I didn't delete the test cases from GitAPITestCase because I had my doubts about this implementation being used instead of the old one. 

Yes, I am willing to remove those tests from GitAPITestCase that are now covered in these tests and also, learn from your recent commits and try to re-work the exisiting test cases in accordance to JUnit 4 principles.

MarkEWaite:
Thanks so much for your patience in my long delay to review the code.  More review is still needed, but I think that you can apply some of the things from my review of GitClientCloneTest to the other test.

rishabhBudhouliya:
Sure Mark! I will apply suitable changes to the other test. Do you want me follow this pattern and create other test classes as well or would we want to hold that for now?

MarkEWaite:
I would love to have additional tests, though I'm more concerned about getting this pull request complete and merged before additional tests are moved.

You'll also need to balance time on this with time to create one or more project proposals for Google Summer of Code.  Since the Jenkins project has been accepted to GSoC, I believe it is now time that students should be preparing their project proposals.

rishabhBudhouliya:
That's a good point, and I agree on the need of balancing both. I know this not the place to say this, but I am thinking of writing the proposal for both, Git Repo caching on agent and Git plugin performance improvements.

Okay, so I have deleted the already ported test cases and one more thing I can do is to fix GitClientFetchTest as discussed. After these changes and your review we can hopefully merge this PR.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientFetchTest.java
```diff
@@ -312,16 +306,21 @@ public void test_fetch_needs_preceding_prune() throws Exception {
...
-        }
+//        try {
+//            /* Fetch parent/a into newArea repo - fails for
+//             * CliGitAPIImpl, succeeds for JGitAPIImpl */
+//            newAreaWorkspace.launchCommand("git", "config", "fetch.prune", "false");
+//            newAreaWorkspace.getGitClient().fetch(new URIish(newAreaWorkspace.getGitFileDir().toString()), refSpecs);
+//            assertTrue("CliGit should have thrown an exception", newAreaWorkspace.getGitClient() instanceof JGitAPIImpl);
+//        } catch (GitException ge) {
+//            final String msg = ge.getMessage();
+//            assertTrue("Wrong exception: " + msg, msg.contains("some local refs could not be updated") || msg.contains("error: cannot lock ref "));
```
rishabhBudhouliya: I have left this code block commented because I was not sure if thrown.expectMessage is correctly implementing the assert condition here.

For now, I have coded two thrown.expectMessages instead of placing an || operator.

rishabhBudhouliya: Also, is it a good practice to use the same ExpectedException rule for two different types of errors in the same test class? Shouldn't we have a different test class for testing the other exception (Despite it is a GitException, it conveys a different message)?

rishabhBudhouliya:
Thanks for the detailed commits on GitClientCloneTest, Mark!
- I have used them as a reference to correct some test cases in GitClientFetchTest.
- Also deleted the ported tests from GitAPITestCase

rishabhBudhouliya:
I have an observation I'd like to share here: 

While creating a test using a git repo created by a "GitSampleRepoRule", does the ".git" persist _after_ the test is over? 
If not so, I can see that the local mirror(s) (cloning a remote repo into the test workspace) persist even after the test is over (though they have no git objects). 

Is this normal behavior?

MarkEWaite:
The intent with the local archive is that it should be usable in all the tests, while a GitSampleRepoRule is used in a single test.  The local archive is "contaminated" by not containing precisely the same history at every run across all time.  The GitSampleRepoRule starts empty and can be assembled to contain precisely the expected content.

rishabhBudhouliya:
> The intent with the local archive is that it should be usable in all the tests, while a GitSampleRepoRule is used in a single test. The local archive is "contaminated" by not containing precisely the same history at every run across all time. The GitSampleRepoRule starts empty and can be assembled to contain precisely the expected content.

Okay, this makes sense. While I'm looking at writing some performance micro-benchmarks for git-clients, I'm thinking of using GitSampleRepoRule as the test repo to conduct those tests. Does that seem like a right approach to you?

MarkEWaite:
Yes, that seems like the right approach to me.  Better a local repository than a repository over the network.

rishabhBudhouliya:
> Yes, that seems like the right approach to me. Better a local repository than a repository over the network.

Okay, great. 
Yes, I wouldn't want to test the performance using the network because of the unintended variance induced by external factors.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientFetchTest.java
MarkEWaite: I see a misunderstanding in the use of expectedException in `test_fetch_needs_preceding_prune`.  I'm glad you raised the question, because I had not run the test in a debugger to confirm that it was reaching the end of the test.

The last executed statement of the test is the statement that is expected to throw the exception.  Any statements after the statement that throws the exception will not be executed.  In this case, the "Prune when a remote is not yet defined" declares that an exception is expected, then the `prune` call throws the exception.  All statements after that prune are ignored.

I will review at least some of the proposed changes to see if we can find a way to use the JUnit 4 facilities well and still test the paths that were tested before.  I suspect the change will be to create multiple tests from the original single overly large test.

MarkEWaite: I just read the JUnit wiki for exception handling.  They note that [ExpectedException has been deprecated](https://github.com/junit-team/junit4/wiki/Exception-testing#expectedexception-rule) in JUnit 4.13.  My apologies that I had you learn a technique that is already deprecated.

I'll need to read that wiki page in more detail to understand the lambda based method that is preferred for asserting exceptions.

MarkEWaite: I've submitted some small commits to [my copy of the branch](https://github.com/MarkEWaite/git-client-plugin/commits/JENKINS-60940) while I work through the test classes.

Thanks so much for the pull request!

rishabhBudhouliya: I've gone through your commits. All the discussion and changes is great for the development of the other test classes as well.

Would prefer a new PR with all the changes you've made or would you rather merge this and merge changes after that?

MarkEWaite: I'll likely push my changes to your branch so that we continue evolving this pull request.  I'm struggling currently with a test that was transitioned and needed a significant change in the transition.

I believe that the `test_fetch_needs_preceding_prune` test was added fairly recently to GitAPITestCase, yet the updated copy in my working repo needs the key expected exception removed.

rishabhBudhouliya: https://github.com/MarkEWaite/git-client-plugin/blob/f5540088eb6f62718aff69d6e72e5e524f8eee77/src/test/java/org/jenkinsci/plugins/gitclient/GitClientFetchTest.java#L296-L306. I've seen that you have commented the part of code which asserts a throw. 

I don't understand it, if the client is an instance of CliGitAPIImpl, we expect it to throw a git exception?

MarkEWaite: Yes, that's an unclear way of saying that JGit does not throw this exception but command line git does throw an exception.

rishabhBudhouliya: ```newAreaWorkspace.getGitClient().fetch(new URIish(newAreaWorkspace.getGitFileDir().toString()), refSpecs);```

I think this is my mistake Mark, this line makes no sense. I am trying to fetch all branches from the same repository calling the fetch.

This should be 
```newAreaWorkspace.getGitClient().fetch(new URIish(bareWorkspace.getGitFileDir().toString()), refSpecs);``` 
So we are trying to fetch parent/a from the bare repository.

MarkEWaite: Ah, I'm so glad you detected that!  I failed to look at the source that was being used for the clone!  Thanks very much.

MarkEWaite:
Ready to merge once CI passes.

rishabhBudhouliya:
That's great, thank you!

PR #514, opened 2020-02-21
[JENKINS-57205] Packed lightweight tags must resolve to meaningful object IDs lest they cause NPEs downstream

brianeray:
## [JENKINS-57205](https://issues.jenkins-ci.org/browse/JENKINS-57205) - Packed lightweight tags must resolve to meaningful object IDs lest they cause NPEs downstream

Between Git Client 2.9.0 and 3.0.0 the bundled JGit bumped up from 4.5.7 to 5.5.1. Starting with JGit 4.9.0 upon repo clone, references are packed instead of left loose. All well and good save that `JGitClientAPIImpl.getTags()` trips up over ligthweight tags packed thus, resolving them to null SHA1s. Downstream anything trying to process them is at high risk of an NPE--as in the pre-build merge scenario in the attached JIRA.

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
@@ -2792,18 +2792,19 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 /* Prefer peeled ref if available (for tag commit), otherwise take first tag reference seen */
                 String tagName = entry.getKey();
                 Ref tagRef = entry.getValue();
-                if (!entry.getValue().isPeeled()) {
+                if (!tagRef.isPeeled()) {
                     Ref peeledRef = repo.peel(tagRef);
                     if (peeledRef.getPeeledObjectId() != null) {
                         tagRef = peeledRef; // Use peeled ref instead of annotated ref
                     }
                 }
-                if (tagRef.isPeeled()) {
+                /* Packed lightweight (non-annotated) tags can wind up peeled with no peeled obj ID */
+                if (tagRef.isPeeled() && tagRef.getPeeledObjectId() != null) {
                     peeledTags.add(new GitObject(tagName, tagRef.getPeeledObjectId()));
                 } else if (!tagNames.contains(tagName)) {
                     peeledTags.add(new GitObject(tagName, tagRef.getObjectId()));
                 }
-                tagNames.add(entry.getKey());
+                tagNames.add(tagName);
             }
         }
         return peeledTags;
--- a/src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
@@ -1,0 +1,162 @@
+package org.jenkinsci.plugins.gitclient;
+
+import hudson.EnvVars;
+import hudson.model.TaskListener;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import hudson.plugins.git.GitException;
+import hudson.plugins.git.GitObject;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
+import java.util.List;
+import java.util.Set;
+
+// Collides with implicit org.jenkinsci.plugins.gitclient.Git.
+//import org.eclipse.jgit.api.Git;
+import static org.eclipse.jgit.lib.Constants.HEAD;
+import org.eclipse.jgit.internal.storage.file.FileRepository;
+import org.eclipse.jgit.internal.storage.file.GC;
+import org.eclipse.jgit.lib.ObjectId;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.rules.TemporaryFolder;
+
+import org.jvnet.hudson.test.Issue;
+
+/**
+ * JGit Client-specific tests. In their own Test file to ward off bloat in 
+ * GitClient and minimized parameterized and other test logic conditional 
+ * on Git implementation.
+ *
+ * @author Brian Ray
+ */
+public class JGitClientTest {
+    /* These tests are only for the JGit client. */
+    private static final String GIT_IMPL_NAME = "jgit";
+
+    /* Instance under test. */
+    private GitClient gitClient;
+
+    /* Lower level "porcelain" API when gitClient can't do something. */
+    private org.eclipse.jgit.api.Git gitAPI;
+
+    /* Repo garbage collector. */
+    private GC gc;
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Rule
+    public TemporaryFolder tempFolder = new TemporaryFolder();
+    private File repoRoot;
+
+    @Before
+    public void setGitClientEtc() throws IOException, InterruptedException {
+        repoRoot = tempFolder.newFolder();
+        gitClient = Git.with(TaskListener.NULL, new EnvVars())
+                       .in(repoRoot)
+                       .using(GIT_IMPL_NAME)
+                       .getClient();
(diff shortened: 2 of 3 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
```diff
@@ -0,0 +1,162 @@
...
+import org.jvnet.hudson.test.Issue;
+
+/**
+ * JGit Client-specific tests. In their own Test file to ward off bloat in 
+ * GitClient and minimized parameterized and other test logic conditional 
+ * on Git implementation.
+ *
+ * @author Brian Ray
+ */
+public class JGitClientTest {
```
brianeray: Test class name is a little artificial. There's no underlying {{JGitClient}} class. But the choice of name is for symmetry with {{GitClientTest}} and for reasons of refactoring away from old JUnit 3 test classes as [discussed here](https://issues.jenkins-ci.org/browse/JENKINS-57205?focusedCommentId=385290&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-385290).

All that said I can rename it to anything that makes more sense.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -2792,18 +2792,19 @@ private StoredConfig getConfig(String GIT_DIR) throws GitException {
                 /* Prefer peeled ref if available (for tag commit), otherwise take first tag reference seen */
                 String tagName = entry.getKey();
                 Ref tagRef = entry.getValue();
-                if (!entry.getValue().isPeeled()) {
+                if (!tagRef.isPeeled()) {
```
brianeray: Mild goldplating to get more mileage out of locals `tagRef` and `tagName` below. If it's too much I can back these out.

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
```diff
@@ -0,0 +1,162 @@
...
+    /* These tests are only for the JGit client. */
+    private static final String GIT_IMPL_NAME = "jgit";
+
+    /* Instance under test. */
+    private GitClient gitClient;
+
+    /* Lower level "porcelain" API when gitClient can't do something. */
+    private org.eclipse.jgit.api.Git gitAPI;
```
MarkEWaite: If you'd be willing, I'd prefer this have a name which more clearly identifies it as coming from JGit directly.  The API naming confusion is unfortunate, but we're stuck with them both.
```suggestion
    private org.eclipse.jgit.api.Git jGitAPI;
```

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
```diff
@@ -0,0 +1,162 @@
...
+    private File repoRoot;
+
+    @Before
+    public void setGitClientEtc() throws IOException, InterruptedException {
+        repoRoot = tempFolder.newFolder();
+        gitClient = Git.with(TaskListener.NULL, new EnvVars())
+                       .in(repoRoot)
+                       .using(GIT_IMPL_NAME)
+                       .getClient();
+        FileRepository repo = (FileRepository) gitClient.getRepository();
```
MarkEWaite: I'd prefer that the test not use the deprecated `getRepository()` API.  It was deprecated long ago and in this case, it is easier to use the `repoRoot` File that is known to point to the root of the git repository.

If necessary, it could be moved after the `gitClient.init_()` call so that it is looking at a repository that is already initialized.
```suggestion
        FileRepositoryBuilder builder = new FileRepositoryBuilder();
        Repository repo = builder.setGitDir(repoRoot).findGitDir().build();
        // Repository repository = builder.setGitDir(new File(repoRoot, ".git")).findGitDir().build();
```

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private void packRefs() throws IOException {
+        gc.packRefs();
+    }
+
+    // No flavor of GitClient has a tag(String) API, only tag(String,String). 
+    // But sometimes we want a lightweight a.k.a. non-annotated tag.
+    private void tag(String name) throws GitException {
```
MarkEWaite: To make it clear in callers, could this be called:
```suggestion
    private void lightweightTag(String name) throws GitException {
```

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
MarkEWaite: I think that `JGitClientTest` is a little too general for the specific case this is testing.  Would you consider renaming it to something like `JGitLightweightTagTest`?  We may even want to include a lightweight tag test in the CLI git code, and could reuse your JGit implementation that creates lightweight tags so that we don't need to extend the git client API to create lightweight tags.

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
```diff
@@ -0,0 +1,162 @@
...
+    /* Instance under test. */
+    private GitClient gitClient;
+
+    /* Lower level "porcelain" API when gitClient can't do something. */
+    private org.eclipse.jgit.api.Git gitAPI;
+
+    /* Repo garbage collector. */
+    private GC gc;
+
+    @Rule
```
MarkEWaite: As far as I can tell, this Rule is not used and could be deleted.

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public ExpectedException thrown = ExpectedException.none();
```
MarkEWaite: As far as I can tell, this Rule is not used and can be deleteed.

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
brianeray: Ah yes you did mention over in the JIRA factoring this down to a more specific test class. I'm happy to make this and all suggested changes. Should be able to turn them around on Monday.

src/test/java/org/jenkinsci/plugins/gitclient/JGitClientTest.java
brianeray: Dangit. Artifact of Spock test framework brain trying to wrap itself around JUnit and coming up short.

MarkEWaite:
While working through the code review in the debugger, I made changes that I hope further simplify the tests.  Thanks very much for your excellent work investigating and resolving the bug!

Could you review my changes in https://github.com/jenkinsci/git-client-plugin/pull/514/commits/1ca79b6e81675582e220c11d60b0bae9f13df39c and comment if you detect places where I've made a mistake?

src/test/java/org/jenkinsci/plugins/gitclient/JGitLightweightTagTest.java
```diff
@@ -0,0 +1,145 @@
...
+        packRefs();
+
+        Set<GitObject> tags = gitClient.getTags();
+
+        assertThat(
+            tags,
+            hasItem(
+                allOf(
+                    hasProperty("name", equalTo(lightweightTagName)),
+                    hasProperty("SHA1", equalTo(firstCommit))
```
MarkEWaite: @brianeray can you help me understand how this assertion operates?  The field name inside the GitObject is `sha1` but the assertion is checking for a `SHA1` property.  When I tried changing it to `sha1`, the tests failed to find the property.  Clearly I have more to learn about writing good Hamcrest assertions.

brianeray: That lowercase `sha1` is exactly how I initially coded it, and it failed the same way you experienced.

Then I scrolled down that field's accessor, which [has different capitalization past the first letter](https://github.com/jenkinsci/git-client-plugin/blob/0ef4d20d057714bc5da618793b1c0c1c7bc9258e/src/main/java/hudson/plugins/git/GitObject.java#L38-L40). IIRC this particular Hamcrest matcher operates against the Java bean API, and thus goes through the getter instead straight at the field.

So it's sort of an accident of mismatched letter case in the bean here rather than idiosyncracy of Hamcrest. I think.

PR #516, opened 2020-02-26
Added no-proxy hosts when using proxy in CliGitAPIImpl

Gyanesha:
## [JENKINS-61193](https://issues.jenkins-ci.org/browse/JENKINS-61192) - When using proxy also export the no_proxy hosts

I have added an ArrayList containing all the hosts which don't need any proxy. Then at the time of setting environment variables for proxy I have also set the hosts(e.g. 169.254.169.254) which will not require any proxy.

## Checklist

## Types of changes

## Further comments

If this is a relatively large or complex change, start the discussion by explaining why you chose the solution you did and what alternatives you considered.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2013,6 +2013,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
 
             if ("http".equalsIgnoreCase(url.getScheme()) || "https".equalsIgnoreCase(url.getScheme())) {
                 if (proxy != null) {
+                    List<String>listOfConfiguredNoProxyHosts = new ArrayList<String>();
+                    listOfConfiguredNoProxyHosts.add("169.254.169.254");
                     boolean shouldProxy = true;
                     for(Pattern p : proxy.getNoProxyHostPatterns()) {
                         if(p.matcher(url.getHost()).matches()) {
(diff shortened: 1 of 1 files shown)


codingtim:
You're not supposed to make the list of non proxy hosts hard coded, these should be retrieved from the configuration on the proxy object. 
`env.put("no_proxy", getNoProxyHosts());`
with implemeation of method: 
```
private String getNoProxyHosts() {
   String noProxyHost = proxy.noProxyHost;
   List<String> noProxyHosts = Lists.newArrayList(Arrays.asList(noProxyHost.split("[ \t\n,|]+")));
   return String.join(",", noProxyHosts);
}
```
The proxy object is of type hudson.ProxyConfiguration. This class already has some utility methods to retrieve the no proxy hosts as a list of Patterns but that is not usable here.

Gyanesha:
Thanks @codingtim for the help. @MarkEWaite  can you tell me whether I should make automated test for checking if the no proxy hosts have been added as no_proxy in the environment variable or something else.

Gyanesha:
How will I check whether the NO_PROXY setting is being honoured ? Do I have to make some http request to no_proxy host ?

MarkEWaite:
> How will I check whether the NO_PROXY setting is being honored ? Do I have to make some http request to no_proxy host ?

I should have been more clear when I said "NO_PROXY is honored".  In this case, I think the reasonable effort task is to place a `NO_PROXY` into the environment and then confirm that the expected value is in the environment. 

As far as I understand it, there are many different ways to configure proxies and each has its own set of common mistakes and challenges.  We can't test more than the most basic "is it set" functionality in the test automation.

Gyanesha:
I wanted to ask one more question, like why we are not putting the proxy settings in global `environment` (EnvVar) variable, rather than local `env` (EnvVar) variable. In the function `launchCommandWithCredentials()`, we are overwriting env with itself, which in the very begining of the function represented global environment.
`env = new EnvVars(env);`
 Don't we want to reflect the changes in our setting to the global variable.

PR #521, opened 2020-03-13
[GSoC-2020] Add JMH module: micro-benchmark tests for git-fetch

rishabhBudhouliya:
### Micro-benchmarking strategy with JMH on git fetch

**_This pull request is an addition to the submitted [proposal](https://docs.google.com/document/d/1vy9tkFbhl5KX7QcjH3QbzFzbKEONw6gaLZSmNXgB06Y/edit?usp=sharing), section 3.1.3) My Benchmarking Strategy with JMH on git fetch_**

Java Micro-benchmarking Harness is a java harness which provides a fairly easy way to run benchmarks and after its integration in the [Role Strategy Plugin](https://github.com/jenkinsci/role-strategy-plugin), it is easier to write them for Jenkins plugins.

1. Creation of a test environment: Used FolderForBenchmark to create local git repositories for the lifetime of benchmark test.
2. Git operations for evaluation: git fetch
3. Results: 

**Results from the GitClientFetchVanillaBenchmark**

Screenshot 2020-03-14 at 1 55 11 AM

**Results from the GitClientFetchBenchmark**

Screenshot 2020-03-14 at 1 56 02 AM

### Observation

1. From vanilla benchmark

repo size (MiB) | git (ms/op) | jgit (ms/op) | diff in perf (ms)|
| :--- | :---         |     :---:      |          ---: |
| 0.034 | 214.55   | 972.12     |  +**757.57**   |
| 4.58 | 1341.60     | 2410.22       | +**1068.62**     |
| 93.64 | 10,313.57    | 24,213.84   | +**13,900.27**    |
| 324.26 | 97,149.34    | 290,453.4  | +**193,304.06**    |

2. From JMH benchmark (GitClientFetchBenchmark)

repo size (MiB) | git (ms/op) | jgit (ms/op) | diff in perf (ms)|
| :--- | :---         |     :---:      |          ---: |
| 0.034 | 95   | 5    |  -**90**   |
| 4.58 | 89    | 8   | -**81**     |
| 93.64 | 164   | 12,656   | +**12,492**    |
| 324.26 | 36,826  | 107,203 | +**70,377**    | 

From both of these tables, it can be said that: 
- The JMH framework is supposed to *isolate* the benchmark tests and provide reduced execution time than the vanilla benchmark. This behavior is confirmed with our results.
- From vanilla benchmark tests, it appears that **git** performs better in all cases than **jgit**. 
- JMH benchmark also shows a similar trend, but with two clear differences: 
      1. It appears that "JGit" is performing better than "git" for two cases, this seems suspicious, and we 
          have tested this further in the next table 
      2. The performance differences are lower in magnitude (as compared to the vanilla test)

The same JMH benchmark was performed with a different benchmark mode, called the **Single Shot Mode** which involves running a benchmark test without warming up the JVM (a cold start). 

Screenshot 2020-03-19 at 5 40 12 PM

Results from this chart confirm my suspicion, "jgit" performs better than "git" **only** in a sufficiently warmed up JVM under the condition that the repository size is **< 5 MiB**.   

All of this information is presented in the proposal as well.


MarkEWaite:
@rishabhBudhouliya I assume there are components of this pull request that we should bring into the git client plugin, but I have not investigated further to identify which components should be included.

Guidance?

rishabhBudhouliya:
@MarkEWaite I mistakenly didn't close this PR. This already merged [PR-553](https://github.com/jenkinsci/git-client-plugin/pull/553) has all of the components required to create and run a git operation performance benchmark in the git client plugin.

The only advantage of this PR as of now is a quick reference to show the data behind our decision of switching git implementations if asked about it.

PR #532, opened 2020-03-31
Require Jenkins 2.204.1 or later

MarkEWaite:
## Require Jenkins 2.204.1 or later

This pull request switches from requiring Jenkins 2.138.4 as minimum version to require Jenkins 2.204.1 as minimum version.  Using a modern version like 2.204.1 includes the trilead-api separation from the Jenkins core and simplifies testing.

## Checklist

## Types of changes

## Further comments

User based justification for choosing 2.204.1 as the new minimum version:

* Over 50% of the 240 000 installations of git client plugin are running git client plugin 3.0.0 or newer.  Over 40 000 git client plugin installations are using Jenkins 2.204.2
* Almost 10x more installations of git client plugin are on Jenkins 2.204.2 than any other single release

Users that are choosing to stay current with the git client plugin are also choosing to stay current with Jenkins releases.  That's a very good thing.  Special thanks to the users for choosing to update to recent releases!

The Jenkinsfile is now configured to test Jenkins 2.204.6 on Windows with Java 8 and Jenkins 2.222.1 or 2.204.6 on Linux with both Java 8 and Java 11.  It intentionally tests the most recent release of Jenkins 2.204.x while only requiring 2.204.1 so that users of 2.204.2 are not forced to update to use the git client plugin.

There are still some spotbugs warnings that will either need to be suppressed, resolved, or hidden by restoring the spotbugs settings that were used until the recent update to spotbugs 4.0.1.  Specifically, the spotbugs threshold was previously set to 'High' with the effort set to 'Low'.  The new settings invert that and set the threshold to 'Low' and the effort to 'Max'.

This change intentionally uses plugin pom 4.0-beta with the assumption that plugin pom 4.0 will be released within the next month or so.  If it is not released in that time, then the most recent released parent will be used instead.

Marching pull request for git plugin at https://github.com/jenkinsci/git-plugin/pull/862

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,8 +5,8 @@ def use_newer_jenkins = random.nextBoolean() // Use newer Jenkins on one build b
 
 // build recommended configurations
 subsetConfiguration = [ [ jdk: '8',  platform: 'windows', jenkins: null                      ],
-                        [ jdk: '8',  platform: 'linux',   jenkins: !use_newer_jenkins ? '2.176.3' : '2.164.1', javaLevel: '8' ],
-                        [ jdk: '11', platform: 'linux',   jenkins:  use_newer_jenkins ? '2.176.3' : '2.164.1', javaLevel: '8' ]
+                        [ jdk: '8',  platform: 'linux',   jenkins: !use_newer_jenkins ? '2.204.6' : '2.222.1', javaLevel: '8' ],
+                        [ jdk: '11', platform: 'linux',   jenkins:  use_newer_jenkins ? '2.204.6' : '2.222.1', javaLevel: '8' ]
                       ]
 
 buildPlugin(configurations: subsetConfiguration, failFast: false)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@
 Random random = new Random() // Randomize which Jenkins version is selected for more testing
 def use_newer_jenkins = random.nextBoolean() // Use newer Jenkins on one build but slightly older on other
 
-// build recommended configurations
+// Test plugin compatibility to recommended configurations
 subsetConfiguration = [ [ jdk: '8',  platform: 'windows', jenkins: null                      ],
                         // Compile with Java 8, test 2.204.6 or 2.222.1 depending on random use_newer_jenkins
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -5,8 +5,8 @@ def use_newer_jenkins = random.nextBoolean() // Use newer Jenkins on one build b
 
 // build recommended configurations
 subsetConfiguration = [ [ jdk: '8',  platform: 'windows', jenkins: null                      ],
-                        [ jdk: '8',  platform: 'linux',   jenkins: !use_newer_jenkins ? '2.176.3' : '2.164.1', javaLevel: '8' ],
-                        [ jdk: '11', platform: 'linux',   jenkins:  use_newer_jenkins ? '2.176.3' : '2.164.1', javaLevel: '8' ]
+                        [ jdk: '8',  platform: 'linux',   jenkins: !use_newer_jenkins ? '2.204.6' : '2.222.1', javaLevel: '8' ],
+                        [ jdk: '11', platform: 'linux',   jenkins:  use_newer_jenkins ? '2.204.6' : '2.222.1', javaLevel: '8' ]
```
fcojfernandez, line 9: I know you're only updating the jenkins version, but I'd like to understand this two lines. Wouldn't be the same as
```suggestion
                        [ jdk: '8',  platform: 'linux',   jenkins: '2.222.1', javaLevel: '8' ],
                        [ jdk: '11', platform: 'linux',   jenkins:  '2.204.6', javaLevel: '8' ]
```
?

MarkEWaite: Good question.  I will add a comment explaining that logic.

The random boolean `use_newer_jenkins` will test Java 8 on Linux with Jenkins 2.204.6 roughly 50% of the time and with Jenkins 2.222.1 roughly 50% of the time.  The negation of that random boolean will test Java 11 on Linux with the version of Jenkins which was not used in the Java 8 test.

If the random boolean is true, then Java 8  will test 2.204.6 and Java 11 will test 2.222.1.

If the random boolean is false, then Java 8 will test 2.222.1 and Java 11 will test 2.204.6..

MarkEWaite: See comments added in 105b7e7f1a18a9c68e33f3cb4e341c1d273c509c and 92f3c2de5fce739e84b1d044e0d7adc2f6824162

PR #553, opened 2020-05-28
Add JMH module: micro-benchmark tests for git-fetch

rishabhBudhouliya:
## Performance micro-benchmarking with JMH on git fetch

Java Micro-benchmarking Harness is a java harness which provides a fairly easy way to run benchmarks and after its integration in the Role Strategy Plugin, it is easier to write them for Jenkins plugins.

- Creation of a test environment: Used FolderForBenchmark to create local git repositories for the lifetime of benchmark test.
- Git operations for evaluation: git fetch
- Added an addition step in JenkinsFile to enable benchmarks on Jenkins CI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,3 +12,4 @@ subsetConfiguration = [ [ jdk: '8',  platform: 'windows', jenkins: null         
                       ]
 
 buildPlugin(configurations: subsetConfiguration, failFast: false)
+runBenchmarks('jmh-report.json')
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -12,3 +12,4 @@ subsetConfiguration = [ [ jdk: '8',  platform: 'windows', jenkins: null
                       ]
 
 buildPlugin(configurations: subsetConfiguration, failFast: false)
+runBenchmarks('jmh-report.json')
```
MarkEWaite, line 15: I had forgotten that Jenkinsfile changes are ignored for anyone that is not a committer to the repository.  It is a safety measure to prevent dangerous changes to the Jenkinsfile.

I'm evaluating the change in my CI infrastructure at home.  Will let you know how it looks in about 45 minutes.

PR #555, opened 2020-05-30
Change in Scope of state object: from thread level to benchmark level

rishabhBudhouliya:
This PR was raised to run the benchmarks with a different configuration, the static state class, which provides the setup and tear-down helper functions to each benchmark test, to be shared across the scope of a benchmark instead of a thread.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -64,3 +64,11 @@ Typical bug review tasks include:
 * Review summary and description
 * Attempt to duplicate the issue
 * Add a comment with results of the attempt to duplicate the issue
+
+[[jmh-benchmarks]]
+== Java Microbenchmark Harness
+
+Git client plugin includes link:https://openjdk.java.net/projects/code-tools/jmh/[Java Microbenchmark Harness (JMH)] benchmark tests to compare the performance of command line git and JGit.
+Run the benchmarks with the command:
+
+* `mvn -P jmh-benchmark test`
--- a/src/test/java/jmh/benchmark/GitClientFetchBenchmark.java
+++ b/src/test/java/jmh/benchmark/GitClientFetchBenchmark.java
@@ -21,7 +21,7 @@ import static org.junit.Assert.assertTrue;
  * A JMH micro-benchmark performance test, it aims to compare the performance of git-fetch using both "git" and "jgit"
  * implementations represented by CliGitAPIImpl and JGitAPIImpl respectively.
  */
-@JmhBenchmark
+//@JmhBenchmark
 public class GitClientFetchBenchmark {
 
     @State(Scope.Benchmark)
--- a/src/test/java/jmh/benchmark/GitClientFetchBenchmark.java
+++ b/src/test/java/jmh/benchmark/GitClientFetchBenchmark.java
@@ -24,7 +24,7 @@ import static org.junit.Assert.assertTrue;
 @JmhBenchmark
 public class GitClientFetchBenchmark {
 
-    @State(Scope.Thread)
+    @State(Scope.Benchmark)
     public static class JenkinsState {
 
         @Param({"git", "jgit"})
@@ -63,7 +63,7 @@ public class GitClientFetchBenchmark {
         }
 
         /**
-         * We want to create a temporary local git repository after each iteration of the benchmark, works just like
+         * We want to create a temporary local git repository after each iteration of the benchmark, similar to
          * "before" and "after" JUnit annotations.
          */
         @Setup(Level.Iteration)
--- a/src/test/java/jmh/benchmark/GitClientFetchVanillaBenchmark.java
+++ b/src/test/java/jmh/benchmark/GitClientFetchVanillaBenchmark.java
@@ -1,145 +1,145 @@
-package jmh.benchmark;
-
-import hudson.EnvVars;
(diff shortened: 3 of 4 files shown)


rishabhBudhouliya:
@MarkEWaite I can see that the pipeline is still running, and it has been almost 2 hours since the build started. 

My change is related to scope of annotation `@State` to `Benchmark` which should share the operation of cloning upstream repo to a local dir and initializing vars with all the invocations of the benchmark test.

MarkEWaite:
> @MarkEWaite I can see that the pipeline is still running, and it has been almost 2 hours since the build started.

I've started a run on the Linux computer mark-pc2.  That is the same machine that ran the last tests in several hours.  I'll let you know the results when I see them.

The last run of the master branch on ci.jenkins.io spent over 3 hours running the benchmarks.  Here's hoping that the modified scope is faster!

MarkEWaite:
I understand the issue with the benchmark duration.  The Pipeline shared library code that supports running JMH benchmarks is too reliable for its own good.  It uses a lockable resource to assure that only one benchmark is running throughout ci.jenkins.io at a time.  The PR has been blocked for hours waiting for the lock so that it can begin running.

I think we'll need to switch to run JMH bencmarks on any machine that is willing to run them.  I'm less concerned about absolute results now and more concerned to understand how the overall benchmark execution process runs.

Would you be willing to propose a change to the Jenkinsfile to switch from using `runBenchmarks('jmh-report.json')` to run maven on a node instead?

rishabhBudhouliya:
@MarkEWaite The test duration has not decreased, in my ec2-instance (core i7, 32GB ram) the test has crossed 3 hours. I need to have a better understanding of the scope variable. 

To confirm that I understand your suggestion, you're saying that I should add a step in the JenkinsFile which runs the maven command `mvn test -P jmh-benchmark`?

Also, another concern I'd like to raise is that running benchmarks from maven doesn't let jmh provide the logs which can be quite useful while exploration. For that purpose, we would have to use `java -jar benchmarks.jar <benchmark options args>`. If you want to see the overall process with a greater detail, I could provide you the standalone project jar, but this is optional.

MarkEWaite:
> To confirm that I understand your suggestion, you're saying that I should add a step in the JenkinsFile which runs the maven command `mvn test -P jmh-benchmark`?

That was the idea I had, but I think your idea in the next paragraph is better than my idea.

> Also, another concern I'd like to raise is that running benchmarks from maven doesn't let jmh provide the logs which can be quite useful while exploration. For that purpose, we would have to use `java -jar benchmarks.jar <benchmark options args>`. 

That's a much better choice of the command to run.  Does not require maven in the PATH and gives us the results to analyze.

rishabhBudhouliya:
@MarkEWaite For that option, we will need to change the pom.xml of git client plugin to create a `target/benchmarks.jar` uber jar.

`<plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-shade-plugin</artifactId>
                <version>3.2.1</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>shade</goal>
                        </goals>
                        <configuration>
                            <finalName>${uberjar.name}</finalName>
                            <transformers>
                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                                    <mainClass>org.openjdk.jmh.Main</mainClass>
                                </transformer>
                                <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
                            </transformers>
                            <filters>
                                <filter>
                                    
                                    <artifact>*:*</artifact>
                                    <excludes>
                                        <exclude>META-INF/*.SF</exclude>
                                        <exclude>META-INF/*.DSA</exclude>
                                        <exclude>META-INF/*.RSA</exclude>
                                    </excludes>
                                </filter>
                            </filters>
                        </configuration>
                    </execution>
                </executions>
            </plugin>`

MarkEWaite:
My full support that you are welcome to propose those changes in a pull request.  Thanks for doing it.

rishabhBudhouliya:
@MarkEWaite While searching for a way to integrate JMH into git client plugin in such a way that it builds the benchmark jar alongside git client plugin jar, I have realized that this integration might be a bit less maintainable and would add three addition maven plugins. 

- build-helper-maven-plugin to attach custom test sources location
- maven-compiler-plugin to run jmh-generator-annprocess annotation processor on test-compile phase 
- maven-assembly-plugin to create runnable jar with benchmarks 

In my amateur opinion, I'd rather create a separate benchmark project, let Maven package the classes from git client project into JAR, and provide it as the dependency to the benchmark project.

rishabhBudhouliya:
@MarkEWaite While we figure out a way to run benchmarks in a better way, I have added a new benchmark test which benchmarks the performance of `git init`. 

This test should run for a considerably reduced duration, we can at least validate the conventional way of running JMH benchmarks on ci.jenkins.io for git client plugin.

rishabhBudhouliya:
@MarkEWaite can we stop the all the build runs for PR-555 except the 5th one, which has the latest benchmark?

MarkEWaite:
> @MarkEWaite can we stop the all the build runs for PR-555 except the 5th one, which has the latest benchmark?

Yes.  We can even merge without waiting for the CI to complete

PR #557, opened 2020-06-07
[JENKINS-62579] Use windows system temp dir more often

MarkEWaite:
## [JENKINS-62579](https://issues.jenkins-ci.org/browse/JENKINS-62579) - Use windows system temp dir for non-Latin-1 workspaces

If the Windows temporary directory path contains Cyrillic characters, command line git fails to find the files needed for SSH and HTTP/HTTPS authentication.  Git plugin 4.0.0 and later have intentionally chosen to place credentials related temporary files near the Jenkins workspace when the path name allows it.  This is another case where the path name does not allow it.

Most readily affects Freestyle projects, since they are able to use a Cyrillic character in the job name easily.  Multibranch Pipeline jobs seem to correctly "hide" the Cyrillic characters when creating the workspace directory name.

## Checklist

## Types of changes

## Further comments

*Do not merge* this draft pull request until automated tests have been included that duplicate the problem and interactive tests have confirmed the code change resolves the issue.


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1876,6 +1876,15 @@ File createTempFile(String prefix, String suffix) throws IOException {
             if (workspaceTmp.getAbsolutePath().matches(".*[ ()|?*].*")) {
                 return createTempFileInSystemDir(prefix, suffix);
             }
+            /* Windows git fails its call to GIT_SSH and GIT_ASKPASS if the
```
timja, line 1879: would be good if there was an upstream bug report to link to

MarkEWaite, line 1879:
```suggestion
            /* JENKINS-62579 - Windows git fails its call to GIT_SSH and GIT_ASKPASS if the
```

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -1876,6 +1876,15 @@ File createTempFile(String prefix, String suffix) throws IOException {
             if (workspaceTmp.getAbsolutePath().matches(".*[ ()|?*].*")) {
                 return createTempFileInSystemDir(prefix, suffix);
             }
+            /* JENKINS-62579 - Windows git fails its call to GIT_SSH and GIT_ASKPASS if the
```
timja, line 1879: Is this because of an issue in Jenkins or in git?

MarkEWaite: I assume it is an issue with git for Windows, since git for Linux works correctly in the same conditions.  I did not spend the time to research the details of the failure.  Git client plugin versions that only used the system temporary directory were able to meet the user needs even with Cyrillic characters in the job name.  This restores that behavior and includes other non ISO-8859-1 project names in the same exclusion.

PR #560, opened 2020-06-11
Add conditional logic in JenkinsFile to run benchmarks conditionally

rishabhBudhouliya:
## Update in JenkinsFile 

This change will ensure that JMH benchmark module will run benchmarks only if it is the "master" branch or any "gsoc-" related branch.

If only one branch is kept for gsoc in the future, for example gsoc-dev, then we can remove the regex and only search for "gsoc-dev"

## Types of changes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,4 +12,8 @@ subsetConfiguration = [ [ jdk: '8',  platform: 'windows', jenkins: null         
                       ]
 
 buildPlugin(configurations: subsetConfiguration, failFast: false)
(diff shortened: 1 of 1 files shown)


rishabhBudhouliya:
@MarkEWaite Since the JenkinsFile change won't be registered in the ci.jenkins.io build, could you please test it in your own ci or merge it to master to check if the logic is correct?

MarkEWaite:
I confirmed this works in my infrastructure with two different test cases.  Merging because the CI failure was due to unreliable EC2 agents.

PR #574, opened 2020-06-22
[GSoC'20] Add redundant fetch fix performance benchmark

rishabhBudhouliya:
A benchmark to test the incremental fetch's average execution time
on the basis of number of commits, branches or tags for large sized
git remote repository (This is a part of the sensitivity analysis)

The repositories being tested: Ruby and Microsoft VSCode

@MarkEWaite Since I will be changing this benchmark code more than I'd like to do, it will be better if we merge it after a stable update.


rishabhBudhouliya:
@MarkEWaite After adding the conditional check of branches on running the benchmarks, this branch should run the benchmarks?

rishabhBudhouliya:
Screenshot 2020-06-22 at 8 33 15 PM

The benchmark results above are tested on a 
```CentOS-7 machine with openjdk version "1.8.0_252"```

File: src/test/java/jmh/benchmark/GitClientRedundantFetchBenchmark.java [hunk 1]
```diff
@@ -0,0 +1,134 @@
...
+        List<RefSpec> narrowRefSpecs = new ArrayList<>();
+        List<RefSpec> wideRefSpecs = new ArrayList<>();
+
+        final FolderForBenchmark dirForRemoteClone = new FolderForBenchmark();
+        File localRemoteDir;
+        File remoteRepoDir;
+        URIish urIish;
+
+        @Param({"https://github.com/ruby/ruby.git",
+                "https://github.com/microsoft/vscode.git"})
```
omkar-dsd, line 43: Depending on external repository might bring uncertainty in tests. My suggestion would be to create a new repo or use existing one.

fcojfernandez: https://github.com/jenkinsci/jenkins might be the closer one in similar characteristics, although lighter than those ones. We might consider on forking them, but I'm not totally sure the community would agree with having those forks just for testing purpose. Have you checked in the mailing list if there exists repositories under the jenkinsci org that match the needs?

rishabhBudhouliya: @fcojfernandez You are right, I actually modified the benchmark to test with some jenkins repositories. 

| Repository Name | .pack file size | # of commits | # of branches | 

Jenkins repo | 366.835mb | 30,051 | 31
-- | -- | -- | --
Analysis model | 42.252 mb | 3,547 | 6
git client plugin | 19.242 mb | 2,448 | 8
git plugin | 34.271mb | 3,854 | 7
parameterised trigger plugin | 4.655 mb | 538 | 2
Jenkins ec2 | 11.014 mb | 1,173 | 10
Jenkins charm | 1.884 mb | 286 | 1

rishabhBudhouliya: @omkar-dsd I agree, depending upon external repositories is not stable, especially such large repositories. 
The benchmarks are still isolated from a network transfer as we clone the upstream remote repository to a local git repository before running the benchmark for the lifetime of a trial.
This local git repository serves as the file:/// remote for our fetch calls which are then benchmarked.

omkar-dsd: If we keep a copy "for lifetime" on our end, which will keep that exact state (files, commits, tags, etc) then that's perfectly fine :). Otherwise someday they might decide to clean up the branches and or increase the branches and ultimately our test will give unexpected results.

MarkEWaite: My opinion here.  I'm willing to be persuaded otherwise, but here are my thoughts...

I don't see the benefit of "freezing" a repository (by fork or by some other technique) for purposes of this test.  I think we are trying to confirm that removing the redundant fetch is an improvement and we're trying to assess the approximate amount of that improvement based on several different environments.  I think that we should be comparing those results within a few hours of one another or within the same 24 hours.  I believe that comparison across a wider time range are less interesting and don't help us answer the question, "Which repository attributes show the greatest difference between JGit and CLI git performance?".

I don't think we should plan to use the test results for long term comparisons of performance between plugin versions.  Too many things change across the weeks or months of plugin development environments.

@rishabhBudhouliya chose a very interesting initial machine for the [benchmark test](https://github.com/jenkinsci/git-client-plugin/pull/574#issuecomment-647576651) and may have discovered another important attribute in the process.  He used CentOS 7.  CentOS 7 delivers command line git 1.8.  Some features of the git plugin (like sparse checkout) do not work as well on CLI git 1.8 as they do on later versions.  It may also be that CLI git 1.8 performs worse than CLI git 2.27.0 (the most recent release).

rishabhBudhouliya:
The latest [commit](https://github.com/jenkinsci/git-client-plugin/pull/574/commits/1af9fb42769f59203bac3e2709f106b84efbd1ca) includes a new benchmark with git ls-remote.

I will be sharing the results of that benchmark tomorrow.

src/test/java/jmh/benchmark/GitClientRedundantFetchBenchmark.java
omkar-dsd: @MarkEWaite I totally agree with your point, and if this benchmark test is for a temporary purpose, then yes, let's go ahead with this. Good to go :)

rishabhBudhouliya:
@MarkEWaite I have fixed the pattern matching in [cd5d41b](https://github.com/jenkinsci/git-client-plugin/pull/574/commits/cd5d41b4a79452b77a63d102df775df6f95d05e4). 

The difference between `==~` and `=~` is that: 
- `==~` returns a boolean and performs a strict match against the specified regular expression. 
- `=~` creates and returns a java.util.regex.Matcher instance.

PR #578, opened 2020-07-06
[JENKINS-62767] Need to bundle org.eclipse.jgit.ssh.jsch

jglick:
[JENKINS-62767](https://issues.jenkins-ci.org/browse/JENKINS-62767) - need to bundle org.eclipse.jgit.ssh.jsch

Amends #561 to handle https://github.com/eclipse/jgit/commit/8d2d683655e2de17cf465fa46af10e0e56b3aaed.

Not tested in any way, except verifying that the list of bundle JARs now looks saner and includes the missing class.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,10 +88,6 @@
       <artifactId>org.eclipse.jgit</artifactId>
       <version>${jgit.version}</version>
       <exclusions>
-        <exclusion><!-- Provided by JSch plugin -->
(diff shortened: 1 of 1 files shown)


jglick:
Was hoping that a passing build would publish incrementals for people to try, but @MarkEWaite #559 broke incremental artifact publishing as per logic in https://github.com/jenkins-infra/pipeline-library/pull/46. Keep it simple…

PR #594, opened 2020-08-04
Add a new command: UnsupportedCommand to communicate compatibility of JGit

rishabhBudhouliya:
This command serves as a way to communicate to the git plugin which git features
are not supported by JGit.

- The reason for not implementing it as a GitCommand implementation is to separate
the need of a `JGitAPIImpl` or `CliGitAPIImpl` to check if a certain additional behaviour
of git is supported or not.
- This class can be instantiated with no need of any information, it can be decorated by a
GitSCMExtension to fill in the features which are not supported.

## How to use it in the Git Plugin? 

- Step 1: Define a function within the GitSCMExtension to be implemented by extensions which are not supported by JGit.
ex: ```public void determineSupportForJGit(GitSCM scm, UnsupportedCommand unsupportedCommand) {
    }```
- Step 2: Let's take GitLFSPull as an example, implement a function called: 
```@Override
    public void determineSupportForJGit(GitSCM scm, UnsupportedCommand cmd) {
        List<RemoteConfig> repos = scm.getRepositories();
        cmd.lfsRemote(repos.get(0).getName()); 
}
```
- Step 3: Use the collected information to determine if JGit can be supported or not using the UnsupportedCommand only.

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/UnsupportedCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/UnsupportedCommand.java
@@ -1,0 +1,110 @@
+package org.jenkinsci.plugins.gitclient;
+
+import com.cloudbees.plugins.credentials.common.StandardCredentials;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * A command to convey unsupported features (Currently, implemented for JGit)
+ *
+ * All of the operations listed below are not implemented in JGit currently.
+ */
+public class UnsupportedCommand {
+
+    private boolean useJGit = true;
+
+    // From CheckoutCommand
+    public UnsupportedCommand sparseCheckoutPaths(List<String> sparseCheckoutPaths) {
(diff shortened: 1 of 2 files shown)


rishabhBudhouliya:
@MarkEWaite Many thanks for writing the test cases, I was working on the git plugin PRs and couldn't add those in time.

While the automated cases written should be more than enough to cover tests functionality wise, I believe we may have to add more test cases within the git plugin to check if the UnsupportedCommand is able to determine support after being used by various extensions.

MarkEWaite:
@rishabhBudhouliya I saw that the message associated with the first commit in this pull request was missing the blank line after the first line.  That causes the `log` output of command line git to be "noisy".  I modified the commit message to add that blank line so that the log output would be more tidy.  I also added javadoc comments for all the public methods of the UnsupportedCommand class so that we don't have javadoc warnings added to the build.

I believe this is ready to merge once the CI job passes.

PR #598, opened 2020-08-11
[JENKINS-63307] Allow zOS encoding of user name, password, and passphrase

MarkEWaite:
## [JENKINS-63307](https://issues.jenkins-ci.org/browse/JENKINS-63307) Allow zOS encoding of user name, password, and passphrase

Changes the character set definition properties.  Properties are now:

* `org.jenkinsci.plugins.gitclient.CliGitAPIImpl.user.name.file.encoding`
* `org.jenkinsci.plugins.gitclient.CliGitAPIImpl.user.passphrase.file.encoding`
* `org.jenkinsci.plugins.gitclient.CliGitAPIImpl.user.password.file.encoding`

Keeps the documentation simpler and the code simpler.

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2166,11 +2166,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
     private File createPassphraseFile(SSHUserPrivateKey sshUser) throws IOException {
-        String charset = "UTF-8";
-        if (isZos() && System.getProperty(CliGitAPIImpl.class.getName() + ".credentials.file.encoding") != null) {
-            charset = Charset.forName(System.getProperty(CliGitAPIImpl.class.getName() + ".credentials.file.encoding")).toString();
-	    listener.getLogger().println("Using passphrase charset '" + charset + "'");
-        }
+        String charset = computeCredentialFileCharset("passphrase", "UTF-8");
         File passphraseFile = createTempFile("phrase", ".txt");
         try (PrintWriter w = new PrintWriter(passphraseFile, charset)) {
             w.println(Secret.toString(sshUser.getPassphrase()));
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2179,26 +2175,33 @@ private File createPassphraseFile(SSHUserPrivateKey sshUser) throws IOException
...
+        try (PrintWriter w = new PrintWriter(usernameFile, charset)) {
             w.println(userPass.getUsername());
         }
         return usernameFile;
     }
 
     private File createPasswordFile(StandardUsernamePasswordCredentials userPass) throws IOException {
-        String charset = "UTF-8";
-        if (isZos() && System.getProperty(CliGitAPIImpl.class.getName() + ".password.file.encoding") != null) {
-            charset = Charset.forName(System.getProperty(CliGitAPIImpl.class.getName() + ".credentials.file.encoding")).toString();
```
rsbeckerca, line 2192: ".credentials.file.encoding" at line 2192 should be ".password.file.encoding"

dennis-behm:
Hi Mark,
I don't see a scenario, that different encodings needs to be set when running the agent on z/OS.  We could further simplify and just use a single property like `org.jenkinsci.plugins.gitclient.CliGitAPIImpl.user.zos.credential.file.encoding` which is used in all the three scenarios.
Randall, what are your thoughts on this?

rsbeckerca:
> 
> 
> Hi Mark,
> I don't see a scenario, that different encodings needs to be set when running the agent on z/OS. We could further simplify and just use a single property like `org.jenkinsci.plugins.gitclient.CliGitAPIImpl.user.zos.credential.file.encoding` which is used in all the three scenarios.
> Randall, what are your thoughts on this?

The reason there are separate encodings come down to how the IBM Dependency Based Build mechanism has been used with Jenkins. There are a set of scripts (of various revisions) that selectively change the encoding of each of the three elements (SSH passphrase, SSL user, SSL password). Personally, I would rather see the scripts go away completely, but until they are gone, z/OS Git Client users have a mixed back of encoding strategies for each element. Ideally this should be controlled by one property, but I do not like the breakage possibilities. I'm fairly new to Jenkins on z/OS so cannot speak to how all customers have done this, but I have gone through the script versions and am concerned that we do need the granularity to control each one separately. I would rather err on the side of prudent.

PR #600, opened 2020-08-14
[JENKINS-63307] Honor property value in zOS credential files

MarkEWaite:
## [JENKINS-63307](https://issues.jenkins-ci.org/browse/JENKINS-63307) - Honor property value in zOS credential files

Use property value rather than property name for the character set.

## Checklist

## Types of changes

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java
@@ -27,7 +27,6 @@ public class CliGitAPIImplTest extends GitAPITestCase {
     private void setCliGitDefaults() throws Exception {
         if (!cliGitDefaultsSet) {
             CliGitCommand gitCmd = new CliGitCommand(null);
-            gitCmd.setDefaults();
         }
         cliGitDefaultsSet = true;
     }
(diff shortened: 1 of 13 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java
```diff
@@ -27,7 +27,6 @@ protected boolean hasWorkingGetRemoteSymbolicReferences() {
     private void setCliGitDefaults() throws Exception {
         if (!cliGitDefaultsSet) {
             CliGitCommand gitCmd = new CliGitCommand(null);
-            gitCmd.setDefaults();
```
jglick, line 30: What then is the purpose of the whole `setCliGitDefaults` method?

MarkEWaite: I believe that it is trying to set the global defaults required of the git installation (must have a user defined and an email address defined otherwise it fails commits with a message that user and email address must be defined).  If I remember correctly, it is trying to set them only once per JVM to reduce failures where multiple JVM's running tests concurrently (`-forkCount=1C`) will collide with each other and cause failures.  However, that's just from memory, not a high precision idea.  I'll need to look at the git history of the file and the comments on the commits before I'm confident of the reasons.  I'm on vacation with my family currently and won't be back until the end of this week.  I can look at it then or other maintainers can investigate if it is needed sooner.

jglick: > it is trying to set the global defaults required of the git installation

That is what this method _was_ doing, until this PR deleted the line that actually did something (AFAICT). Thus in #920 I just delete the now-useless method.

PR #573, opened 2020-06-22
Handle auto-installing PortableGit - return git.exe path when getHome() is a directory and getHome()/bin/git.exe exists

Spitfire1900:
## [JENKINS-xxxxx](https://issues.jenkins-ci.org/browse/JENKINS-xxxxx) - summarize pull request in one line

Describe the big picture of your changes here to explain to the maintainers why this pull request should be accepted.
If it fixes a bug or resolves a feature request, include a link to the issue.

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc) doc
- [ ] I have referenced the Jira issue related to my changes in one or more commit messages
- [ ] I have added tests that verify my changes
- [ ] Unit tests pass locally with my changes
- [ ] I have added documentation as necessary
- [ ] No Javadoc warnings were introduced with my changes
- [ ] No spotbugs warnings were introduced with my changes
- [x] I have interactively tested my changes

## Types of changes

What types of changes does your code introduce? _Put an `x` in the boxes that apply_

- [ ] Infrastructure change (non-breaking change which updates dependencies or improves infrastructure)
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Further comments

We have multiple nodes that run on Windows and we'd been manually copying unpacked PortableGit installs to them in order to use it. While we've switched to using auto-installers for much of our tools this was impossible for Git because of errors like the following (with a little bit of manual internal printing to loggers and stderr):

```
Started by user Kyle Gottfried
Checking out git https://github.com/Spitfire1900/jenkins-test.git into C:\Jenkins-PortableGit\Jenkinshome\workspace\Test@script to read Jenkinsfile
Unpacking file:///C:/Jenkins-PortableGit/PortableGit.zip to C:\Jenkins-PortableGit\Jenkinshome\tools\git.exe on Jenkins
[WARNING] using C:\Jenkins-PortableGit\Jenkinshome\workspace\Test@script as workspace, C:\Jenkins-PortableGit\Jenkinshome\tools\git.exe as git exe

java.lang.Throwable
[...]

PR #605, opened 2020-08-26
document Installing PortableGit Automatically

Spitfire1900:
## Document installing PortableGit automatically

#573 was originally opened to handle auto-installing PortableGit, however auto-installing logic was already implemented but not documented. This includes said documentation.

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc) doc
- [X] I have added documentation as necessary
- [X] No Javadoc warnings were introduced with my changes
- [X] No spotbugs warnings were introduced with my changes

## Types of changes

- [X] Other: documentation change which documents an unmentioned feature

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -71,7 +71,17 @@ JGit with Apache HTTP Client continues to delivered to assure compatibility.
 
 [#installing-portablegit-automatically]
 === Installing PortableGit Automatically
-The git client plugin supports installing PortableGit automatically from a zip or tar.gz file. First follow PortableGit's installation instructions and then repack so that the git executable is at `<zipname>.zip\bin\git.exe`. Leave Path to Git executable as git.exe and specify `bin\git.exe` for the Subdirectory of extracted archive. PortableGit will be installed to `$JENKINS_HOME\tools\git.exe` and the fully qualified path to the git exe will be `$JENKINS_HOME\tools\git.exe\bin\git.exe`
+The git client plugin can install link:https://git-scm.com/download/win[Git for Windows Portable] automatically from a zip file.
+
+* Download and install link:https://github.com/git-for-windows/git/releases/[Git for Windows Portable] from its 7z.exe file
+* Create a zip file of the installation as `PortableGit-a.bb.c.zip`
+* Upload that zip file to an HTTP server
+* Set the `Download URL for binary archive` as the URL of the uploaded zip file
+* Leave `Path to Git` executable as `git`
+* Specify `PortableGit-a.bb.c\bin\git.exe` for the `Subdirectory of extracted archive
+
+Git for Windows Portable will be installed on each agent in `tools\git\PortableGit-a.bb.c`.
+The path to the git executable will be `tools\git\PortableGit-a.bb.c\bin\git.exe`.
 
 [#windows-credentials-manager]
 == Windows Credentials Manager
--- a/README.adoc
+++ b/README.adoc
@@ -73,12 +73,13 @@ JGit with Apache HTTP Client continues to delivered to assure compatibility.
 === Installing PortableGit Automatically
 The git client plugin can install link:https://git-scm.com/download/win[Git for Windows Portable] automatically from a zip file.
 
-* Download and install link:https://github.com/git-for-windows/git/releases/[Git for Windows Portable] from its 7z.exe file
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -69,6 +69,10 @@ A workaround was implemented to provide JGit but use Apache HTTP client for auth
 The issue in JGit has now been resolved and delivered in git client plugin releases.
 JGit with Apache HTTP Client continues to delivered to assure compatibility.
 
+[#installing-portablegit-automatically]
+=== Installing PortableGit Automatically
+The git client plugin supports installing PortableGit automatically from a zip or tar.gz file. First follow PortableGit's installation instructions and then repack so that the git executable is at `<zipname>.zip\bin\git.exe`. Leave Path to Git executable as git.exe and specify `bin\git.exe` for the Subdirectory of extracted archive. PortableGit will be installed to `$JENKINS_HOME\tools\git.exe` and the fully qualified path to the git exe will be `$JENKINS_HOME\tools\git.exe\bin\git.exe`
```
MarkEWaite, line 74: Thanks for the pull request!  This file prefers sentence per line formatting so that code review is easier.  I still need to test the instructions before we merge this change.

It would be clearer (at least to me) if the archive that is created were named "git-2.28.0" rather than "git.exe".  Having a directory named "git.exe" would tend to confuse me.  Is there something I'm not understanding which makes it critical that the directory must be named "git.exe" rather than (for example) "git-2.28.0"?
```suggestion
The git client plugin supports installing PortableGit automatically from a zip or tar.gz file.
First follow PortableGit's installation instructions and then repack so that the git executable is at `<zipname>.zip\bin\git.exe`.
Leave Path to Git executable as git.exe and specify `bin\git.exe` for the Subdirectory of extracted archive.
PortableGit will be installed to `$JENKINS_HOME\tools\git.exe` and the fully qualified path to the git exe will be `$JENKINS_HOME\tools\git.exe\bin\git.exe`
```

README.adoc
```diff
@@ -78,7 +78,7 @@ The git client plugin can install link:https://git-scm.com/download/win[Git for
 * Upload that zip file to an HTTP server
 * Set the `Download URL for binary archive` as the URL of the uploaded zip file
 * Leave `Path to Git` executable as `git`
-* Specify `PortableGit-a.bb.c\bin\git.exe` for the `Subdirectory of extracted archive
+* Specify `PortableGit-a.bb.c\bin\git.exe` for the `Subdirectory of extracted archive`.  This points to the git.exe in the archive relative to the root of the archive.
```
MarkEWaite, line 81: Thanks very much for the clarification and for your patience with my adaptations!

I think we should also include a suggestion in one of the earlier steps about the assumed contents of the zip file created by the user.
```suggestion
* Specify `PortableGit-a.bb.c\bin\git.exe` for the `Subdirectory of extracted archive`.
  This points to the git.exe in the archive relative to the root of the archive.
  This assumes the `bin` directory is stored in the archive as `PortableGit-a.bb.c\bin`
```

README.adoc
```diff
@@ -69,6 +69,20 @@ A workaround was implemented to provide JGit but use Apache HTTP client for auth
 The issue in JGit has now been resolved and delivered in git client plugin releases.
 JGit with Apache HTTP Client continues to delivered to assure compatibility.
 
+[#installing-portablegit-automatically]
+=== Installing PortableGit Automatically
+The git client plugin can install link:https://git-scm.com/download/win[Git for Windows Portable] automatically from a zip file.
+
+* Download and install link:https://github.com/git-for-windows/git/releases/[Git for Windows Portable] from its 7z.exe file
+* Create a zip file of the installation as `PortableGit-a.bb.c.zip`
```
MarkEWaite, line 77: Proposed additional comment to assure the zip file contains the expected layout.

```suggestion
* Create a zip file of the installation as `PortableGit-a.bb.c.zip`.
  The root of the zip file should contain a directory named `PortableGit-a.bb.c` so that the entire git installation is inside a single directory in the zip file
```

README.adoc
Spitfire1900: The first additional line was added, the second clarifying assumptions about the 'bin' directory is likely not needed.

Spitfire1900:
Thank you for merging 😸

PR #615, opened 2020-09-21
[JENKINS-63727] Add support to convey compatibility info related to JGit and Git Publisher

rishabhBudhouliya:
## [JENKINS-63727](https://issues.jenkins-ci.org/browse/JENKINS-63727) - Add support to convey compatibility info related to JGit and Git Publisher

The UnsupportedCommand now features a support to convey if JGit should be suggested
as a viable implementation in the scenario where Git Publisher is added as a post build
action.

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/UnsupportedCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/UnsupportedCommand.java
@@ -179,7 +179,9 @@ public class UnsupportedCommand {
     }
 
     /**
-     * JGit doesn't support Git Publisher
+     * JGit doesn't support Git Publisher.
+     * @param isEnabled if true, then git publisher post-build action is enabled in this context
+     * @return this for chaining
      */
     public UnsupportedCommand gitPublisher(boolean isEnabled) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/UnsupportedCommand.java
```diff
@@ -178,6 +178,16 @@ public UnsupportedCommand useBranch(String submodule, String branchname) {
         return this;
     }
 
+    /**
+     * JGit doesn't support Git Publisher
```
MarkEWaite, line 182:
```suggestion
     * JGit doesn't support Git Publisher.
     * @param isEnabled if true, then git publisher post-build action is enabled in this context
     * @return this for chaining
```

MarkEWaite (changes requested):
Needs Javadoc to avoid a pair of Javadoc warnings

MarkEWaite (approved):

rishabhBudhouliya:
@MarkEWaite Apologies for not being able to add your suggestions in due time, many thanks for doing so!

PR #619, opened 2020-09-28
Skip JMH benchmarks based on commit message content

MarkEWaite:
## Skip JMH benchmarks based on commit message content

The Jenkinsfile already requires a commit message that includes the word "benchmark" or "Benchmark".  This change makes the unit test have the same condition.  JMH benchmarks will be run in CI if and only if the word "[Bb]enchmark" is included in one of the last 3 messages.

## Checklist

## Types of changes


MarkEWaite:
Closing this pull request because it needs more work before it is ready to submit.  The change should:

* Allow JMH reports to be run by users on-demand
* Only run JMH reports on ci.jenkins.io when a specific flag condition is satisfied

PR #635, opened 2020-10-27
Add "--first-parent" to "git log" for Jenkins

drtuxwang:
We have been patching and building git-client with these changes for a few years and used them on our Jenkin servers.

This helps with git polling as the history is cleaner. For example a insertion followed by deletion does not trigger a build. We would like to contribute the fix to upstrem:

git log --first-parent --full-history --no-abbrev --format=raw -M -m --raw <previously-built-commit>..<feature-branch-merge-into-CI>

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc) doc

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1235,7 +1235,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     /** {@inheritDoc} */
     @Override
     public List<String> showRevision(ObjectId from, ObjectId to, Boolean useRawOutput) throws GitException, InterruptedException {
-        ArgumentListBuilder args = new ArgumentListBuilder("log", "--full-history", "--no-abbrev", "--format=raw", "-M", "-m");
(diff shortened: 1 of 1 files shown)


MarkEWaite (changes requested):
I generally won't accept a pull request that changes default behavior, has no new automated tests, and does not update the documentation.

You need to

* Explain in detail the strengths and weaknesses of a change to default behavior, including why the change in default behavior will not critically impact a relevant portion of the 250 000 installations of the Jenkins git client plugin.  Specifically will need to explain why a test assertion should be changed and how that test assertion affects actual plugin use
* If the change in default behavior will impact a relevant portion of the 250 000 installations of the git client plugin then provide a way that administrators can choose to enable or disable that modification
* Provide automated tests that show the condition before your change and explain why that condition was incorrect and how the new code resolves the issue
* Update the documentation (README.adoc) to describe the change and its impact on users

MarkEWaite:
Closing the pull request due to no response for 5 days and my objections to the proposed change.

drtuxwang:
This issue was reported in 2014:
https://issues.jenkins-ci.org/browse/JENKINS-23606

We will use a local patch on our Jenkins. Thanks.

PR #643, opened 2020-12-06
[JENKINS-25913] Add option to keep ignored files

gzoom13:
## [JENKINS-25913](https://issues.jenkins-ci.org/browse/JENKINS-25913) - Add option to keep ignored files

Add new parametrized clean method in GitClient and implementations. This is required to preserver artifacts between builds.

## Checklist

## Types of changes


MarkEWaite (commented):
Thanks for the pull request.

Optional suggestions to make the new javadoc comments better than the existing javadoc comments.

I've been hesitant to add this type of additional functionality to the git client plugin because I believe this type of cleaning is better done from Jenkins Pipeline steps using `sh`, `bat`, or `powershell` steps that give the user direct control of the complete command used to perform the cleaning.

Can you confirm that you need this because you are working with Freestyle or Matrix jobs and are not ready to move those jobs to PIpeline jobs?

gzoom13:
Hello @MarkEWaite. We need this because in our organization we do not have an access to command line utilities in centralized Jenkins. So it was planned to implement clean functionality preserving build artifacts in git plugin. So we need this change it git client as foundation for it. Even if we will build our custom plugin we would like to use this git client.

Aurel:
We'd also love this change to make it in at some point. We're using complicated multi-jobs and haven't had the opportunity to move to pipeline yet; but we'd love this to make our job management easier. We use a common workspace between runs of our jobs because if we wipe all of our ignored files, we'd have a 4+ hour import time for each run (Unity has to reimport all assets).

Once this makes it in, I'd love to see a similar change made to the [Git SCM Plugin](https://github.com/jenkinsci/git-plugin) to expose these parameters to the CleanBeforeCheckout and CleanAfterCheckout traits/extensions on the SCM Step. 

This would prevent us from having to run those commands separately in a shell step, which would clean up our job readability.

nfalco79:
@MarkEWaite you had close JENKINS-25913 time ago. You can close at this point also the pull request.
Anyway I fall into this because also our company we have the same requirement. We absolutely have to not delete untracked files like the .m2 or npm cache folder that is relocate to the workspace folder. Those kind of configuration are provided by plugin to avoid collision between concurrent build on the same node. Those plugin configuration for example decrease build time from 3:09 minute for an empty organisation pom.xml to 3 seconds (after the first build).

We can't neither implement a custom trait with a git reset only command because the new interface GitClient does not expose the reset method like IGitAPI and GitAPI did. So ... please, consider at least to give us the ability to execute the reset command.

In the JIRA issue you replied:
> I don't plan to implement this because Jenkins Pipeline allows the user to perform those operations from an sh step or a bat step after checkout.

This is true also for a git clean but you had implement as a trait

MarkEWaite:
> in case of switch to JGit implementation the git command is not available to the workspace to the shell bat step

That's correct.  Command line git is the canonical implementation of the git plugin and git client plugin.  The JGit implementation in the the git client plugin lacks several features that are provided by command line git.  The JGit implementation is sufficient for many uses, but is not well suited for more complicated use cases (like submodules) or for larger repositories.  A Google Summer of Code project found that JGit performs significantly worse than command line git for large repositories.

PR #673, opened 2021-02-22
[JENKINS-64913] Label private key files on SELinux

jimklimov:
## [JENKINS-64913](https://issues.jenkins-ci.org/browse/JENKINS-64913) - SSH on SELinux enforced systems does not trust unlabeled key files

As detailed in the JIRA ticket, on modern Linux systems which do activate SELinux used as Jenkins controller or build agents, it refuses to check out over git+ssh in setups that work otherwise when SELinux is not enforced. This PR explores active labeling of the temporary private key file so the ssh client would not refuse to read it on such systems.

At least initially, no unit tests are added because hitting this situation requires that a testing system is a modern Linux where `root` (human admin or their distro maintainer) ran `setenforce 1` and possibly other configuration activities.

The trigger for added logic is existence of `/usr/bin/chcon` (hardcoded at the moment, per common Linux distros I could look at), and it should be an inexpensive operation - useless but harmless if SELinux is not currently active on the box (which is something that can be toggled at run-time externally to Jenkins).

## Checklist

## Types of changes

## Further comments

The change is not very big, but rather complicated research led to it. It is detailed in the [JIRA ticket](https://issues.jenkins-ci.org/browse/JENKINS-64913).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2184,7 +2184,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             w.println("cat " + unixArgEncodeFileName(passphrase.getAbsolutePath()));
         }
         ssh.setExecutable(true, true);
-        fixSELinuxLabel(ssh, "ssh_exec_t");
+        // fixSELinuxLabel(ssh, "ssh_exec_t");
         return ssh;
     }
 
@@ -2209,7 +2209,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             w.println("esac");
         }
         askpass.setExecutable(true, true);
-        fixSELinuxLabel(askpass, "ssh_exec_t");
+        // fixSELinuxLabel(askpass, "ssh_exec_t");
         return askpass;
     }
 
@@ -2394,17 +2394,18 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             w.println("ssh -i \"" + key.getAbsolutePath() + "\" -l \"" + user + "\" -o StrictHostKeyChecking=no \"$@\"");
         }
         ssh.setExecutable(true, true);
-        fixSELinuxLabel(ssh, "ssh_exec_t");
         //JENKINS-48258 git client plugin occasionally fails with "text file busy" error
         //The following creates a copy of the generated file and deletes the original
         //In case of a failure return the original and delete the copy
         String fromLocation = ssh.toString();
         String toLocation = ssh_copy.toString();
         //Copying ssh file
+        // fixSELinuxLabel(ssh, "ssh_exec_t");
         try {
             new ProcessBuilder("cp", fromLocation, toLocation).start().waitFor();
             isCopied = true;
             ssh_copy.setExecutable(true,true);
+            // fixSELinuxLabel(ssh_copy, "ssh_exec_t");
             //Deleting original file
             deleteTempFile(ssh);
         }
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2394,7 +2394,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
             w.println("ssh -i \"" + key.getAbsolutePath() + "\" -l \"" + user + "\" -o StrictHostKeyChecking=no \"$@\"");
         }
         ssh.setExecutable(true, true);
-        fixSELinuxLabel(key, "ssh_exec_t");
+        fixSELinuxLabel(ssh, "ssh_exec_t");
         //JENKINS-48258 git client plugin occasionally fails with "text file busy" error
         //The following creates a copy of the generated file and deletes the original
         //In case of a failure return the original and delete the copy
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2064,6 +2064,49 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     @SuppressFBWarnings(value = "DMI_HARDCODED_ABSOLUTE_FILENAME",
         justification = "Path operations below intentionally use absolute '/usr/bin/chcon' at this time (as delivered in relevant popular Linux distros)"
         )
+    private void fixSELinuxLabel(File key, String label) {
+        if (!launcher.isUnix()) return;
+        if (Files.isExecutable(Paths.get("/usr/bin/chcon"))) {
+            // JENKINS-64913: SELinux Enforced forbids SSH client to read
+            // untrusted key files. Here we assume a SELinux capable system
+            // if the tool exists, so label the key file for SSH to use.
+            // TOTHINK: Should this be further riddled with sanity checks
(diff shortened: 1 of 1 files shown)


jimklimov:
Solution as currently is was verified to fix the situation for private keys stored without a passphrase in Jenkins Credentials. It still fails for keys with a passphrase:

````
Command "git..." returned status code 128:
stderr: ssh_askpass: exec(/home/jenkins-worker/jenkins/workspace/jobname/subrepo@tmp/jenkins-gitclient-pass123123.sh): Permission denied
````

Probably that needs some other permission label...

**Comment by jimklimov** [comment]
Getting the executable labels like `ssh_exec_t` or `bin_t` proved much harder: `chcon` says "Permission denied" to those attempts for the running Jenkins agent, however same commands work (and do impact executability of ssh-askpass script wrapper) in interactive shell. But that has a different context (staff_u vs user_u), if that matters...

**Comment by jimklimov** [comment]
That last CI fault was unrelated to code:
````
Caused by: java.net.ConnectException: Connect to github.com:443 [github.com/140.82.112.3] failed: Connection timed out (Connection timed out) github.com
````

**Comment by jimklimov** [comment]
Regarding the executable context, had a fruitful discussion on #selinux freenode channel with "grift", where it was confirmed that by design of security model, an agent.jar running from a logged-in SSH session gets a "user_t" context with little abilities and can not arbitrarily elevate. In that context, `ssh` (and its SELinux rule template) indeed chooses to not allow exec'ing arbitrary code editable by arbitrary users - such as "not an askpass program from a system location", and a low-privilege "user_t" has no rights to "relabel" that arbitrary code in random location to allow ssh to run it.

A proper way forward would be to define a SELinux template (set of roles, transitions, etc.) that could be assigned to Jenkins JVMs, automatically by SELinux after the sysadmin registers the ruleset, and inherited by its child processes, so that the master or agent can modify contexts of files in their namespace (e.g. under "workspace" directory). Such template could be a required pre-installable (package) to make a SELinux enforcing machine usable by Jenkins or its agents with whatever means of launching they would use, so sounds like quite a productizable DRY solution.

On a side note, my Jenkins master JVM runs from a system service (autogenerated per init-script from the RPM package) and has a process context of `system_u:system_r:initrc_t:s0`; during a git-checkout (such as Jenkinsfile lightweights) it manipulates data in `/var/lib/jenkins/cache/git-*{,@tmp}` directories which are labeled as `system_u:object_r:var_lib_t:s0` and this setup apparently has no problems using the wrapper scripts and assigning new `chcon` contexts to them, and even without assigning the contexts (I guess by virtue of being a highly privileged `system_u`). Probably other service-based agents (swarm, jnlp, docker, ...) can (be configured to?..) enjoy such setup as well.

While such change can be eventually needed for a general solution, for the short problem of git checkouts it seems easier to just put temporary pre-unencrypted key files and be done with it. Or use ssh-agent (and ssh-add) which is also usable with the low privileged processes. But then there's also a "standard" askpass mode - and helper scripts - for username/password logins... maybe can break into git config to place an auth helper at that point.

**Review by MarkEWaite — APPROVED** [review]
@jimklimov this seems ready to me.  Any concerns from you that would prevent me from merging it?

**Comment by jimklimov** [comment]
Not at the moment. As noted earlier, the system I have access to does limit
execution of arbitrarily placed executables and scripts from sshd_t
context, so the fixed use-case is for unencrypted temporary key files, and
for other situations there are different possible solutions beyond the
scope of this small PR (e.g. pre-unencryption of a key which can be abother
PR in git-client, and elsewhere selinux templates for a dedicated jenkins
context allowing arbitrary execs under workspace dir).

Thinking of it, I did not run this build of the plugin on other platforms.
But the condition on /usr/bin/chcon in certain path limits the impact
(however without other checks at the moment - e.g. for Linux, or that chcon
is what we think it is - may be or not be bulletproof safe; at least in
/usr/bin instead of $PATH it is less likely a deliberately hijacked program
name), and failures to exec it are tolerated and just reported so it should
not break ability to build jobs.

**Comment by MarkEWaite** [comment]
@jimklimov I'm considering reverting this pull request due to the noise it adds to console logs for all users of the git plugin.  Since the fetch process fails without this change on SELinux, I believe that indicates very few users are running agents on SELinux.  Yet, all users will have entries in their build logs like this:

```
15:57:33   > git /usr/bin/chcon --type=ssh_home_t /tmp/jenkins-gitclient-ssh2105122448559165972.key
15:57:33  [WARNING] Failed (1) performing chcon helper command: git /usr/bin/chcon --type=ssh_home_t /tmp/jenkins-gitclient-ssh2105122448559165972.key:
15:57:33  === STDERR:
15:57:33  /usr/bin/chcon: can't apply partial context to unlabeled file '/tmp/jenkins-gitclient-ssh2105122448559165972.key'
15:57:33  
15:57:33  ====
15:57:33  IMPACT: if SELinux is enabled, access to temporary key file may be denied for git+ssh below
```

and entries like this:

```
15:57:33   > git /usr/bin/chcon --type=ssh_home_t /var/jenkins_home/workspace/nch-pipeline-gitea_JENKINS-35687@libs/globalPipelineLibraryMarkEWaite@tmp/jenkins-gitclient-ssh654677047516219217.key
15:57:33  [WARNING] Failed (1) performing chcon helper command: git /usr/bin/chcon --type=ssh_home_t /var/jenkins_home/workspace/nch-pipeline-gitea_JENKINS-35687@libs/globalPipelineLibraryMarkEWaite@tmp/jenkins-gitclient-ssh654677047516219217.key:
15:57:33  === STDERR:
15:57:33  /usr/bin/chcon: can't apply partial context to unlabeled file '/var/jenkins_home/workspace/nch-pipeline-gitea_JENKINS-35687@libs/globalPipelineLibraryMarkEWaite@tmp/jenkins-gitclient-ssh654677047516219217.key'
15:57:33  
15:57:33  ====
15:57:33  IMPACT: if SELinux is enabled, access to temporary key file may be denied for git+ssh below
```

That's a lot of log noise for 250 000 users when very few will benefit from the change that generates that log noise.  My apologies that I didn't comment on that log noise before merging the change.  I hadn't experimented with the change interactively and had not read the change carefully enough to assure that the build log would be well-behaved for typical users.

I'd like to release the git client plugin shortly after the March 17, 2021 release of JGit 5.11.0.  Do you have time to remove that build log noise for typical users before that planned release date?  If not, I'll revert this merge, deliver the git client plugin 3.7.0 release, then merge it again with the assumption that you or I will remove the log noise before it is delivered to users.

jimklimov:
Fair point, I too did not pay much attention to this on other systems...
sorry about that.

Tracing `getenforce` now, I see it reading a "1" (string) from
/sys/fs/selinux/enforce on that system (or "0" when enforcing is disabled -
selinux may be still enabled, but does not get into our way fatally then).
I suppose trying to read that is a cheap way to toggle this activity in the
git-client plugin and so avoid the noise on systems that have selinux but
did not enable it at the moment. Forking a `getenforce` call might be more
reliable, but more expensive in coding (check if it exists, grab stdout...)
and forking per se. I'm trying to learn if that sysfs point is "committed".
At worst, if it moves later, we won't `chcon` and the few people with
selinux and ssh keys won't be able to clone again and complain maybe.

jimklimov:
At least, this peppering with prints did collect the info :-)

Another suggested (to me, on IRC) methods to check if selinux is enabled at
all (as a subsystem) included:
* check for existence of /sys/fs/selinux
* check /proc/self/attr/current - if it exists, and contains not "kernel"
but a colon-separated context tuple, then there is something to worry about

MarkEWaite:
> At least, this peppering with prints did collect the info :-) 

Agreed wholeheartedly.  The system reporting that message is a Debian container running Jenkins 2.277.1.  I thought that SELinux was purely a CentOS / Red Hat thing, but apparently the core utilities to implement it are also included in Debian.

jimklimov:
Exploring a remedy in another PR #690.

PR #692, opened 2021-03-20
[JENKINS-38699] Limit CLI git operations to the workspace

arpoch:
[JENKINS-38699](https://issues.jenkins-ci.org/browse/JENKINS-38699) - Validate .git directory with --resolve-git-dir

As the validation of a git repository presently is checked using `git rev-parse --is-inside-work-tree` but if the .git directory is corrupted then current repository's parent directories are searched for a valid .git directory. To limit/confine the traversal, `--is-inside-work-tree` is replaced by `--resolve-git-dir` which check if path to the .git directory of the current git repository is valid.

## Checklist

## Types of changes

## Further comments

Switched from the original implementation as a breaking change to instead add a new API that allows the caller to define if parent directories should be searched for a git repository.  Retains the existing behavior for the CLI git implementation (searching parent directories if the workspace contains a `.git` directory) and the existing behavior of the JGit implementation (never searching parent directories).  Retains the previous implementation and includes a statement in the javadoc about the behavioral difference between the CLI git and JGit implementations of `hasGitRepo()`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -339,9 +339,9 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     @Override
     public boolean hasGitRepo() throws GitException, InterruptedException {
         if (hasGitRepo(".git")) {
-            // Check if this is a valid git repo with --is-inside-work-tree
+            // Check if this is a valid git repo with --resolve-git-dir
             try {
-                launchCommand("rev-parse", "--is-inside-work-tree");
+                launchCommand("rev-parse", "--resolve-git-dir",workspace.getAbsolutePath()+File.separator+".git");
             } catch (Exception ex) {
                 ex.printStackTrace(listener.error("Workspace has a .git repository, but it appears to be corrupt."));
                 return false;
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -341,7 +341,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         if (hasGitRepo(".git")) {
             // Check if this is a valid git repo with --resolve-git-dir
             try {
-                launchCommand("rev-parse", "--resolve-git-dir",workspace.getAbsolutePath()+File.separator+".git");
+                launchCommand("rev-parse", "--resolve-git-dir",
+                        workspace.getAbsolutePath() + File.separator + ".git");
             } catch (Exception ex) {
                 ex.printStackTrace(listener.error("Workspace has a .git repository, but it appears to be corrupt."));
                 return false;
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -330,7 +330,9 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     }
 
     /**
-     * Returns true if the current workspace has a git repository.
+     * Returns true if this workspace has a git repository.
+     * Also returns true if this workspace contains an empty .git directory and
+     * a parent directory has a git repository.
      *
      * @return true if this workspace has a git repository
      * @throws hudson.plugins.git.GitException if underlying git operation fails.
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -339,9 +339,9 @@ public void init() throws GitException, InterruptedException {
     @Override
     public boolean hasGitRepo() throws GitException, InterruptedException {
         if (hasGitRepo(".git")) {
-            // Check if this is a valid git repo with --is-inside-work-tree
+            // Check if this is a valid git repo with --resolve-git-dir
             try {
-                launchCommand("rev-parse", "--is-inside-work-tree");
+                launchCommand("rev-parse", "--resolve-git-dir",workspace.getAbsolutePath()+File.separator+".git");
```
jimklimov, line 344: A bit of whitespace would be good for readability and line wrappings.

Just in case (sanity check) - can we expect the new CLI option to be available everywhere? (Is it not a recent addition to Git toolkit?) Otherwise maybe should handle a failure of such lookup and retry with one that nobody complained about ;)

arpoch: As per Git 1.8.0 documentation pages the --resolve-git-dir is supported, link I am referencing from https://code.google.com/archive/p/git-core/downloads?page=9

MarkEWaite (commented):
Thanks!

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -598,6 +598,16 @@ public void testHasGitRepo() throws Exception {
...
         GitClient emptyClient = Git.with(TaskListener.NULL, new EnvVars()).in(emptyDir).using(gitImplName).getClient();
         assertFalse("Empty repo '" + emptyDir.getAbsolutePath() + "' initialized", emptyClient.hasGitRepo());
+
+        File badGitDir = tempFolder.newFolder("parentDir", "childDir", ".git");
+        assertEquals(".git", badGitDir.getName());
+        assertEquals("childDir", badGitDir.getParentFile().getName());
+        assertEquals("parentDir", badGitDir.getParentFile().getParentFile().getName());
+        gitClient.init_().workspace(badGitDir.getParentFile().getParentFile().getAbsolutePath()).execute();
+        File gitDir = gitClient.withRepository((repo, channel) -> repo.getDirectory());
+        assertTrue("Missing " + gitDir, gitDir.isDirectory());
```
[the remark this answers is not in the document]
arpoch: The `gitDir` is just a check to ensure that `.git` directory has been created in the `parentDir` directory as a result of `git.init_()`.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
arpoch: Implemented the suggested changes.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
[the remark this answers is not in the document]
arpoch: Implemented the suggested changes and created a separate test case for the issue as well.

MarkEWaite (approved):

arpoch:
Thanks @MarkEWaite for getting involved and providing a new functionality. But JENKINS-38699 issue still persists as the git-plugin currently uses the `hasGitRepo()` method. So should a PR for the issue(JENKINS-38699) be opened in git-plugin which could use the functionality implemented by this PR?

MarkEWaite:
> Thanks @MarkEWaite for getting involved and providing a new functionality. But JENKINS-38699 issue still persists as the git-plugin currently uses the `hasGitRepo()` method. So should a PR for the issue(JENKINS-38699) be opened in git-plugin which could use the functionality implemented by this PR?

Thanks @arpoch .  Yes, that's what will be needed.  I'm preparing that change now with a series of interactive and automated tests to confirm that JENKINS-38699 is resolved.  As far as I can tell from reading the git plugin code, all the cases of `hasGitRepo()` should be replaced with `hasGitRepo(false)`.  Would you be willing to look at the git plugin and see if that matches your reading as well?

The behavior of `hasGitRepo()` for CLI git was different enough from the behavior for JGit that I didn't want to risk changing either of their implementations.  I don't know which plugins depend on the JGit behavior and which depend on the CLI git behavior.  The new methods are consistent between CliGit and JGit.

arpoch:
I completely agree on replacing `hasGitRepo()` with `hasGitRepo(false)`, I will definitely look into the git-plugin code-base as well and update you further.

MarkEWaite:
@arpoch See https://github.com/jenkinsci/git-plugin/pull/1064 for the proposed change to git plugin

arpoch:
I searched for `hasGitRepo()` method usage in git-plugin source code and automated tests as well, the jenkinsci/git-plugin#1064(https://github.com/jenkinsci/git-plugin/pull/1064) changes match with my results.

PR #693, opened 2021-03-31
[JENKINS-64913] CliGitAPIImpl.java: less noise with "/usr/bin/chcon"

jimklimov:
## [JENKINS-64913](https://issues.jenkins-ci.org/browse/JENKINS-64913) - revise the SELinux handling to be less noisy in optimal case

Previous PRs for JENKINS-64913 introduced basic support for SELinux, where key files for SSH may have to be labeled, and temporary files made by git-client-plugin were originally rejected. This new feature came with informational logging, especially noisy in case of failures.

As comments to the JIRA ticket exposed, there are systems and cases where SELinux is present, but "chcon" tool fails to modify labels, but yet somehow the rest of the git+ssh tooling does not mind using the key file as it is and so in the end noise about "chcon" failure is irrelevant.

This PR explores the idea of caching failure details of helpers (starting currently with this "chcon") and only flush them to build log if the actual git operation did not succeed.

## Checklist

## Types of changes

## Further comments

This change was not yet tested in practice. It seems there should be several cases covered, not all of these I can do on my own:
* git client on systems without SELinux (perhaps not Linux at all) => should not log any of this
* git client on systems with SELinux present but inactive => may briefly log detection and evaluation of SELinux setup
* git client on systems with SELinux present and active, requiring the key file label change and successfully applying that change => should briefly log detection and evaluation of SELinux setup and the running of chcon before git; no errors to report
* git client on systems with SELinux present and active, not succeeding (e.g. having the rights) to change it, but not really requiring the key file label change  => should briefly log detection and evaluation of SELinux setup and the running of chcon before git, but if git succeeded - should not detail the failure of chcon
* git client on systems with SELinux present and active, requiring the key file label change and NOT successfully applying that change => should briefly log detection and evaluation of SELinux setup and the running of chcon before git, but if any errors happen with both git and chcon only then should chcon issues be visible - during the report of git failure, to help troubleshoot it

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -57,6 +57,7 @@ import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.PosixFilePermissions;
 import java.nio.file.attribute.UserPrincipal;
 import java.text.MessageFormat;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -68,6 +69,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


jimklimov:
Tested on the original system that had concerns about SELinux:
* when `setenforce 1` is active, `chcon` is logged as applied and no errors are reported; git checkouts succeed
* when `setenforce 0` is active, `chcon` is not even called (not logged at least); still, no errors are reported, and git checkouts succeed

Per report in the JIRA issue that originally triggered this follow-up PR, "All warnings have gone and I couldn't see any other side effects." so presumably on a system where "chcon" is called and fails but git succeeds, there is at least no big scary log about problems that do not bite us.

MarkEWaite (approved):

PR #712, opened 2021-06-08
JENKINS-28335 - GSOC-2021 Git username and password binding 

arpoch:
[JENKINS-28335](https://issues.jenkins.io/browse/JENKINS-28335) - Add Git Credentials binding for Username and Password

Please refer to the Project page for more details-
https://www.jenkins.io/projects/gsoc/2021/projects/git-credentials-binding/

## Checklist

## Types of changes

## Further Comments

Javadoc comments for classes and interfaces has to be made.


MarkEWaite:
The test failure in the CI configuration is unrelated to this change.  We may need to replace that test or rework it significantly, since it depends on a submodule default branch name of `master` and the submodule has updated its default branch name to be `main`.

MarkEWaite:
Interactive testing that I've performed has looked very good.  I've tested a single private GitHub repository with the following scenarios:

* Freestyle project with authenticated repository clone, was able to use credential binding to perform `git fetch --all`.  Without credential binding, the `git fetch --all` fails
* Declarative pipeline project with sh step
* Declarative pipeline project with PowerShell step
* Declarative pipeline project with bat step

I have configurations for many other providers and many other scenarios, but don't know if I'll have time to spend on those before I go offline.

Those passing cases have run on interesting agents that include s390x mainframe, OpenBSD amd64, Ubuntu aarch64, and Windows 10.

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindTest.java
```diff
@@ -0,0 +1,81 @@
...
+    private GitUsernamePasswordBind gitCredBind=null;
+
+    @Before
+    public void basicSetup() throws IOException {
+        //File init
+        rootDir = tempFolder.getRoot();
+        rootFilePath = new FilePath(rootDir.getAbsoluteFile());
+
+        //Credential init
+        credentials = new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, "80y7g2fjekru8h9237ywes2", "GIt Username and Password Binding Test", this.username, this.password);
```
MarkEWaite, line 61: Could you provide more context or a comment for the purpose of the magic string?  I assume that `80y7g2fjekru8h9237ywes2` is the credentials ID.  I'm accustomed to them having a different format.  A comment would help here to explain that this is a credentials ID.
```suggestion
        credentials = new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, "80y7g2fjekru8h9237ywes2", "GIt Username and Password Binding Test", this.username, this.password);
```

MarkEWaite (commented):
Initial comments from first review.  Great work @arpoch !  Interactive testing with GitHub and Bitbucket have found no issues so far.

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindTest.java
arpoch: I will use a hash function for this purpose, it will be more certain than a magic string.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Override
+    public void setKeyBindings(@Nonnull StandardCredentials credentials) {
+        credMap.put(USERNAMEKEY,((StandardUsernamePasswordCredentials) credentials).getUsername());
+        credMap.put(PASSWORDKEY,((StandardUsernamePasswordCredentials) credentials).getPassword().getPlainText());
+    }
+
+    @Override
+    public void setRunEnviornmentVariables(@Nonnull FilePath filePath, @Nonnull TaskListener listener) throws IOException, InterruptedException {
+        if(!Functions.isWindows() && getCliGitAPIInstance("git",new File(filePath.toURI()),listener,new EnvVars()
```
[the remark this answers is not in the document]
arpoch: To make this approach more reliable, I have to ensure that `GitExe` returns `git` and not `jgitapache` or `jgit`. Also should I make the `GitExe` result a criteria for this binding to work, the binding scope is limited to `git`.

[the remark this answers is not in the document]
arpoch: To make this approach more reliable, I have to ensure that `GitExe` returns `git` and not `jgitapache` or `jgit`. Also should I make the `GitExe` result a criteria for this binding to work, the binding scope is limited to `git`.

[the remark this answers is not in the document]
rishabhBudhouliya: @arpoch You need to ensure a few things here:
- each Git installation is wrapper around a Jenkins concept called [GitTool](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/hudson/plugins/git/GitTool.java)
- A git installation contains a name of installation, location of that installation and additional properties around it (you can check this out in the [Jenkins configuration page](https://plugins.jenkins.io/git-client/#enabling-jgit))
- A git installation (cli git, jgit or jgitapache) varies from a Jenkins agent(node).
- A cli git installation can be of different types depending upon the operating system (git.exe for windows)
- For your case, please check this code out: [link](https://github.com/jenkinsci/git-plugin/blob/8cb44bfbf8ffb7b3bfcd62a2bc09d6efad54bf3f/src/main/java/jenkins/plugins/git/GitToolChooser.java#L340) I wrote this to determine the available git installations and choose the appropriate one. I believe you can extend the same logic for your application

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
```diff
@@ -0,0 +1,156 @@
...
+        protected Class<StandardUsernamePasswordCredentials> type() {
+            return StandardUsernamePasswordCredentials.class;
+        }
+
+        @Override
+        public boolean requiresWorkspace() {
+            return true;
+        }
+    }
+}
```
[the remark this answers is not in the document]
rishabhBudhouliya: +1

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -0,0 +1,20 @@
...
+import java.io.File;
+import java.io.IOException;
+
+public interface GitCredentialBindings {
+
+    void setKeyBindings(@Nonnull StandardCredentials credentials);
+
+    void setRunEnviornmentVariables(@Nonnull FilePath filePath, @Nonnull TaskListener listener) throws IOException, InterruptedException;
+
+    CliGitAPIImpl getCliGitAPIInstance(String gitExe, File workspace, TaskListener listener, EnvVars environment);
```
rishabhBudhouliya, line 18: @arpoch @MarkEWaite I am not sure if we should directly call the CliGitAPIImpl. The correct usage is to call [Git](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/Git.java) and then let it decide which implementation to use (in our case it will always be CliGitAPIImpl)

rishabhBudhouliya: with that change, the contract should have a `getGitClientInstance` or `getGitClient` which seems more suitable.

rishabhBudhouliya:
@arpoch Great work on the binding, my initial review on the PR is to:
- work on the GitTool management, avoid using the gitExe string directly
- the GitCredentialBindings contract might need a minor update

I will review the code further, thank you for being patient.

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
arpoch: Great suggestion @rishabhBudhouliya. Working on that currently will try to make a commit by today. Thanks.

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -1,20 +1,22 @@
...
+import hudson.plugins.git.GitTool;
+import org.apache.commons.io.FilenameUtils;
 
 import javax.annotation.Nonnull;
-import java.io.File;
 import java.io.IOException;
 
 public interface GitCredentialBindings {
 
+    String GIT_TOOL_NAME = FilenameUtils.removeExtension(GitTool.getDefaultInstallation().getGitExe());
```
rishabhBudhouliya, line 14: I am not sure we should assume that build will be using the git default installation. The user might provide a custom git installation and we should always respect the user's choice first, if not found (or jgit/jgitapache) then fall back to the default installation.

rishabhBudhouliya: You could refer to this piece of code used to resolve user's git installation: https://github.com/jenkinsci/git-plugin/blob/8cb44bfbf8ffb7b3bfcd62a2bc09d6efad54bf3f/src/main/java/hudson/plugins/git/util/GitUtils.java#L94

arpoch: Will definitely look into that and make necessary changes.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
```diff
@@ -0,0 +1,156 @@
...
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+
+public class GitUsernamePasswordBind extends MultiBinding<StandardUsernamePasswordCredentials> implements GitCredentialBindings {
+    final static private String USERNAMEKEY =  "Git_Username";
```
justinharringa, line 30: kinda nitpicky but would be good to stick with Java and shell naming standards. If I've interpreted these correctly, these suggestions should work.
```suggestion
    final static private String USERNAME_KEY =  "GIT_USERNAME";
```

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
```diff
@@ -0,0 +1,156 @@
...
+    final static private String PASSWORDKEY =  "Git_Password";
```
justinharringa, line 31:
```suggestion
    final static private String PASSWORD_KEY =  "GIT_PASSWORD";
```

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
```diff
@@ -0,0 +1,156 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+
+public class GitUsernamePasswordBind extends MultiBinding<StandardUsernamePasswordCredentials> implements GitCredentialBindings {
```
justinharringa, line 29: Nitpick but probably should be named with "Binding" as the suffix rather than Bind
```suggestion
public class GitUsernamePasswordBinding extends MultiBinding<StandardUsernamePasswordCredentials> implements GitCredentialBindings {
```

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
```diff
@@ -0,0 +1,156 @@
...
+    final static private String USERNAMEKEY =  "Git_Username";
+    final static private String PASSWORDKEY =  "Git_Password";
+    private Map<String,String> credMap = new LinkedHashMap<>();
+
+    @DataBoundConstructor
+    public GitUsernamePasswordBind(String credentialsId) {
```
justinharringa, line 35:
```suggestion
    public GitUsernamePasswordBinding(String credentialsId) {
```

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+        @Override
+        protected FilePath write(StandardUsernamePasswordCredentials credentials, FilePath workspace)
+                throws IOException, InterruptedException {
+            FilePath gitEcho;
+            if (!Functions.isWindows()) {
+                gitEcho  = workspace.createTempFile("auth", ".sh");
+                gitEcho.write("#!/bin/sh\n" +
+                        "case $1 in\n" +
+                        "        Username*) echo " + this.userVariable + "\n" +
```
justinharringa, line 114: Should `username` and `password` perhaps be all caps or all lowercase?
```suggestion
                        "        username*) echo " + this.userVariable + "\n" +
```

justinharringa:
Good stuff! I agree with Mark/Rishabh on their comments. I only really had naming nitpicks as feedback really.

MarkEWaite (changes requested):
Some items that I learned while working with the code in a debugger today.  Not sure that all of them need action, but I suspect that some of them do.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
justinharringa: Yeah, those would work as well. Updated the suggestion to `GIT_USERNAME_KEY` but open to any of those 3 options.

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    public void test_GenerateGitScript_write() throws IOException, InterruptedException {
+        GitUsernamePasswordBind.GenerateGitScript tempGenScript = new GitUsernamePasswordBind.GenerateGitScript(this.username,this.password,credentials.getId());
+        FilePath tempSriptFile = tempGenScript.write(credentials,rootFilePath);
+        assertNotNull(tempGenScript);
+        assertTrue("Username not found",tempSriptFile.readToString().contains(this.username));
+        assertTrue("Password not found",tempSriptFile.readToString().contains(this.password));
+    }
+}
```
MarkEWaite, line 72: I was feeling like writing an automated test.  Here is a test that is largely copied from UsernamePasswordMultiBindingTest in the credentials binding plugin.  It needs more work because it relies on shell expansion of the environment variable to generate the test output.  The BatchFile and Shell builders could both be replaced with a call to an Ant builder, but I ran out of time to make that change.
```suggestion

    @Test
    public void test_FreeStyleProject() throws Exception {
        FreeStyleProject p = r.createFreeStyleProject();
        List<? extends MultiBinding<?>> bindings = Collections.<MultiBinding<?>>singletonList(new GitUsernamePasswordBind(credentials.getId()));
        SecretBuildWrapper secretWrapper = new SecretBuildWrapper(bindings);
        p.getBuildWrappersList().add(secretWrapper);
        if (isWindows()) {
            p.getBuildersList().add(new BatchFile("@echo off\necho %" + GitUsernamePasswordBind.USERNAMEKEY + "%:%" + GitUsernamePasswordBind.PASSWORDKEY + "% > auth.txt"));
        } else {
            p.getBuildersList().add(new Shell("set +x\necho \"$" + GitUsernamePasswordBind.USERNAMEKEY + ":$" + GitUsernamePasswordBind.PASSWORDKEY + "\" > auth.txt"));
        }
        r.configRoundtrip((Item) p);
        SecretBuildWrapper wrapper = p.getBuildWrappersList().get(SecretBuildWrapper.class);
        FreeStyleBuild b = r.buildAndAssertSuccess(p);
        r.assertLogNotContains(this.password, b);
        assertThat(b.getWorkspace().child("auth.txt").readToString().trim(), is(username + ':' + password));
        assertThat(new TreeSet<>(b.getSensitiveBuildVariables()).toString(), is("[" + GitUsernamePasswordBind.PASSWORDKEY + ", " + GitUsernamePasswordBind.USERNAMEKEY + "]"));
    }

    /**
     * inline ${@link hudson.Functions#isWindows()} to prevent a transient
     * remote classloader issue
     */
    private static boolean isWindows() {
        return File.pathSeparatorChar == ';';
    }

}
```

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindTest.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(Parameterized.class)
+public class GitUsernamePasswordBindTest {
+    @Parameterized.Parameters
```
MarkEWaite, line 26: Easier to diagnose in some cases if the test includes the parameter values in the output.
```suggestion
    @Parameterized.Parameters(name="User {0}: Password {1}")
```

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind.java
arpoch: The `Username` and `Password` are the part of the script, so this shouldn't  be changed, here the script is performing a switch case. Looking for the `Username` keyword on the terminal prompt.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -0,0 +1,162 @@
...
+
+import org.jenkinsci.Symbol;
+import org.jenkinsci.plugins.credentialsbinding.BindingDescriptor;
+import org.jenkinsci.plugins.credentialsbinding.MultiBinding;
+import org.jenkinsci.plugins.credentialsbinding.impl.AbstractOnDiskBinding;
+import org.jenkinsci.plugins.credentialsbinding.impl.UnbindableDir;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
```
MarkEWaite, line 24: Need to be removed.  The `javax.annotation` classes are deprecated in Jenkins.  We should use the findbugs annotations for `NonNull` instead.
```suggestion
import javax.annotation.Nonnull;
import javax.annotation.Nullable;
```

src/main/resources/org/jenkinsci/plugins/gitclient/GitSSHUserPrivateKeyBind/config.jelly
```diff
@@ -0,0 +1,26 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:st="jelly:stapler" xmlns:c="/lib/credentials">
```
MarkEWaite, line 22: This file is in an unexpected location.  I assume it may sometimes replace the credentials binding plugin config.jelly file at this location.  I think users will be surprised to have one plugin replacing the config.jelly file provided by another plugin.  Can you point me to an explanation why this is needed in the git client plugin credentials binding implementation?

src/main/resources/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBind/config.jelly
```diff
@@ -0,0 +1,26 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:st="jelly:stapler" xmlns:c="/lib/credentials">
```
MarkEWaite, line 22: I think that this jelly file needs to be placed in a directory that matches the class name of the object it is defining.  I assume this jelly file is not reachable at its current location

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindingTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @Test
+    public void test_FreeStyleProject() throws Exception {
+        FreeStyleProject prj = r.createFreeStyleProject();
+        prj.getBuildWrappersList().add(new SecretBuildWrapper(Collections.<MultiBinding<?>>
+                singletonList(new GitUsernamePasswordBinding(credentialID))));
+        if (isWindows()) {
+            prj.getBuildersList().add(new BatchFile("@echo off\necho %Git_Username%:%Git_Password% > auth.txt"));
+        } else {
+            prj.getBuildersList().add(new Shell("set +x\necho $Git_Username:$Git_Password > auth.txt"));
```
MarkEWaite, line 158: I would expect this test to fail because I thought that shell variable names were case sensitive.  Since the shell variable name is `GIT_USERNAME` it seems like `Git_Username` should resolve to the empty string.

MarkEWaite (changes requested):
Would like the javax.annotation references removed.  Their use is deprecated in Jenkins code.

See https://groups.google.com/g/jenkinsci-dev/c/uE1wwtVi1W0/m/gLxdEJmlBQAJ for more discussion of the rationale for removing those references

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindTest.java
```diff
@@ -0,0 +1,72 @@
...
+        gitCredBind = new GitUsernamePasswordBind(credentials.getId());
+    }
+
+    @Test
+    public void test_GenerateGitScript_write() throws IOException, InterruptedException {
+        GitUsernamePasswordBind.GenerateGitScript tempGenScript = new GitUsernamePasswordBind.GenerateGitScript(this.username,this.password,credentials.getId());
+        FilePath tempSriptFile = tempGenScript.write(credentials,rootFilePath);
+        assertNotNull(tempGenScript);
+        assertTrue("Username not found",tempSriptFile.readToString().contains(this.username));
+        assertTrue("Password not found",tempSriptFile.readToString().contains(this.password));
```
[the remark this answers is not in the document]
arpoch: `assertThat`, this method seems to be deprecated.
https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertThat(java.lang.String,%20T,%20org.hamcrest.Matcher)

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
arpoch: Things got messy after merging the test branch I was working on in the `gitUsernamePassword` branch. But lesson learn will be more care full next time. Making things right in the `gitUsernamePassword` branch and will make a well verified commit this time, my apologizes.

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindTest.java
arpoch: Will work on replacing the BatchFile and Shell builders with an Ant builder and update further

MarkEWaite: @arpoch I think that I've realized there is an easy way to make the test repeatable and reliable and still use the `Shell` and `Bat` steps.  No need to spend time on the Ant builder idea.  It would be very complicated, while the technique I'm envisioning feels simple.

The idea is to use shell steps like this:

```
new Shell("env | grep " + GitUsernamePasswordBind.PASSWORDKEY + "' > auth ; " +
                 "env | grep " + GitUsernamePasswordBind.USERNAMEKEY + "' >> auth")
```

That does not rely on the shell expansion or shell quoting and will still write the value of those two environment variables to the output file for later use in assertions.  I think it will significantly simplify the test and let us avoid all the complexity of using an `ant` step.

src/test/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBindTest.java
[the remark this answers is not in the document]
MarkEWaite: The `assertThat` from junit is deprecated.  The `assertThat` from hamcrest is not deprecated.

MarkEWaite:
@arpoch I pushed changes to the test that makes it less susceptible to shell expansion issues.  I'm going to sleep now, but I'm hopeful that the tests will pass so that you can create the documentation in the README and the online help in the help.html file.

We still need additional interactive testing, but it feels like you're making very good progress.

MarkEWaite:
@arapoch the most recent change introduces a findbugs warning that needs to be corrected.

> Possible null pointer dereference in org.jenkinsci.plugins.gitclient.GitCredentialBindings.gitToolName(Run, TaskListener) due to return value of called method

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -29,43 +29,47 @@
...
+                                return gitTool.getGitExe();
+                            }
+                        }
                     }
                 }
+                listener.getLogger().println("No Git Tool is configured on the current node");
             }
-        } else {
-            return null;
+            listener.getLogger().println("Is this node available...");
```
justinharringa, line 72: I'm guessing that this is just debugging information and can be removed before we merge?

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -29,43 +29,47 @@
...
+                                //Check for properties
+                                GitTool gitTool = new GitTool(tool.getName(), tool.getHome(), null);
+                                //node specific git home for git tool
+                                gitTool = gitTool.forNode(currentNode, listener);
+                                return gitTool.getGitExe();
+                            }
+                        }
                     }
                 }
+                listener.getLogger().println("No Git Tool is configured on the current node");
```
justinharringa, line 70: Would we want to keep this for merge? I could see it potentially being useful for the user but I wonder if this should be emitted in the job logs or server/admin logs.

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -2,21 +2,75 @@
...
 
 public interface GitCredentialBindings {
 
-    String GIT_TOOL_NAME = FilenameUtils.removeExtension(GitTool.getDefaultInstallation().getGitExe());
+    void setKeyBindings(@NonNull StandardCredentials credentials);
 
-    void setKeyBindings(@Nonnull StandardCredentials credentials);
-
-    void setRunEnviornmentVariables(@Nonnull FilePath filePath, @Nonnull TaskListener listener) throws IOException, InterruptedException;
+    void setRunEnviornmentVariables(@NonNull FilePath filePath, @NonNull TaskListener listener) throws IOException, InterruptedException;
```
justinharringa, line 25:
```suggestion
    void setRunEnvironmentVariables(@NonNull FilePath filePath, @NonNull TaskListener listener) throws IOException, InterruptedException;
```

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -2,21 +2,75 @@
...
+                    //It's on an agent type
+                    ToolLocationNodeProperty nodeToolLocation = currentNode.getNodeProperty(ToolLocationNodeProperty.class);
+                    //Tool location property is configured
+                    if (nodeToolLocation != null) {
+                        //Get Tool locations(could be many)
+                        List<ToolLocationNodeProperty.ToolLocation> toolList = nodeToolLocation.getLocations();
+                        //Iteration
+                        for (ToolLocationNodeProperty.ToolLocation tool : toolList) {
+                            //If 'Default' Git Tool is chosen on an agent then all the Git Tools installations added
+                            // to the master nodes will be available on the agent as well
```
justinharringa, line 58:
```suggestion
                            // to the controller node will be available on the agent as well
```

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -2,21 +2,75 @@
...
+                            //Check not an instance of jgit or jgitapache using tool descriptor not type key
+                            if (tool.getType().clazz.equals(GitTool.class)) {
+                                //Check for properties
+                                GitTool gitTool = new GitTool(tool.getName(), tool.getHome(), null);
+                                //node specific git home for git tool
+                                gitTool = gitTool.forNode(currentNode, listener);
```
justinharringa, line 64: Does this work as an alternative? https://javadoc.jenkins.io/hudson/tools/ToolLocationNodeProperty.html#getHome-hudson.tools.ToolInstallation-

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -29,43 +29,47 @@
...
-                    if (tool.getType().clazz.equals(GitTool.class)) {
-                        //Check for properties
-                        GitTool gitTool = new GitTool(tool.getName(), tool.getHome(), null);
-                        //Comparing and changing git home for git tool
-                        gitTool = gitTool.forNode(currentNode, listener);
-                        return gitTool.getGitExe();
+            //Check node is not null
+            if (currentNode != null) {
+                if (currentNode.getNodeName().equals("")) {
+                    //It's on the master node
```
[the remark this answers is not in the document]
arpoch: Thanks for the correction Justin,
The logic for getting a git tool in an agent node, depends on the `ToolLocationNodeProperty` property of that node, so if converting a Jenkins instance into a Node type, we will be depending on the `ToolLocationNodeProperty` to get a specific git tool. This approach could be less promising in case of a Jenkins controller.

Their are two ways to add tools in a Jenkins controller
* `Global Tool Configuration` - If the user depends on the tools from `Global Tool Configuration`  then the above logic will not work as Jenkins controller is using global tool and no specific tools are set if taking controller as an agent.
* `Configure System`- If the user sets tool location for the master agent and does not depends on tools from `Global Tool Configuration` then the above logic will work.

So its on the users, how they configure, if they add git tools using  `Configure System` an no git tools are added in the `Global Tool Configuration`, then any new agent created won't be getting a git tools as well.

I have made some more changes based on what I have discussed above. Thanks @justinharringa for bring this up.

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
```diff
@@ -0,0 +1,76 @@
...
+    GitClient getGitClientInstance(TaskListener listener) throws IOException, InterruptedException;
+
+    default String gitToolName(Run<?, ?> run, TaskListener listener) throws IOException, InterruptedException {
+
+        Executor buildExecutor = run.getExecutor();
+        ToolDescriptor<? extends GitTool> toolType;
+        if (buildExecutor != null) {
+            Node currentNode = buildExecutor.getOwner().getNode();
+            //Check node is not null
+            if (currentNode != null) {
```
rishabhBudhouliya, line 36: @arpoch always try to break up complicated business logic, it has two clear benefits:
- clean, cognitively simpler code to maintain
- nested branches of `if else` tend to miss edge cases in my experience

for example, if (currentNode != null && currentNode.getNodeName().equals("")) can be combined together, right?

arpoch: Thanks for the tip Rishabh,  but I made some improvements in the code which makes this suggested change outdated.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -0,0 +1,158 @@
...
+            if (!Functions.isWindows()) {
+                gitEcho = workspace.createTempFile("auth", ".sh");
+                gitEcho.write("#!/bin/sh" + newLine +
+                        "case $1 in" + newLine +
+                        "        Username*) echo " + this.userVariable + newLine +
+                        "                ;;" + newLine +
+                        "        Password*) echo " + this.passVariable + newLine +
+                        "                ;;" + newLine +
+                        "        esac" + newLine, null);
+                gitEcho.chmod(0500);
```
rishabhBudhouliya, line 121: nitpick: can we take this out as a static final variable, something like `FILE_PERMISSION=0500`

rishabhBudhouliya (commented):
I believe we can leverage the existing logic to get the right tool (considering we are not concerned too much about the right user git tool, more about whether it is the correct type)

Impressive work, Harshit!

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
arpoch: Not checked it yet, removed the logger for now

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
arpoch: [translateFor](https://javadoc.jenkins.io/hudson/tools/ToolInstallation.html#translateFor-hudson.model.Node-hudson.model.TaskListener-). It just overrides the node specific values for the git-tool instance variables,  `githome` value for a default git is `git` or `git.exe` but it could be different when the build running on an agent.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -0,0 +1,158 @@
...
+                isAtLeastVersion(2, 3, 0, 0)) {
+            credMap.put("GIT_TERMINAL_PROMPT", "false");
+        } else {
+            credMap.put("GCM_INTERACTIVE", "false");
+        }
+    }
+
+    @Override
+    public GitClient getGitClientInstance(TaskListener listener) throws IOException, InterruptedException {
+        Git gitInstance = Git.with(listener, new EnvVars()).using(gitTool);
```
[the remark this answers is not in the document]
arpoch: The magic happens on the line 52 when the `gitTool` is checked for null, if its not null we can assure that everything is going fall in place as expected.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
arpoch: It has a limited usage, not sure if this would be useful.

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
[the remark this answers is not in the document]
rishabhBudhouliya: This is interesting, what concerns me is the fact that the multi branch project uses the `GitUtils.resolveGitTool` approach to resolve the correct git installation.
If I understand it correctly, it manages both Jenkins controller/agent situation. Certainly, I need to look more, in the meantime do point out any mistake in my assumptions.

src/main/java/org/jenkinsci/plugins/gitclient/GitCredentialBindings.java
rishabhBudhouliya: I think I missed one thing, are we making sure that we are only looking for a cli git installation and not a jgit or jgitapache one?

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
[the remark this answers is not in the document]
rishabhBudhouliya: can getToolName return `jgit` or `jgitapache` as a tool? if yes then it could be a problem

MarkEWaite:
@arpoch my interactive tests of Freestyle projects using credential binding were not successful.  I've attached a zip file of the job definitions that I'm using in case they are any help to you.  You can also refer to the meeting recording from today's session, since the failures were demonstrated during the exploration today.

[JENKINS-28335.zip](https://github.com/jenkinsci/git-client-plugin/files/6698692/JENKINS-28335.zip)

MarkEWaite (commented):
Doc comments for your consideration

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -0,0 +1,158 @@
...
+            super(gitUsername + ":" + gitPassword, credentialId);
+            this.userVariable = gitUsername;
+            this.passVariable = gitPassword;
+        }
+
+        @Override
+        protected FilePath write(StandardUsernamePasswordCredentials credentials, FilePath workspace)
+                throws IOException, InterruptedException {
+            FilePath gitEcho;
+            String newLine = System.lineSeparator();    //Platform dependent newLine
```
jtnord, line 111: this looks wrong to me.  IIUC this runs on the controller not the agent so `System.lineSeparator()` is just a bad guess at waht the agent is running.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -0,0 +1,158 @@
...
+            if (!Functions.isWindows()) {
```
jtnord, line 112: same here - this code appears to be run on the controller.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    @Override
+    public GitClient getGitClientInstance(TaskListener listener) throws IOException, InterruptedException {
+        Git gitInstance = Git.with(listener, new EnvVars()).using(gitTool);
+        return gitInstance.getClient();
+    }
+
+    protected static final class GenerateGitScript extends AbstractOnDiskBinding<StandardUsernamePasswordCredentials> {
```
jtnord, line 96: am I missing somethign here that informs the credential plugin that  the actual username and password should be masked?

If I ran GIT_ASK_PASS from the CLI (outside of git) would the password be revealed in plain text?

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        @Override
+        protected FilePath write(StandardUsernamePasswordCredentials credentials, FilePath workspace)
+                throws IOException, InterruptedException {
+            FilePath gitEcho;
+            String newLine = System.lineSeparator();    //Platform dependent newLine
+            if (!Functions.isWindows()) {
+                gitEcho = workspace.createTempFile("auth", ".sh");
+                gitEcho.write("#!/bin/sh" + newLine +
```
jtnord, line 114: should not assume `#!bin/sh`  the shell can be set for each agent and may be different

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
jtnord, line 112: NIT: rather than `if (!boolean) {something} else { something else }`  do `if (boolean) { somethingelse } else {something}`

jtnord (commented):
should there be some inline or other docs telling users they need to disable the use of the credential cache to prevent the password being cached locally (and if not using ephemeral agents being used by another job randomly?)

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
MarkEWaite: As far as I can tell, the credentials binding plugin classes provide the masking of the username and password.  At least that is what I've seen in my interactive testing.

src/main/java/org/jenkinsci/plugins/gitclient/GitUsernamePasswordBinding.java
```diff
@@ -43,14 +43,18 @@ public GitUsernamePasswordBinding(String credentialsId) {
         return StandardUsernamePasswordCredentials.class;
     }
 
+    private static void setUnixNodeType(boolean value) {
+        unixNodeType = value;
```
arpoch, line 47: @MarkEWaite, I am a bit doubtful regarding the use of `unixNodeType` as a class variable and not as instance variable. The scenario which I think could cause a problem is, when the user adds multiple `GitUsernamePassword` bindings and executes these in different operating system environments, won't  `unixNodeType` value be same for the all the `GitUsernamePassword` bindings since it's a class variable or am I lacking some insight over the concept of Descriptors and Describables?

I thought of making `unixNodeType` as an instance type variable and passing it to the `GenerateGitScript` constructor as a parameter.

MarkEWaite: I am also doubtful of using it as a class variable rather than an instance variable.  The code change that I submitted was only done so that I could get an incremental build from ci.jenkins.io in order to do some exploratory tests on my Jenkins instance.

MarkEWaite:
Closed in preference to https://github.com/jenkinsci/git-plugin/pull/1104

PR #713, opened 2021-06-09
[JENKINS-65838] Do not set proxy authentication if username and/or password are not null but empty strings

ababushk:
## [JENKINS-65838](https://issues.jenkins.io/browse/JENKINS-65838) - "Proxy authentication error" during repository cloning when username and password are empty in Jenkins Update Center proxy settings

When proxy is set from Jenkins Update Center UI with empty username and password the plugin still tries to use authentication and user gets "Proxy Authentication Required" message during Git LFS pull
## Checklist

- [x] I have added documentation as necessary (no documentation changes necessary)

## Types of changes


MarkEWaite:
Is it feasible to add one or more tests that will show the problem?  I don't use a proxy host so I have no way to reliably test that this change is effective and does not have undesired side effects.

Do you have other users that can test the proposed change and report their results?

ababushk:
It might be feasible... I think I can move proxy setting code to separate method and then write a test for it.

I can ask other people in my company to try to reproduce the bug and to test the fix. I'm not sure I will be able to find people outside my org with the same proxy server configuration which has revealed the bug.

MarkEWaite (approved):

PR #724, opened 2021-07-06
Check the least Cli Git version required 

arpoch:
## Compare the Cli Git version with the required Git version

The `CliGitAPIImpl` class provides a method `isAtLeastVersion` which returns true is the required git is at-least the cli git version. Since this method have package-private scope thus restricting it's usage outside the plugin/package. To overcome this issue a new public scoped method `compareLeastGitVersion` is create with similar working. 

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -300,7 +300,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * Compare the current cli git version with the required version.
      * Finds if the current cli git version is at-least the required version
      *
-     * Returns True if the current cli git version is al-least the required version
+     * Returns True if the current cli git version is at least the required version
      *
      * @param major required major version for command line git
      * @param minor required minor version for command line git
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -302,11 +302,11 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      *
      * Returns True if the current cli git version is al-least the required version
      *
-     * @param major ,required major version for cli git
-     * @param minor ,required minor version for cli git
-     * @param rev ,required revision for cli git
-     * @param bugfix ,required patches for cli git
-     * @return true if the cli git version is al-least the required version
+     * @param major required major version for command line git
+     * @param minor required minor version for command line git
+     * @param rev required revision for command line git
+     * @param bugfix required patches for command line git
+     * @return true if the command line git version is at least the required version
      **/
     public boolean compareLeastGitVersion(int major, int minor, int rev, int bugfix) {
         return isAtLeastVersion(major,minor,rev,bugfix);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -296,6 +296,22 @@ private void getGitVersion() {
...
+     *
+     * Returns True if the current cli git version is al-least the required version
+     *
+     * @param major ,required major version for cli git
+     * @param minor ,required minor version for cli git
+     * @param rev ,required revision for cli git
+     * @param bugfix ,required patches for cli git
+     * @return true if the cli git version is al-least the required version
+     **/
+    public boolean compareLeastGitVersion(int major, int minor, int rev, int bugfix) {
```
MarkEWaite, line 311: I believe that Java compare functions are generally expected to return an int with values less than zero meaning the current object is "less" than the argument, zero meaning that the current object is the same as the argument, and greater than zero if the current object is "greater" than the argument.  See the [String.compareTo javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#compareTo-java.lang.String-] for more details.

I think that you want a method name that uses the word `is` rather than `compare` or you need to create a new implementation that does a comparison.  I believe that I used `is` in the original definition because I didn't envision a time when we would need to sort a list of git implementations by their version number.  I don't think we need to sort git implementations by version number for this case, though I could be misunderstanding your need for this API.

If we need ordering / sorting of CliGitAPIImpl objects, then it is probably best to implement the [Comparable interface](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html)
```suggestion
    public boolean compareLeastGitVersion(int major, int minor, int rev, int bugfix) {
```

I'd also be fine if you changes isAtLeastVersion from a private method to a public method.  If the `isAtLeastVersion` name is not clear enough, then adding a new public method with a better name is also fine.

MarkEWaite (changes requested):

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -296,6 +296,22 @@ private void getGitVersion() {
         return gitVersion >= requestedVersion;
     }
 
+    /**
+     * Compare the current cli git version with the required version.
+     * Finds if the current cli git version is at-least the required version
+     *
+     * Returns True if the current cli git version is al-least the required version
```
MarkEWaite, line 303: Spelling fix
```suggestion
     * Returns True if the current cli git version is at least the required version
```

MarkEWaite (approved):
Would like one spelling fix, but otherwise I believe it is ready to go

MarkEWaite:
@rishabhBudhouliya and @justinharringa , any issue for you in the proposed change?  I'd like to release it before the end of this week if it looks good to you.

justinharringa:
I'm curious if there's benefit to having this method separate instead of just making the original method public. Is there other logic we'd place here?

rishabhBudhouliya:
+1 to @justinharringa's concern. I am not sure how the package protection of `isAtLeastVersion` is still applicable after wrapping it around a public method.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -296,6 +296,22 @@ private void getGitVersion() {
         return gitVersion >= requestedVersion;
     }
 
+    /**
+     * Compare the current cli git version with the required version.
+     * Finds if the current cli git version is at-least the required version
```
MarkEWaite, line 301:
```suggestion
     * Finds if the current cli git version is at-least the required version.
```

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -296,6 +296,22 @@ private void getGitVersion() {
...
+     *
+     * Returns True if the current cli git version is at least the required version
```
MarkEWaite, line 303:
```suggestion
     * Returns True if the current cli git version is at least the required version.
```

MarkEWaite (approved):

PR #741, opened 2021-09-04
Adapting `GitToolResolverTest` to jenkinsci/jenkins#5425

basil:
I noticed that this test had started failing when running BOM tests against a recent incremental of Jenkins core. It appears that this started failing after jenkinsci/jenkins#5425. This patch corrects the problem so that the test works with both old and new versions of Jenkins core.

### DIFF
--- a/src/test/java/hudson/plugins/git/GitToolResolverTest.java
+++ b/src/test/java/hudson/plugins/git/GitToolResolverTest.java
@@ -6,6 +6,7 @@ import hudson.tools.AbstractCommandInstaller;
 import hudson.tools.BatchCommandInstaller;
 import hudson.tools.CommandInstaller;
 import hudson.tools.InstallSourceProperty;
+import hudson.util.VersionNumber;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

MarkEWaite:
Thanks @basil .  I've had to make similar changes to extend test support when the default git branch is no longer called `master`.

timja:
@MarkEWaite is it possible to get this released for PCT in bom please?

MarkEWaite:
Should be able to do so this week

MarkEWaite:
> Should be able to do so this week

Released git client plugin 3.10.0.  Requires Jenkins 2.289.1 and delivers JGit 5.13.0 instead of JGit 5.12.0 that was included in git client plugin 3.9.0.

I assume that it is enough to fix the `GtiResolverTest` on 2.289.1 and later, since the test should not fail on earlier versions

timja:
https://github.com/jenkinsci/bom/pull/664

PR #750, opened 2021-09-15
Introduce new suppressMergeCommitDiff parameter in showRevision 

anatoly-spb:
The -m flag breaks polling changes mechanism, for example: 

https://issues.jenkins.io/browse/JENKINS-20389
https://issues.jenkins.io/browse/JENKINS-23606
https://issues.jenkins.io/browse/JENKINS-36180
https://issues.jenkins.io/browse/JENKINS-65932

This pull request is the first step to fix above issues. The next step is to call showRevision in https://github.com/jenkinsci/git-plugin/blob/master/src/main/java/hudson/plugins/git/GitSCM.java#L2063 with suppressMergeCommitDiff=true to suppress merge commit diff in change log.

## Checklist

## Types of changes

- [x]  Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1286,7 +1286,18 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     /** {@inheritDoc} */
     @Override
     public List<String> showRevision(ObjectId from, ObjectId to, Boolean useRawOutput) throws GitException, InterruptedException {
+    	return showRevision(from, to, useRawOutput, false);
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public List<String> showRevision(ObjectId from, ObjectId to, Boolean useRawOutput, Boolean supressMergeCommitDiff) throws GitException, InterruptedException {
         ArgumentListBuilder args = new ArgumentListBuilder("log", "--full-history", "--no-abbrev", "--format=raw", "-M");
+        
+        if (!supressMergeCommitDiff) {
+            args.add("-m");
+        }
+        
         if (useRawOutput) {
             args.add("--raw");
         }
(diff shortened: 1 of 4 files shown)


MarkEWaite:
The change breaks compatibility as shown by the automated test that fails.  It will need to be a flag that is passed into the API rather than an "always on" change.  The default of the flag will need to be "off" so that we don't break compatibility for users.

MarkEWaite (changes requested):
This will break compatibility with existing use cases.  It needs to be only enabled when a new argument is toggled to enable the new behavior.

anatoly-spb:
> This will break compatibility with existing use cases. It needs to be only enabled when a new argument is toggled to enable the new behavior.

Thank you @MarkEWaite. May be you know where is polling change mechanism implemented to call showRevision  with additional argument?

MarkEWaite:
Yes, that sounds about right.  Include tests to confirm the old behavior and the new behavior

anatoly-spb:
@MarkEWaite are you going to merge this request or what should I do to push this work will be done?

MarkEWaite:
> ... Include tests to confirm the old behavior and the new behavior

@anatoly-spb as far as I can tell, you have not created any tests to confirm the new behavior is what you expect.  The existing behavior might be covered by existing tests, but the new behavior is not covered by tests.  Without tests, it is far too easy for a later change to undo or damage the behavior that you want.  Please provide tests

MarkEWaite (changes requested):
Needs tests to confirm new behavior and to assure that current behavior is retained

MarkEWaite:
Closing because tests have not been added to confirm the new behavior works as expected.

anatoly-spb:
> Closing because tests have not been added to confirm the new behavior works as expected.

Could you help me to add such tests?

murerfel:
Any chances this will get re-opened and re-evaluated? It seems it would fix a problem that has been plaguing us for years now (https://issues.jenkins.io/browse/JENKINS-23606)

MarkEWaite:
> Any chances this will get re-opened and re-evaluated? It seems it would fix a problem that has been plaguing us for years now (https://issues.jenkins.io/browse/JENKINS-23606)

With my current workload, there is no chance that I will reopen it and re-evaluate it.  Others are welcome to reopen it and re-evaluate it, but my workload does not allow me to do the additional work needed on this pull request.

@anatoly-spb indicated that new tests have not been added because help is needed to create those new tests.  I don't have capacity right now to create those new tests.  If others would like to create those new tests and report that they have deployed the incremental build of the plugin into production, that would might persuade me or other maintainers of the plugin to evaluate the changes.

@murerfel are you willing to add the new tests, merge the latest changes from the master branch, deploy an incremental build into your environment, and report the results here?

murerfel:
Thank you, @MarkEWaite, the quick response is appreciated. Unfortunately, I have neither the required expertise nor the resources for such an endeavor at this point. So we'll just have to be patient for now.

PR #771, opened 2021-11-05
Delete `RemotingTest`

jglick:
Not sure what this was originally intended to test, exactly; probably some sort of exploratory code useful during the development of the `git-plugin` 2.0 refactoring. Certainly the fact that Git commands can be run on an agent should be well covered elsewhere in a more realistic manner. This test popped up as a mysterious failure while experimenting with https://github.com/jenkinsci/jenkins-test-harness/pull/344 and https://github.com/jenkinsci/jenkins/pull/5885, so best to remove it to avoid contaminating test output with artificial tests which do not correspond to actual use cases.

Also, it fails for me locally anyway:

```
org.eclipse.jgit.api.errors.ServiceUnavailableException: Signing service is not available
	at org.eclipse.jgit.api.CommitCommand.sign(CommitCommand.java:290)
	at org.eclipse.jgit.api.CommitCommand.call(CommitCommand.java:247)
	at org.jenkinsci.plugins.gitclient.JGitAPIImpl.commit(JGitAPIImpl.java:460)
```


MarkEWaite (approved):

basil:
@MarkEWaite Would be nice to get a release of this. Without a release, this test fails in PCT / `bom-weekly` for 2.326.

MarkEWaite:
I should be able to release it within the next 48 hours.  I'd like some interactive test time before the release and I need to confirm that the Javadoc build failure is fixed.

PR #814, opened 2022-02-02
JENKINS-60940: Convert git client plugin tests from JUnit 3 to JUnit 4

kdaud:
## [JENKINS-60940](https://issues.jenkins.io/browse/JENKINS-60940) - Convert JUnit 3 tests to JUnit 4

Convert JUnit 3 tests to JUnit 4 so that they are easier to maintain and easier to understand.

## Checklist

## Types of changes

- [x] Tests


MarkEWaite:
@kdaud thanks very much for doing this.  Once I've started the review of the changes, please don't use `--force` to force push changes.  When new changes replace existing changes, the comments are no longer associated with the change.  

It is better that the changes evolve as a series of small commits and then we can decide before merging if the many small commits should be squashed into a single commit.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITest.java
```diff
@@ -58,7 +58,7 @@
  */
 
 @RunWith(Parameterized.class)
-public class GitAPITest {
+public abstract class GitAPITest {
```
MarkEWaite, line 61: I'd rather we not change this parameterized test class that checks command line git and two variants of JGit into an abstract class.  If you find that you truly need an abstract class, create a new class to receive those tests that need the abstract class.

MarkEWaite (changes requested):

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITest.java
kdaud: Thanks @MarkEWaite for the quick review, my need was to convert this class to an abstract class so that its able to host test methods that consumes abstract methods. 

> If you find that you truly need an abstract class, create a new class to receive those tests that need the abstract class.

I get now have the way to go based on the guidance.

kdaud:
@MarkEWaite have updated the commit accordingly !

MarkEWaite (changes requested):
@kdaud the change you've proposed removes all the tests from GitAPITestCase and places them in an abstract test class GitAPICaseTest.  Because GitAPICaseTest is abstract, none of the tests are executed.  That's not what is wanted with the transition from JUnit 3 to JUnit 4.

The instructions in [JENKINS-60940](https://issues.jenkins.io/browse/JENKINS-60940) specifically advise that the transformation from JUnit 3 to JUnit 4 be performed one test at a time.  

It may be possible to do more conversions in a single step with your abstract test class idea, but if so, then the tests must execute.  It is not enough to copy the test source code to a new location that causes them to not be executed.

MarkEWaite:
I'm closing this in favor of the process started in #815

PR #824, opened 2022-03-02
Replace JUnit 3 tests with JUnit 4 tests

Hrushi20:
## [JENKINS-60940](https://issues.jenkins.io/browse/JENKINS-60940) - Replace JUnit 3 tests with JUnit 4 tests

Created a new class for testing GitAPITestCase class which is a non abstract class. Updating the junit version from v3 to v4.

## Checklist

## Types of changes


Hrushi20:
@MarkEWaite, I created a new class named GitAPITestUpdate which is a non abstract class, that will contain all the tests related to GitAPITestCase. I just initialized the class with the basic functionalities. I'll add the test methods in each commit in an incremental manner.

MarkEWaite:
> @MarkEWaite, I created a new class named GitAPITestUpdate which is a non abstract class, that will contain all the tests related to GitAPITestCase. I just initialized the class with the basic functionalities. I'll add the test methods in each commit in an incremental manner.

I don't object strenuously to this approach, but I suspect you would have a better outcome if you migrated one test at a time and inserted the additional items needed (WorkingArea type items) to support the test only as needed.  I know that one of my concerns was that the number of tests per test class is quite large  (142 tests in JUnit 4 GitClientTest, 60 tests in JUnit 4 GitAPITest, 60 tests in JUnit 4 GitAPITest, and 48 tests left to convert in GitAPITestCase) and that makes it more challenging to use multiple test processes to run tests faster on multi-core machines.

Hrushi20:
I just added the Working Area type items initially because most of the tests in GitAPITestCase use working Area class. If I add any of the test, the working area is required to run the test. I only added Working Area class and it's dependencies. Slowly I can start adding the testing in each step.

Hrushi20:
Hi @MarkEWaite, Quick question on upgrading code to junit4. Should I upgrade the test cases that are deprecated too?

Hrushi20:
I forced push due to a wrong commit message.

Hrushi20:
Thanks @MarkEWaite, I'll upgrade them as well.

MarkEWaite:
Thanks for the pull request.  

There are some things that are lost in the draft pull request that you're proposing.

* `GitAPITestCase` is extended by `CliGitAPIImplTest`, `JGitAPIImplTest`, and `JGitApacheAPIImplTest` so that the same test source code is executed with three different git implementations.  I don't see how those three implementations are being tested in the tests in 
* `test_lsTree_non_recursive` seems to have been deleted and not replaced with a new implementation.  I don't see any new calls to `lsTree()` in the new code

There are some tests that migrated but have not been removed from their original JUnit 3 implementation class.  Tests like `testPushDeprecatedSignature` and `testTrackingSubmodule` still exist in the JUnit 3 implementation.

Some of the tests were renamed while others retain their original names.  It is a little easier for me to review if they retain the same name or if they are consistently all renamed in the transition.

The formatting in the new file uses leading tabs.  The contributing document advises that 

> New code should follow the link:https://github.com/jenkinsci/scm-api-plugin/blob/master/CONTRIBUTING.md#code-style-guidelines[SCM API code style guidelines].

in the [SCM API code style guidelines](https://github.com/jenkinsci/scm-api-plugin/blob/master/CONTRIBUTING.md#indentation) it says:

> **Use spaces**. Tabs are banned.

It also states that Java blocks are 4 spaces.

The `// moved` comments on private methods seem like they won't help you much.  Until the private method is unused, it needs to remain in the JUnit 3 file.  When the private method is no longer used, the spotbugs checks should tell us that the method is unused and can be deleted.

Hrushi20:
Thanks for the review @MarkEWaite,
* I thought I'll extend the classes `CliGitAPIImplTest`, `JGitAPIImplTest` & `JGitApacheAPIImplTest` in the end to GitAPITestUpdate, but I'll update the extended class in the upcoming PR.
* I also thought of removing the junit3 tests together at once, but understood it's better to remove them after upgrading it to junit v4.
* After upgrading a few tests I realized that the names used are all camelCase. Hence I abruptly changed it to camelCase. The names for all the tests will be renamed to camelCase.

Hrushi20:
I converted tabs to spaces. Just asking out of curiosity :bulb:, tabs makes the file size smaller than spaces, Ascii value of Tab used once will be smaller when compared to Ascii value of Space used four times. Then why use space :thinking: :thinking: :thinking:

Hrushi20:
Moved all the tests from `GitAPITestCase` to `GitAPITestUpdate`. Need to remove the unnecessary methods in GitAPITestCase and then delete the file.

MarkEWaite:
> I converted tabs to spaces. Just asking out of curiosity 💡, tabs makes the file size smaller than spaces, Ascii value of Tab used once will be smaller when compared to Ascii value of Space used four times. Then why use space 🤔 🤔 🤔

Good question.  I like code formatting standards because they avoid back and forth changes that involve only style changes.  The choice of tabs vs. spaces is somewhat arbitrary (see the [holy war on stackexchange](https://softwareengineering.stackexchange.com/questions/57/tabs-versus-spaces-what-is-the-proper-indentation-character-for-everything-in-e) for discussion of both sides), but the SCM API coding standard chose spaces and the Netbeans IDE that I sometimes use prefers spaces and the emacs configuration that I frequently use prefers spaces.

Eventually, I hope to implement mandatory automated formatting of new files as they are added to the repository.  Existing files cannot be automatically formatted because they will cause conflicts and break the many pending pull requests in the repository.

Hrushi20:
Thanks @MarkEWaite. Didn't know there was a war between Tabs and spaces :smiley:. The automated formatting would make things easier. Can we create a config file for different kinds of IDE to support the required code style?

Hrushi20:
Hey @MarkEWaite, can you review the PR?

MarkEWaite:
@Hrushi20 I'm out of the office this week on holiday.  I won't review it until next week.

Hrushi20:
Oki @MarkEWaite. Enjoy your holidays.

MarkEWaite:
@Hrushi20 the tests need to pass in CI.  There isn't much motivation for me to review code that has failing tests.  If you need help understanding why the tests are failing, I may be able to assist with that, but generally I'd prefer that tests are passing before I review the code.

Hrushi20:
@MarkEWaite All of the test which are failing throw this error ```hudson.plugins.git.GitException: org.eclipse.jgit.api.errors.ServiceUnavailableException: Signing service is not available
```
The tests in the CliGitImpl are running successfully. Only the tests in the JGitImpli are failing either because the methods doesn't support JGit or the method is deprecated.

File: src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java [hunk 1]
```diff
@@ -0,0 +1,2161 @@
...
+        // JGit submodule implementation doesn't handle renamed submodules
+        if (w.igit() instanceof CliGitAPIImpl) {
+            assertTrue("modules/sshkeys does not exist", w.exists("modules/sshkeys"));
+        }
+        assertFixSubmoduleUrlsThrows();
+    }
+
+
+    /* Shows the submodule update is broken now that tests/getSubmodule includes a renamed submodule */
+    @NotImplementedInJGit
```
MarkEWaite, line 924: This annotation is not honored in the implementation you've created for JUnit 4.  It was used in JUnit 3 to cause the test to be skipped when the implementation was JGit.  In JUnit 4 you can either move the test into a class that is dedicated to testing the CliGit implementation or you can add a conditional at the start of the test that exits immediately if the git implementation is not command line git.

I prefer that it be moved to a separate file so that the test count is not incorrectly suggesting that this test is run for JGit.
```suggestion
```

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -0,0 +1,2161 @@
...
+        if (w.igit() instanceof CliGitAPIImpl) {
+            assertTrue("modules/sshkeys does not exist", w.exists("modules/sshkeys"));
+        }
+        assertFixSubmoduleUrlsThrows();
+
+        String shallow = Paths.get(".git", "modules", "module", "1", "shallow").toString();
+        assertFalse("shallow file existence: " + shallow, w.exists(shallow));
+    }
+
+    @NotImplementedInJGit
```
MarkEWaite, line 945: Annotation is not honored by JUnit 4 implementation.  Remove and either move the test to a CliGit specific file or guard the test with a conditional
```suggestion
```

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -0,0 +1,2161 @@
...
+
+        String shallow = Paths.get(".git", "modules", "submodule", "shallow").toString();
+        assertEquals("shallow file existence: " + shallow, hasShallowSubmoduleSupport, w.exists(shallow));
+
+        int localSubmoduleCommits = w.cgit().subGit("submodule").revList(defaultBranchName).size();
+        int remoteSubmoduleCommits = remote.cgit().subGit("dir-submodule").revList(defaultBranchName).size();
+        assertEquals("submodule commit count didn't match", hasShallowSubmoduleSupport ? 2 : remoteSubmoduleCommits, localSubmoduleCommits);
+    }
+
+    @NotImplementedInJGit
```
MarkEWaite, line 997: Same as earlier comments on this annotation
```suggestion
```

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -0,0 +1,2161 @@
...
+        assertTrue("modules/firewall does not exist", w.exists("modules/firewall"));
+        assertTrue("modules/ntp does not exist", w.exists("modules/ntp"));
+        // JGit submodule implementation doesn't handle renamed submodules
+        if (w.igit() instanceof CliGitAPIImpl) {
+            assertTrue("modules/sshkeys does not exist", w.exists("modules/sshkeys"));
+        }
+        assertFixSubmoduleUrlsThrows();
+    }
+
+    @NotImplementedInJGit
```
MarkEWaite, line 1015: Same here
```suggestion
```

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -0,0 +1,2161 @@
...
+     * Test getRemoteReferences with listing all references
+     */
+     @Test
+    public void testGetRemoteReferences() throws Exception {
+        Map<String, ObjectId> references = w.git.getRemoteReferences(remoteMirrorURL, null, false, false);
+        assertTrue(references.containsKey("refs/heads/" + DEFAULT_MIRROR_BRANCH_NAME));
+        assertTrue(references.containsKey("refs/tags/git-client-1.0.0"));
+    }
+
+    @NotImplementedInJGit
```
MarkEWaite, line 1096: Same here
```suggestion
```

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -0,0 +1,2161 @@
...
+        assertEquals("Wrong blob 1 sha1", expectedBlob1SHA1, tree.get(0).getObject());
+        assertEquals("Wrong blob 2 sha1", expectedBlob2SHA1, tree.get(1).getObject());
+        assertEquals("Wrong number of tree entries", 2, tree.size());
+        final String remoteUrl = "https://github.com/jenkinsci/git-client-plugin.git";
+        w.git.setRemoteUrl("origin", remoteUrl);
+        assertEquals("Wrong origin default remote", "origin", w.igit().getDefaultRemote("origin"));
+        assertEquals("Wrong invalid default remote", "origin", w.igit().getDefaultRemote("invalid"));
+    }
+
+    @NotImplementedInJGit
```
MarkEWaite, line 1475: Same here
```suggestion
```

MarkEWaite (changes requested):
Need to move the methods that are not implemented for JGit into a file that is not executed for JGit.

MarkEWaite:
> @MarkEWaite All of the test which are failing throw this error
> 
> ```
> hudson.plugins.git.GitException: org.eclipse.jgit.api.errors.ServiceUnavailableException: Signing service is not available
> ```
> 
> The tests in the CliGitImpl are running successfully. Only the tests in the JGitImpli are failing either because the methods doesn't support JGit or the method is deprecated.

Check the CI jobs.  I don't see that error on any of the failing tests.

Hrushi20:
I am moving the methods to another abstract class. This abstract class is inheriting the GitAPITestUpdate. The fields present in the GitAPITestUpdate are private and static. Can I convert them to package protected so that the sub class can access those fields? 

Eg: 

```
 private static String defaultBranchName = "mast" + "er"; 
```
I am not able to use default Branch Name in the subclass.

MarkEWaite:
> I am moving the NotImplementedInJGit methods to another abstract class. This abstract class is inheriting the GitAPITestUpdate. The fields present in the GitAPITestUpdate are private and static. Can I convert the required fields to package protected so that the sub class can access those fields?
> 
> Eg:
> 
> ```
>  private static String defaultBranchName = "mast" + "er"; 
> ```

You're welcome to make them package protected so that they can be accessed elsewhere in the test package.

Hrushi20:
Hey @MarkEWaite, all tests have passed successfully :smile: :smile: :smile:

MarkEWaite:
Thanks @Hrushi20 !  Since you've requested a review, could you change the pull request so that it is no longer "Draft"?

Screenshot 2022-03-30 084208

MarkEWaite (approved):
Thanks!

jimklimov:
FYI: Seems this major relocation did not account (carry over) at least my recent fix from PR #886, will re-post shortly.

I did not trawl git history further to see if other tests were changes while this PR was in the works, and whether those changes got lost too or got accounted for.

PR #847, opened 2022-05-08
Make a comment more inclusive

Ladyprowess:
Rather than "allow" use "whitelist"

- [x] Make sure you are opening from a topic/feature/bugfix branch (right side) and not your main branch!
- [x] Link to relevant issues in Github or Jira
- [x] Link to the relevant pull request, esp. upstream and downstream changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CloneCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CloneCommand.java
@@ -93,7 +93,7 @@ public interface CloneCommand extends GitCommand {
     CloneCommand noCheckout();
 
     /**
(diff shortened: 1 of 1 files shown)


MarkEWaite (changes requested):
Thanks for the pull request, but this is the opposite direction from what we want.  We want to change "whitelist" and "blacklist" to better phrases like "allow" and "block".

MarkEWaite:
I'm closing this one without merging because the current phrasing is preferred ahead of the proposed phrasing.  Thanks for the pull request!

Ladyprowess:
> I'm closing this one without merging because the current phrasing is preferred ahead of the proposed phrasing. Thanks for the pull request!

Okay @MarkEWaite Noted!

PR #848, opened 2022-05-08
Make a comment more inclusive

Ladyprowess:
Rather than "allow" use "whitelist"

- [x] Make sure you are opening from a topic/feature/bugfix branch (right side) and not your main branch!
- [x] Link to relevant issues in Github or Jira
- [x] Link to the relevant pull request, esp. upstream and downstream changes


MarkEWaite (changes requested):
We want to replace usages of "whitelist" and "blacklist" with better terms like "allow" and "forbid".

Ladyprowess:
> We want to replace usages of "whitelist" and "blacklist" with better terms like "allow" and "forbid".

Okay, @MarkEWaite thank you, I thought otherwise

PR #881, opened 2022-07-29
[JENKINS-69149] Use Java for accept-new in StrictHostKeyChecking 

Pldi23:
## [JENKINS-69149](https://issues.jenkins.io/browse/JENKINS-69149) Use Java for accept-new in StrictHostKeyChecking

As `accept-new` is not supported in several environments this PR implement it's logic manually.

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifier.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifier.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -18,9 +19,28 @@ public class AcceptFirstConnectionVerifier extends HostKeyVerifierFactory {
 
     @Override
     public AbstractCliGitHostKeyVerifier forCliGit(TaskListener listener) {
-        return tempKnownHosts -> {
-            listener.getLogger().println("Verifying host key using known hosts file, will automatically accept unseen keys");
-            return "-o StrictHostKeyChecking=accept-new -o HashKnownHosts=yes";
+        return (tempKnownHosts, url) -> {
+            String hostnamePort = url.getPort() > 0 ? url.getHost() + ":" + url.getPort() : url.getHost();
+            listener.getLogger().printf("Verifying host key for %s using %s, will automatically accept unseen keys %n", hostnamePort, getKnownHostsFile().toPath());
+            String strictHostKeyChecking;
+            if (!Files.exists(getKnownHostsFile().toPath())) {
+               strictHostKeyChecking = "no";
+            } else {
+
+                List<String> hostKeyLines = Files.readAllLines(getKnownHostsFile().toPath());
+                boolean isNew = hostKeyLines.stream()
+                        .map(line -> line.split("\\s+")[0])
+                        .noneMatch(hostname -> HostNameHashVerifier.checkHashed(hostname, hostnamePort) ||
+                                HostNameHashVerifier.checkHashed(hostname, url.getHost()) ||
+                                hostname.equals(url.getHost()) ||
+                                (hostname.contains(url.getHost()) && (url.getPort() <= 0 || hostname.contains(String.valueOf(url.getPort())))));
+                if (isNew) {
+                    strictHostKeyChecking = "no";
+                } else {
+                    strictHostKeyChecking = "yes";
+                }
+            }
+            return "-o StrictHostKeyChecking=" + strictHostKeyChecking + " -o HashKnownHosts=yes";
         };
     }
 
--- a/src/main/java/org/jenkinsci/plugins/gitclient/verifier/HostNameHashVerifier.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/verifier/HostNameHashVerifier.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2007-2008 Trilead AG (http://www.trilead.com)
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+a.) Redistributions of source code must retain the above copyright
+    notice, this list of conditions and the following disclaimer.
+b.) Redistributions in binary form must reproduce the above copyright
+    notice, this list of conditions and the following disclaimer in the
+    documentation and/or other materials provided with the distribution.
+c.) Neither the name of Trilead nor the names of its contributors may
+    be used to endorse or promote products derived from this software
+    without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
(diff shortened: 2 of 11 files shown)


src/main/java/org/jenkinsci/plugins/gitclient/verifier/HostNameHashVerifier.java
```diff
@@ -0,0 +1,171 @@
...
+
+import com.trilead.ssh2.crypto.Base64;
+import com.trilead.ssh2.crypto.digest.MessageMac;
+import com.trilead.ssh2.crypto.digest.SHA1;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
+
+public class HostNameHashVerifier {
```
Pldi23, line 100: Copy pasted from [link](https://github.com/jenkinsci/trilead-ssh2/blob/main/src/com/trilead/ssh2/KnownHosts.java#L190-L228) to not write it by myself

src/main/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifier.java
```diff
@@ -18,9 +19,28 @@ public class AcceptFirstConnectionVerifier extends HostKeyVerifierFactory {
...
-            return "-o StrictHostKeyChecking=accept-new -o HashKnownHosts=yes";
+        return (tempKnownHosts, url) -> {
+            String hostnamePort = url.getPort() > 0 ? url.getHost() + ":" + url.getPort() : url.getHost();
+            listener.getLogger().printf("Verifying host key for %s using %s, will automatically accept unseen keys %n", hostnamePort, getKnownHostsFile().toPath());
+            String strictHostKeyChecking;
+            if (!Files.exists(getKnownHostsFile().toPath())) {
+               strictHostKeyChecking = "no";
+            } else {
+
+                List<String> hostKeyLines = Files.readAllLines(getKnownHostsFile().toPath());
```
Pldi23, line 30: I am now thinking about how to synchronize reading here and then writing that most likely happens [there](https://github.com/jenkinsci/git-client-plugin/blob/38abd1db42054d461f5933c25e53140684accd63/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L2676)

Pldi23:
@MarkEWaite  @dwnusbaum please review (I don't have permissions to add reviewers)

src/main/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifier.java
Pldi23: added synchronization in [80f6d57](https://github.com/jenkinsci/git-client-plugin/pull/881/commits/80f6d578ed8bf3b68e4c89cd63a81cf18e52b3a8)

driverpt:
Any ETA on this one ?

dwnusbaum:
Personally, I am more inclined to change the default host key verification strategy either back to no host key checking (and log a warning in this mode if we do not already) or strict host key checking and/or improve error messages to make it easier for users to understand that changing the mode globally is very easy before I would reimplement `accept-new` in Java code. I am just not convinced that the complexity of maintaining this mode ourselves is worth it.

We knew that this mode was not going to be supported on operating systems with old versions of OpenSSH and that those users would need to change the default mode. We did not realize how widespread those OSs still were. Users who are mentioning that they are downgrading `git-client` in the Jira issue to me indicate that our error messages and/or documentation are not clear or not visible enough, because it would be much easier for those users to just change the mode to no verification than to downgrade the plugin and restart Jenkins.

Pldi23:
Maybe making just switch default strategy to 'Known hosts file' which is using `StrictHostKeyChecking=yes` is a good idea, we know that users used `StrictHostKeyChecking=no` before the fix so theirs `known_hosts` file should be already populated with needed hosts, so compatibility will be retained.

Pldi23:
One more downside in changing default strategy now, could be that some of users already performed the setup of strategies and I worry that we could simply rewrite their setup, so simple change [here](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/GitHostKeyVerificationConfiguration.java#L24) will not help.

Pldi23:
> Users who are mentioning that they are downgrading git-client in the Jira issue indicate to me that our error messages and/or documentation are not clear or not visible enough, because it would be much easier for those users to just change the mode to no verification than to downgrade the plugin and restart Jenkins.

Agree, the most visible would be to print one more message below this [one ](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifier.java#L22)
`listener.getLogger().println(“If you are using OpenSSH < 7.6 please choose another strategy to verify ssh host key in ‘Manage Jenkins’ -> ‘Configure Global Security’ -> ‘Git Host Key Verification Configuration'”);`

dwnusbaum:
> we know that users used StrictHostKeyChecking=no before the fix so theirs known_hosts file should be already populated with needed hosts, so compatibility will be retained.

Oh, I did not realize this was the case.

> One more downside in changing default strategy now, could be that some of users already performed the setup of strategies and I worry that we could simply rewrite their setup, so simple change [here](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/GitHostKeyVerificationConfiguration.java#L24) will not help.

True, but even so, I would just change that line, not anything else. Users who downgraded or have not updated yet would still benefit, and anyone who has updated and picked a default strategy explicitly would be unaffected, right?

> The only bad thing that if we choose to make 'Known hosts file' default it could brake some of our automation which are using git-client (usually tests should connect to clean server where no entries in known_hosts at that moments).

Perhaps we could mitigate this by making the default be `Main.isUnitTest ? new AcceptFirstConnectionStrategy() /* Or NoHostKeyVerificationStrategy()  */ : new KnownHostsFileVerificationStrategy()`. Test environments that are running old versions of OpenSSH are a plugin developer problem, not a Jenkins user problem, and generally speaking the automated tests themselves are not going to be connecting to real external systems via SSH anyway, so I am not sure if we even care about potential MitM attacks.

> Agree, the most visible would be to print one more message below this [one](https://github.com/jenkinsci/git-client-plugin/blob/master/src/main/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifier.java#L22)
listener.getLogger().println(“If you are using OpenSSH < 7.6 please choose another strategy to verify ssh host key in ‘Manage Jenkins’ -> ‘Configure Global Security’ -> ‘Git Host Key Verification Configuration'”);

I think that kind of message is a good idea, but is there a way for us to wrap some operation in a try/catch and only print that message if the checkout fails?

Pldi23:
Created #882 as alternative based on discussion with @dwnusbaum

src/main/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifier.java
```diff
@@ -18,9 +19,28 @@ public class AcceptFirstConnectionVerifier extends HostKeyVerifierFactory {
...
+                boolean isNew = hostKeyLines.stream()
+                        .map(line -> line.split("\\s+")[0])
```
jtnord, line 32: I do not beleive this correctly handles the syntax for older ssh versions (which is why we are here)

whilst newer versions will use `|1|kbrFoDAcx....` older versions do not and are handled manually by checking for a `String.equals` or `String.contains` -  but this implementation is not correct for several cases.


The fallthought would be 
`(hostname.contains(url.getHost()) && (url.getPort() <= 0 || hostname.contains(String.valueOf(url.getPort())`

however if the `url` was `10.10.20.10:20` and the entry contained was `10.10.20.10:10` then this would say it was not new when infact it is.

There is also the possibility of multiple entries with the same key

e.g.

```
github.com,192.30.252.130 ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==
```

is a valid entry.  (the same key is used by github.com and the IP address (that is actually github.com)

I do not think this code works at all correctly in these cases either.

Pldi23: Thanks James, yes, there could be several tricky cases which current implementation do not take into account currently. For now Mark prefers #882 over this PR. If once decision will be changed I'll make this changes.

Pldi23:
closed because #882 was chosen as the solution

PR #882, opened 2022-07-29
[JENKINS-69149] Set 'Known hosts file' strategy as default

Pldi23:
## [JENKINS-69149](https://issues.jenkins.io/browse/JENKINS-69149) Set 'Known hosts file' strategy as default

Alternatively to #881.
Change default ssh host key checking strategy to 'Known hosts file' and add hint log for users that uses cli git and 'Accept first connection' to avoid using it on OpenSSH < 7.6.


MarkEWaite:
I prefer this over #881 but want to do some interactive testing with CentOS 7 systems and want to check the upgrade behavior for users that have already installed the git client plugin with the security fix.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2697,6 +2698,9 @@ private String launchCommandIn(ArgumentListBuilder args, File workDir, EnvVars e
 
             return stdout;
         } catch (GitException | InterruptedException e) {
+            if (getHostKeyFactory() instanceof AcceptFirstConnectionVerifier) {
+                listener.getLogger().println("If you are using OpenSSH < 7.6 please choose another strategy to verify ssh host key in ‘Manage Jenkins’ -> ‘Configure Global Security’ -> ‘Git Host Key Verification Configuration'");
```
dwnusbaum, line 2702: Can we tighten the exception check for this specific message? For example, `if (e.getMessage().contains("unsupported option "accept-new")) { ... }`.

Also, it could be nice to make the message into a hyperlink to https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification using `HyperlinkNote`.

Also, I think you need to update `README.md` to explain the new default setting.

One other thing - for Jenkins instances on fresh servers, `KnownHostsFileVerificationStrategy` will fail because `known_hosts` will be empty, right? That was one nice thing about `AcceptFirstConnectionVerifier`. Maybe we should also check what that error message is like and print a special message here in that case as well.

Pldi23: >Can we tighten the check for printing this message? For example, if (e.getMessage().contains("unsupported option "accept-new")) { ... }.

>Also, it could be nice to make the message into a hyperlink to https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification using HyperlinkNote.

>Also, I think you need to update README.md to update the reference to the default setting. Maybe a help.html file also needs to be updated?

done in [9dea3f2](https://github.com/jenkinsci/git-client-plugin/pull/882/commits/9dea3f22efd47be226dd568c9dcb866f8b99e1b1)


> One other thing - for Jenkins instances on fresh servers, KnownHostsFileVerificationStrategy will fail because known_hosts will be empty, right? That was one nice thing about AcceptFirstConnectionVerifier. Maybe we should also check what that error message is like and print a special message here in that case as well.

In that case the error message would be a 'usual' message, signalling that no host keys was provided for Host, see 
console screenshots:

Cli git:
CliGit console
 
JGit:
JGit console

I guess on fresh servers `known_hosts` file most likely not exists, what if we manually check it and print hint message in that case? see [f8d6af0](https://github.com/jenkinsci/git-client-plugin/pull/882/commits/f8d6af0386d821e1b00b2b6105a727974752a733)

MarkEWaite:
I'd like to see additional interactive testing to check some cases that I think might hide surprises.  I hope to assist with that Monday or Tuesday so that we can merge and release this change this in the first week of August.

Interactive test results I'd like to see:

* New install of controller on CentOS 7 (or AWS Linux 2 or ...), confirm that known hosts strategy is used as default and that ssh authenticated repository clone fails as expected because the known hosts file is not yet available.  Confirm that once the known hosts file is created on the controller, then ssh authenticated repository clone passes
* New install of agent on CentOS 7, confirm that known hosts is used as default and that ssh authenticated repository clone fails as expected because the known hosts file is not yet available.  Confirm that once the known hosts file is created on the agent, then ssh authenticated repository clone passes
* Upgrade from git client plugin 3.11.0 on CentOS 7 controller, confirm that known hosts is used as default and that it fails until the known hosts file is create
* Upgrade from git client plugin 3.11.0 on CentOS 7 agent, confirm that known hosts is used as default and that it fails until the known hosts file is create
* Upgrade from git client plugin 3.11.0 on CentOS 7 controller, confirm that known hosts is used as default and that it fails until the known hosts file is create
* Upgrade from git client plugin 3.11.1 on CentOS 7 controller, confirm that previously configured strategy is not reset by the installation of the new plugin version and that it fails until the known hosts file is create

Pldi23:
Thanks @MarkEWaite , I’ve tested all 5 cases (p3 and p5 looks the same) locally on MacOS + docker linux agent. Please re-verify fix on CentOS 7.

Could you also please test that hint message for AcceptFirstConnection displayed correctly (I don’t have environment to reproduce `unsupported option "accept-new”.` exception)?

Steps to quickly test it:

1. use System with OpenSSH < 7.6.
2. use cli git as Git installation.
3. use 'Accept first connection' as strategy.
4. start a build and check console displays message "If you are using OpenSSH < 7.6 please choose another strategy to verify ssh host key in ‘Manage Jenkins’ -> ‘Configure Global Security’ -> ‘Git Host Key Verification Configuration'".

src/main/java/org/jenkinsci/plugins/gitclient/verifier/KnownHostsFileVerifier.java
```diff
@@ -54,4 +64,11 @@ public boolean verifyServerHostKey(String hostname, int port, String serverHostK
         }
     }
 
+    private void logHint(TaskListener listener) {
+        listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification/","You're using 'Known host file' strategy to verify ssh host keys," +
+                " but your known_hosts file not exists, please go to " +
```
jtnord, line 69:
```suggestion
                " but your known_hosts file does not exist, please go to " +
```

src/main/java/org/jenkinsci/plugins/gitclient/verifier/KnownHostsFileVerifier.java
```diff
@@ -66,8 +66,8 @@ public boolean verifyServerHostKey(String hostname, int port, String serverHostK
 
     private void logHint(TaskListener listener) {
         listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification/","You're using 'Known host file' strategy to verify ssh host keys," +
-                " but your known_hosts file not exists, please go to " +
-                "‘Manage Jenkins’ -> ‘Configure Global Security’ -> ‘Git Host Key Verification Configuration' " +
+                " but your known_hosts file does not exist, please go to " +
+                "‘Manage Jenkins’ -> ‘Configure Global Security’ -> ‘Git Host Key Verification Configuration‘ " +
```
jimklimov, line 70: Should there be non-ASCII "funny quotes" emitted here? I think standard single-quotes can suffice to keep things simple?

Pldi23: let's use simple quotes to be safe

src/main/java/org/jenkinsci/plugins/gitclient/verifier/KnownHostsFileVerifier.java
```diff
@@ -54,4 +64,11 @@ public boolean verifyServerHostKey(String hostname, int port, String serverHostK
         }
     }
 
+    private void logHint(TaskListener listener) {
+        listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification/","You're using 'Known host file' strategy to verify ssh host keys," +
```
MarkEWaite, line 68: The trailing slash needs to be removed.  The id on the page does not include a trailing slash
```suggestion
        listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification","You're using 'Known host file' strategy to verify ssh host keys," +
```

MarkEWaite (changes requested):
Thanks for the changes.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2697,6 +2698,10 @@ private String launchCommandIn(ArgumentListBuilder args, File workDir, EnvVars e
 
             return stdout;
         } catch (GitException | InterruptedException e) {
+            if (e.getMessage().contains("unsupported option \"accept-new\"")) {
+                listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification/",
```
MarkEWaite, line 2702: Remove trailing slash from anchor in page
```suggestion
                listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification",
```

src/main/java/org/jenkinsci/plugins/gitclient/verifier/KnownHostsFileVerifier.java
```diff
@@ -54,4 +64,11 @@ public boolean verifyServerHostKey(String hostname, int port, String serverHostK
         }
     }
 
+    private void logHint(TaskListener listener) {
+        listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification","You're using 'Known host file' strategy to verify ssh host keys," +
```
MarkEWaite, line 68: Usually phrased as plural
```suggestion
        listener.getLogger().println(HyperlinkNote.encodeTo("https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification","You're using 'Known hosts file' strategy to verify ssh host keys," +
```

README.adoc
```diff
@@ -109,7 +109,7 @@ It is best to disable Windows Credentials Manager when installing Git on Jenkins
 == Ssh Host Key verification
 
 Git Client plugin provides various options to verify the SSH keys presented by Git repository host servers.
-By default, Git Client plugin uses "Accept first connection" strategy, which automatically adds keys for hosts that have not been seen before to `known_hosts`, and does not allow connections to previously-seen hosts with modified keys.
+By default, Git Client plugin uses "Known hosts file" strategy, which verifies all host keys using the known_hosts file.
```
MarkEWaite, line 112:
```suggestion
By default, Git Client plugin uses the "Known hosts file" strategy to verify all host keys using the known_hosts file.
```

MarkEWaite (approved):
I think we're ready to merge and release.  Any objections from other reviewers?

Do we need to backport this change to other branches that were used for the security fix?

basil (approved):

MarkEWaite:
@Pldi23 I think that we should merge the 3.11.x branch into the master branch rather than merge this pull request.  That has the side effect of squashing all the changes made in this pull request into the few that arrived on the 3.11.x branch.  Is that OK with you?

Pldi23:
>@Pldi23 I think that we should merge the 3.11.x branch into the master branch rather than merge this pull request. That has the side effect of squashing all the changes made in this pull request into the few that arrived on the 3.11.x branch. Is that OK with you?

@MarkEWaite yes, sure

MarkEWaite:
Thankkos @Pldi23 .  I'll release 3.10.0.2 now and will try to release 3.11.2 within the next hour.

MarkEWaite:
Thanks to @Pldi23 for all the work on this one.  Included in the master branch from #888.  Closing this pull request

PR #923, opened 2022-10-29
Allow file:// clones of submodules in tests

MarkEWaite:
## Allow file:// clones of submodules in tests

Allow local git clones to use the file:// protocol by setting protocol.file.allow=always in the global git configuration for the current user.

Command line git 2.38.1 and patches to earlier versions disallow local git submodule cloning with the file:// protocol. The change resolves a security issue but that security issue is not a threat to these tests.

## Checklist

## Types of changes

- [x] Test improvement

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2314,6 +2314,38 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         }
     }
 
+    /* Arguments that will be inserted into every command line git
+     * call immediately after the "git" command.  Intended to be used
+     * for specific testing situations internal to the plugin.
+     *
+     * NOT INTENDED FOR USE OUTSIDE THE PLUGIN.
+     */
+    private List<String> extraGitCommandArguments = new ArrayList<>();
+
+    /* package protected for use in tests.
+     *
+     * NOT INTENDED FOR USE OUTSIDE THE PLUGIN.
+     */
+    void setExtraGitCommandArguments(@NonNull List<String> args) {
+        extraGitCommandArguments = new ArrayList<>(args);
+    }
+
+    /* package protected for use in tests.
+     *
+     * Allow local git clones to use the file:// protocol by setting
+     * protocol.file.allow=always on the git command line.
+     *
+     * Command line git 2.38.1 and patches to earlier versions
+     * disallow local git submodule cloning with the file:// protocol.
+     * The change resolves a security issue but that security issue is
+     * not a threat to these tests.
+     *
+     * NOT INTENDED FOR USE OUTSIDE THE PLUGIN.
+     */
+    void allowFileProtocol() {
+        setExtraGitCommandArguments(Arrays.asList("-c", "protocol.file.allow=always"));
+    }
+
     /* Escape all double quotes in filename, then surround filename in double quotes.
      * Only useful to prepare filename for reference from a DOS batch file.
      */
@@ -2647,6 +2679,10 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         // if we haven't already set it.
         if (!env.containsKey("GIT_ASKPASS")) {
             freshEnv.put("GIT_ASKPASS", "echo");
+        }
+        /* Prepend extra git command line arguments if any */
+        if (!extraGitCommandArguments.isEmpty()) {
+            args = args.prepend(extraGitCommandArguments.stream().toArray(String[]::new));
         }
         String command = gitExe + " " + StringUtils.join(args.toCommandArray(), " ");
         try {
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2321,7 +2321,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      * NOT INTENDED FOR USE OUTSIDE THE PLUGIN.
      */
     @NonNull
-    private List<String> extraGitCommandArguments = new ArrayList<>();
+    private List<String> extraGitCommandArguments = Collections.<String>emptyList();
 
     /* package protected for use in tests.
      *
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2327,7 +2327,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      *
      * NOT INTENDED FOR USE OUTSIDE THE PLUGIN.
      */
-    void setExtraGitCommandArguments(@NonNull List<String> args) {
+    private void setExtraGitCommandArguments(@NonNull List<String> args) {
         extraGitCommandArguments = new ArrayList<>(args);
     }
 
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -2685,7 +2685,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
         }
         /* Prepend extra git command line arguments if any */
         if (!extraGitCommandArguments.isEmpty()) {
-            args = args.prepend(extraGitCommandArguments.stream().toArray(String[]::new));
+            args = args.prepend(extraGitCommandArguments.toArray(new String[0]));
         }
         String command = gitExe + " " + StringUtils.join(args.toCommandArray(), " ");
         try {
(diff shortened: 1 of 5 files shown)


src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -258,6 +259,32 @@ public void setGitClient() throws IOException, InterruptedException {
         gitCmd.run("config", "user.email", "email.from.git.client@example.com");
     }
 
+    private boolean fileProtocolIsAllowed = false;
+
+    /**
+     * Removes the protocol.file.allow setting from the global git
+     * configuration of the current user.
+     */
+    @After
```
jglick, line 299: Better to factor this out into a common method, even a `TestRule`.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -439,6 +438,58 @@ public void tearDown() throws Exception {
...
+     * than that we defend the git settings of git plugin developers
+     * from changes performed by the tests.
+     */
+    protected void allowFileProtocol() throws Exception {
+        w.launchCommand("git", "config", "--global", "protocol.file.allow", "always");
+        fileProtocolIsAllowed = true;
+    }
+
+    /**
+     * Removes the protocol.file.allow setting from the global git
```
jglick, line 469: What if the developer’s Git config deliberately included this flag before running tests?

jglick (commented):
This does not seem like a good idea. What if the test JVM crashes? Etc.

Acc. to https://git-scm.com/docs/git-config#FILES you can set `XDG_CONFIG_HOME` at least on Unix.

You could also just fix `git-client` to pass `-c protocol.file.allow=always` when some internal flag is set, which we would set during tests.

jglick:
(BTW when you force-push to a PR branch you break GH’s review flow, making it impossible to see at a glance what is new vs. what I already reviewed.)

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
MarkEWaite: This will remove the flag even if it was previously set.  I didn't feel that was a critical condition to preserve in the developer's git config.  If you feel that it is, I'm happy to investigate some way of detecting the current state and preserving the current state.  I've not been very successful with those techniques in the past, but I'm willing to try again if it is a serious concern.

jglick:
In the longer term, I would suggest redesigning CLI Git tests to use testcontainers.org so that tests run in a well-defined environment, rather than suddenly breaking the moment someone happens to update the Git version used in CI machines for purposes of plugin source code checkout. I was in the process of a similar conversion in `mercurial-plugin` when I worked on it.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
MarkEWaite: I'll see if I can find a way to factor it into a common location.  

I've never written a `TestRule` before and I'm running out of weekend time that I can invest, so if a `TestRule` is needed, then I'll likely need to wait a few weeks before I can address the issue.  Next weekend and the following weekend will be filled with watching the grandchildren while their parents help a Lego League event.

MarkEWaite:
> (BTW when you force-push to a PR branch you break GH’s review flow, making it impossible to see at a glance what is new vs. what I already reviewed.)

Sorry about that.  I had hoped that I had "won the race" and force pushed before you started reviewing.  Will make sure that I don't force push in the future.  Thanks for the review comments.

I like the idea of adding an extra argument to command line git calls in these special cases.  Would completely avoid changes to global git configuration.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
jglick: https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExternalResource.html is the easiest example. Just a suggestion of an idiomatic way to remove the duplicated code.

MarkEWaite:
@jglick I've simplified the code significantly and switched to using a command line modifier rather than a global configuration modifier.  Once the tests are successful, it is ready for a review.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
MarkEWaite: Resolved by switching from a change of global configuration to changing the arguments to command line git.  Thanks for the suggestion!

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2314,6 +2314,41 @@ void fixSshKeyOnWindows(Path key) throws GitException {
...
+     *
+     * Allow local git clones to use the file:// protocol by setting
+     * protocol.file.allow=always on the git command line.
+     *
+     * Command line git 2.38.1 and patches to earlier versions
+     * disallow local git submodule cloning with the file:// protocol.
+     * The change resolves a security issue but that security issue is
+     * not a threat to these tests.
+     *
+     * NOT INTENDED FOR USE OUTSIDE THE PLUGIN.
```
jglick, line 2346: Clear enough from lack of `public` modifier, but OK.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -2648,6 +2683,10 @@ private String launchCommandIn(ArgumentListBuilder args, File workDir, EnvVars e
         if (!env.containsKey("GIT_ASKPASS")) {
             freshEnv.put("GIT_ASKPASS", "echo");
         }
+        /* Prepend extra git command line arguments if any */
+        if (!extraGitCommandArguments.isEmpty()) {
+            args = args.prepend(extraGitCommandArguments.stream().toArray(String[]::new));
```
jglick, line 2688: or more simply (and [faster](https://stackoverflow.com/a/55414113/12916)!)
```suggestion
            args = args.prepend(extraGitCommandArguments.toArray(new String[0]));
```

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -258,12 +257,21 @@ String contentOf(String path) throws IOException {
             return FileUtils.readFileToString(file(path), "UTF-8");
         }
 
+        private CliGitAPIImpl cachedCliGitAPIImpl = null;
```
jglick, line 260: Why are you bothering with a cache of a flyweight object in a unit test?

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -924,10 +931,10 @@ public void testGetSubmodules() throws Exception {
         assertFixSubmoduleUrlsThrows();
     }
 
-    @Ignore("TODO see comment in CliGitAPIImplTest.setupGitAPI")
     @Test
     public void testSubmoduleUpdateShallow() throws Exception {
         WorkingArea remote = setupRepositoryWithSubmodule();
+        w.cgit().allowFileProtocol(); // CLI git 2.38.1 requires protocol.file.allow=always for this test
```
jglick, line 937: I think the comment can be dropped here—Javadoc of the method being called explains this in detail.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientTest.java
```diff
@@ -258,6 +256,23 @@ public void setGitClient() throws IOException, InterruptedException {
         gitCmd.run("config", "user.email", "email.from.git.client@example.com");
     }
 
+    /**
+     * Allow local git clones to use the file:// protocol by setting
+     * protocol.file.allow=always on the git command line of the
+     * GitClient argument that is passed.
+     *
```
jglick, line 263:
```suggestion
     * <p>
```

for anyone actually reading as HTML

jglick (approved):
Glad to see that this worked!

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
MarkEWaite: > Why are you bothering with a cache of a flyweight object in a unit test?

There are tests that fail unless the same cgit object is returned for multiple calls to a WorkingArea object.  The case that inspired it was JGit as the tested implementation, but the test needs to use CGit features to create submodules before exercising JGit to confirm the submodule checkout works as expected.  The caching makes the change "sticky".

I can add that information as a comment on the cache so that my future self and others can read the rationale in the source code.

jglick: Yes a comment would be helpful, since the term “cache” is usually understood as solely a performance optimization.

MarkEWaite: Done in 0a828e3f3caf241efe3b16d51d5038a258035c36

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
MarkEWaite: Done in 7f8b9879477b0fd68b667720e9cbf46237e20dbe

jglick (approved):

PR #939, opened 2022-12-03
Require Java 11 and Jenkins 2.361.4 or newer

MarkEWaite:
## Require Java 11 and Jenkins 2.361.4 or newer

- Test Java 11 and Java 17 in CI
- Do not include JGit or Java 8 links in Javadoc
- Remove site generation customization
- Require Jenkins 2.371.4 or newer
- Use JGit 6.4.0
- Update JGit javadoc link to JGit 6.4.0
- Remove Java 8 reference from contributing guide

Switch from Java 8 to Java 11 as minimum Java version.  Use JGit 6.4.0 instead of JGit 5.13.x.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc) doc
- [x] I have added tests that verify my changes
- [x] Unit tests pass locally with my changes
- [x] I have added documentation as necessary
- [x] No Javadoc warnings were introduced with my changes
- [x] No spotbugs warnings were introduced with my changes
- [x] I have interactively tested my changes

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Further comments

JGit project has dropped support for Java 8 with the release of JGit 6.0.  It is time to run exclusively on Java 11 or newer.

- [x] Needs #940 to be merged to the master branch before this can be merged.
- [x] JMH benchmark comparison between JGit 5.13.1 and JGit 6.4.0.  The JMH comparison shows no significant regressions in JGit 6.4.0 performance compared to JGit 5.13.1.  Measurements on large repository fetch operations showed a small improvement in JGit 6.4.0 compared to JGit 5.13.1.  No attempt was made to confirm the details of the differences, only happiness that JGit 6.4.0 seems to consistently match or exceed JGit 5.13.1 performance
- [x] https://github.com/jenkinsci/bom/pull/1619 needs to be run successfully to confirm that this pull request does not break something in the plugin bill of materials.
- [x] The fetch/clone failures on my cluster with FreeBSD agents running Java 11.0.16 were resolved by restarting the agents after installation of the pre-release that includes JGit 6.4.0

PR #944, opened 2022-12-05
Allow tests from shallow clone repos

MarkEWaite:
## Allow tests from shallow cloned repositories

Do not attempt to use the local repository as a reference repository when the local repository is a shallow clone.  A shallow clone repository has no content to offer as a reference.  Fixes 64 test failures when the test repository is a shallow clone.

https://github.com/jenkinsci/bom/pull/1613 describes the experiment to run the plugin compatibility tests with shallow clones.

https://github.com/jenkinsci/git-plugin/pull/1365 implemented a similar change in the git plugin so that shallow cloned repositories will not fail the tests in that plugin.

Will make a few tests in a shallow repository a little slower because they will require a full clone of the git client plugin repository rather than reusing the existing clone in the test workspace.  Doubtful that the performance difference will be detectable within the general variability of CI test jobs.

## Checklist

## Types of changes

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
@@ -347,6 +347,11 @@ public abstract class GitAPITestUpdate {
         w = new WorkingArea();
     }
 
+    private boolean isShallow() {
+        File shallowMarker = new File(".git", "shallow");
+        return shallowMarker.isFile();
+    }
+
     /**
      * Populate the local mirror of the git client plugin repository. Returns
      * path to the local mirror directory.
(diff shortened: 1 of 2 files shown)


MarkEWaite:
@dependabot close

Included in #939 as part of the larger change from JGit 5.13.1 to JGit 6.4.0 and from Java 8 to Java 11.

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITestUpdate.java
```diff
@@ -372,7 +377,13 @@ protected String localMirror() throws IOException, InterruptedException {
                      * the final destination directory.
                      */
                     Path tempClonePath = Files.createTempDirectory(targetDir.toPath(), "clone-");
-                    w.launchCommand("git", "clone", "--reference", f.getCanonicalPath(), "--mirror", "https://github.com/jenkinsci/git-client-plugin", tempClonePath.toFile().getAbsolutePath());
+                    String repoUrl = "https://github.com/jenkinsci/git-client-plugin.git";
```
jglick, line 380: Fine as a hotfix, but https://github.com/jenkinsci/git-plugin/pull/1365#pullrequestreview-1198101408

PR #954, opened 2023-01-07
[JENKINS-70271] - Support JGit shallow clone

MarkEWaite:
[JENKINS-70271] - Support JGit shallow clone

[JENKINS-70271](https://issues.jenkins.io/browse/JENKINS-70271) notes that JGit 6.3.0 adds support for shallow clone. The shallow clone support that it adds is available for the fetch command, but is not available for submodules.  

This pull request allows shallow clone for JGit and updates several tests to use JGit shallow clone.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc) doc
- [x] I have referenced the Jira issue related to my changes in one or more commit messages
- [x] I have added tests that verify my changes
- [x] Unit tests pass locally with my changes
- [x] I have added documentation as necessary
- [x] No Javadoc warnings were introduced with my changes
- [x] No spotbugs warnings were introduced with my changes
- [ ] I have interactively tested my changes

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Further comments

Needs further testing to confirm that there are no unexpected behaviors in normal operations or in upgrades from previous versions.

PR #956, opened 2023-01-09
[JENKINS-69159] Replace jsch with Apache Mina ssh in JGit

MarkEWaite:
## [JENKINS-69159](https://issues.jenkins.io/browse/JENKINS-69159) Replace jsch with Apache Mina ssh in JGit

A [2020 blog post from a jsch consumer](https://www.matez.de/index.php/2020/06/22/the-future-of-jsch-without-ssh-rsa/) notes that the maintainers of the jsch library are not responding to requests.  The jsch library was last released in November 2018.

The JGit project provides the [org.eclipse.jgit.ssh.apache library](https://git.eclipse.org/r/plugins/gitiles/jgit/jgit/+/master/org.eclipse.jgit.ssh.apache/) in addition to the [org.eclipse.jgit.ssh.jsch library](https://git.eclipse.org/r/plugins/gitiles/jgit/jgit/+/master/org.eclipse.jgit.ssh.jsch/).  The org.eclipse.jgit.ssh.apache library appears to be a complete replacement for git client plugin use cases.  The JGIt project has declared [their jsch library is deprecated](https://git.eclipse.org/r/plugins/gitiles/jgit/jgit/+/master/org.eclipse.jgit.ssh.jsch/) and may be removed at any time.  Jenkins has Apache Mina API plugins already available.

A [GitHub query](https://github.com/search?q=org%3Ajenkinsci+org.eclipse.jgit.ssh.jsch&type=code) shows only the git client plugin uses org.eclipse.jgit.ssh.jsch.  Replacement looks safe because there are no known consumers of `org.eclipse.jgit.ssh.jsch`.

## Checklist

- [x] I have interactively tested my changes - **required before merge**

## Types of changes

- [x] New feature

## Further comments

Would like review from @jtnord in case I've made some mistake in dependency declarations or other usage of the Apache Mina API plugins.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,14 +100,6 @@
       <artifactId>mina-sshd-api-core</artifactId>
     </dependency>
     <dependency>
-      <!-- Use mina api sftp plugin rather than JGit transitive dependency inclusion of Apache Mina sshd sftp -->
-      <!-- This must be listed before its consumers -->
-      <groupId>io.jenkins.plugins.mina-sshd-api</groupId>
-      <artifactId>mina-sshd-api-sftp</artifactId>
-      <!-- Not included in the plugin bom yet -->
-      <version>2.9.2-50.va_0e1f42659a_a</version>
-    </dependency>
-    <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>trilead-api</artifactId>
     </dependency>
@@ -150,6 +142,21 @@
       <artifactId>org.eclipse.jgit.ssh.apache</artifactId>
       <version>${jgit.version}</version>
       <exclusions>
+        <exclusion>
+          <!-- Prevent inclusion of transitive dependency, OSGi not needed by git client plugin -->
+          <groupId>org.apache.sshd</groupId>
+          <artifactId>sshd-osgi</artifactId>
+        </exclusion>
+        <exclusion>
+          <!-- Prevent inclusion of transitive dependency, sftp not needed by git client plugin -->
+          <groupId>org.apache.sshd</groupId>
+          <artifactId>sshd-sftp</artifactId>
+        </exclusion>
+        <exclusion>
+          <!-- Prevent inclusion of transitive dependency -->
+          <groupId>net.i2p.crypto</groupId>
+          <artifactId>eddsa</artifactId>
+        </exclusion>
         <exclusion>
           <!-- provided by jenkins-core -->
           <groupId>org.slf4j</groupId>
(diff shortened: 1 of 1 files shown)


jtnord:
Please don't add a direct dependency on `eddsa` 0.3.0.  
This should be an optional dependency and an optional dependency of Mina.  (and used by other things like ssh agents).

As well as the above the EDDSA code is not FIPS compliant (currently trilead will pull a similar library in in unconditionally)

MarkEWaite:
> Please don't add a direct dependency on `eddsa` 0.3.0. This should be an optional dependency and an optional dependency of Mina. (and used by other things like ssh agents).
> 
> As well as the above the EDDSA code is not FIPS compliant (currently trilead will pull a similar library in in unconditionally)

Thanks! I've excluded the eddsa transitive dependency in 051890d7d9bb856364c975c08c58451a9ff0cd59.  I will test clone with ED-25519 private keys to confirm that we don't lose some critical functionality   There are no direct references to classes from eddsa in the JGit Java source code.

> Are you sure you need the OSGI module and can not just exclude it and bring in the correct plugins from the mina sshd library, the code is the same and subsumes individual libraries, and adds some metadata for OSGI - but as we are not in an osgi container this is just noise.

Thanks!  I've excluded the osgi transitive dependency in 051890d7d9bb856364c975c08c58451a9ff0cd59.  Thanks for noting that it should not be needed.

I've excluded the sftp transitive dependency in 051890d7d9bb856364c975c08c58451a9ff0cd59 because I don't know of any use case where the Jenkins git client plugin would need sftp support.

MarkEWaite:
Tested with ED-25519 private keys accessing GitHub, Bitbucket, and others using JGit.  No issues detected.  All cases were well behaved without the sftp transitive dependency, the osgi transitive dependency, or the eddsa transitive dependency.

MarkEWaite:
@jtnord I believe this is ready to merge.  I assume no objections from you.  

My intent is to release this improvement and #954 before the end of January 2023 as git client plugin 4.1.0 unless you're aware of a reason to delay the release.  Thanks for the earlier comments!

jtnord:
> Tested with ED-25519 private keys accessing GitHub, Bitbucket, and others using JGit. No issues detected. All cases were well behaved without the sftp transitive dependency, the osgi transitive dependency, or the eddsa transitive dependency.

the `trilead-api` which you depend on currently ships the eddsa library which is why it is working.

really needs to be a separate plugin now as there are multiple consumers of it.

image

jtnord (approved):
looks good, not tested.

PR #1000, opened 2023-04-09
Enable Spotless for code formatting

basil:
Take advantage of the new Spotless support in the plugin parent POM.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.58</version>
+    <version>4.59</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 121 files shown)


basil:
@markewaite I have been doing formatting projects for years, and this is the № 1 complaint I get. (The № 2 complaint I get is that backports become more difficult.) My canned response to the № 1 complaint is that it is easy to resolve the conflicts by running `git merge -X ours` (which will prefer local changes over the autoformatting commit) followed by `mvn spotless:apply` (which will format the local changes). And if running these two commands on each PR branch is considered too much work (another complaint I often get) then I can volunteer to do it for you. It would take me approximately 5 minutes.

MarkEWaite:
> this is the № 1 complaint I get

I'm glad that I'm not alone...

Based on the steps that you described, let me run an experiment to see if I can do that myself.  Thanks for the pointer to the `git merge -X ours` command.

MarkEWaite (approved):

PR #1005, opened 2023-04-19
Activate Spotless via Maven property rather than file

basil:
Many thanks for adopting the standard Spotless configuration, which has enabled me to streamline its activation in plugin parent POM 4.61. It is now possible to delete `.mvn_exec_spotless` and add `<spotless.check.skip>false</spotless.check.skip>` instead, which I think makes for a more readable configuration.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.60</version>
+    <version>4.61</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):
Thanks!

basil:
Thanks @MarkEWaite, were there any other places you were using `.mvn_exec_spotless` instead of the property-based activation? And if so are you interested in converting them to this new approach? Because I want to get everything moved over and remove the old file-based activation from the POM. If you don't have time to do this then I can do it if you show me the other places. I tried a GitHub search but it wasn't finding them for some reason (even though it was finding `.mvn_exec_node` files - probably too new to be indexed.)

PR #1077, opened 2023-12-06
Use gson-api plugin instead of bundling gson-api jar

jonesbusy:
## Depend on gson-api plugin instead of bundling gson-api jar

Do not bundle the gson-api jar  file as a transitive dependency when there is already a gson-api Jenkins plugin available.

Added to bom so that the version number can be retrieved from the plugin bill of materials in a future BOM release:

* https://github.com/jenkinsci/bom/pull/2723

- [x] Remove version when bom is released
- [x] Rebase PR when #1078 merged

## Checklist

## Types of changes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
         <artifactId>bom-2.401.x</artifactId>
-        <version>2612.v3d6a_2128c0ef</version>
+        <version>2641.v88e707466454</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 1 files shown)


MarkEWaite:
@jonesbusy this looks great to me.  The plugin bill of materials will release with the gson-api plugin added this Saturday.  Then this pull request can be updated to depend on that new version of the plugin bill of materials and the version property can be removed from the dependency reference.

jonesbusy:
Perfect will leave the PR like this until the  BOM release

MarkEWaite:
> Perfect will leave the PR like this until the BOM release

In the interim, I'll submit a separate pull request to update the minimum Jenkins version to 2.401.3 because I want a separate changelog entry for that change.

jonesbusy:
> In the interim, I'll submit a separate pull request to update the minimum Jenkins version to 2.401.3 because I want a separate changelog entry for that change.

Sure I will rebase when merged and remove the version when the bom is released

MarkEWaite (approved):
Thanks very much!

PR #1096, opened 2024-01-14
[JENKINS-70303] Remove leading and trailing spaces from refspec

Jagrutiti:
## [JENKINS-70303](https://issues.jenkins.io/browse/JENKINS-70303) - Remove leading and trailing spaces from refspec

If a refspec is included in the checkout scm definition of a multibranch Pipeline using JGit for checkout and the refspec has a leading space character, the checkout fails with the JGit message

```
Remote does not have  available for fetch.
```

The same refspec with leading spaces does not affect command line git checkout.  It works in either case with command line git.  Leading space handling by command line git is preferred rather than the way that the git client plugin honors the leading spaces and passes them to JGit.

Command line git does not see the issue because extra spaces between arguments of a command line program are generally not an issue.  Lucky circumstances that allow the CLI git checkout to be better behaved than the JGit checkout.

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1,5 +1,28 @@
 package org.jenkinsci.plugins.gitclient;
 
+import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
+import com.cloudbees.plugins.credentials.common.StandardCredentials;
+import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.EnvVars;
+import hudson.FilePath;
+import hudson.Launcher;
+import hudson.Launcher.LocalLauncher;
+import hudson.Proc;
+import hudson.Util;
+import hudson.console.HyperlinkNote;
+import hudson.model.TaskListener;
+import hudson.plugins.git.Branch;
+import hudson.plugins.git.GitException;
+import hudson.plugins.git.GitLockFailedException;
+import hudson.plugins.git.GitObject;
+import hudson.plugins.git.IGitAPI;
+import hudson.plugins.git.IndexEntry;
+import hudson.plugins.git.Revision;
+import hudson.util.ArgumentListBuilder;
+import hudson.util.Secret;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.ByteArrayOutputStream;
@@ -45,30 +68,6 @@ import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-
-import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
-import com.cloudbees.plugins.credentials.common.StandardCredentials;
-import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
-import edu.umd.cs.findbugs.annotations.CheckForNull;
-import edu.umd.cs.findbugs.annotations.NonNull;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import hudson.EnvVars;
-import hudson.FilePath;
-import hudson.Launcher;
-import hudson.Launcher.LocalLauncher;
-import hudson.Proc;
-import hudson.Util;
-import hudson.console.HyperlinkNote;
-import hudson.model.TaskListener;
-import hudson.plugins.git.Branch;
-import hudson.plugins.git.GitException;
-import hudson.plugins.git.GitLockFailedException;
-import hudson.plugins.git.GitObject;
-import hudson.plugins.git.IGitAPI;
-import hudson.plugins.git.IndexEntry;
-import hudson.plugins.git.Revision;
-import hudson.util.ArgumentListBuilder;
-import hudson.util.Secret;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
@@ -871,7 +870,8 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
                 }
 
                 if (refspecs == null) {
-                    refspecs = Collections.singletonList(new RefSpec("+refs/heads/*:refs/remotes/" + origin + "/*".trim()));
+                    refspecs = Collections.singletonList(
+                            new RefSpec("+refs/heads/*:refs/remotes/" + origin + "/*".trim()));
                 }
                 fetch_().from(urIish, refspecs)
                         .shallow(shallow)
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1,5 +1,28 @@
 package org.jenkinsci.plugins.gitclient;
 
+import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey;
+import com.cloudbees.plugins.credentials.common.StandardCredentials;
+import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.EnvVars;
+import hudson.FilePath;
+import hudson.Launcher;
+import hudson.Launcher.LocalLauncher;
+import hudson.Proc;
+import hudson.Util;
+import hudson.console.HyperlinkNote;
+import hudson.model.TaskListener;
+import hudson.plugins.git.Branch;
+import hudson.plugins.git.GitException;
+import hudson.plugins.git.GitLockFailedException;
+import hudson.plugins.git.GitObject;
+import hudson.plugins.git.IGitAPI;
+import hudson.plugins.git.IndexEntry;
+import hudson.plugins.git.Revision;
+import hudson.util.ArgumentListBuilder;
+import hudson.util.Secret;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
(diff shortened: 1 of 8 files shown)


Jagrutiti:
I have tested this PR using `mvn clean install`. All the test cases in  [CliGitAPIImplTest.java](https://github.com/Jagrutiti/git-client-plugin/blob/62b403af0498464046822c41f9bde39066fbeb92/src/test/java/org/jenkinsci/plugins/gitclient/CliGitAPIImplTest.java) pass.

How shall I write a test case of my own? I mean how do I test `refspecs`?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -23,7 +23,16 @@
 import hudson.plugins.git.Revision;
 import hudson.util.ArgumentListBuilder;
 import hudson.util.Secret;
-import java.io.*;
+import java.io.BufferedReader;
```
Jagrutiti, line 26: This must have happened because of my IDE settings to never import `*`. 

Should I undo this?

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
```diff
@@ -861,7 +870,8 @@ public void execute() throws GitException, InterruptedException {
                 }
 
                 if (refspecs == null) {
-                    refspecs = Collections.singletonList(new RefSpec("+refs/heads/*:refs/remotes/" + origin + "/*"));
+                    refspecs = Collections.singletonList(
+                            new RefSpec("+refs/heads/*:refs/remotes/" + origin + "/*".trim()));
```
Jagrutiti, line 874: As this hard-coded we can remove `trim()`. I was not sure if I should.

MarkEWaite (changes requested):
Thanks very much for the pull request.  The two cases seem reasonable at first glance.  The automated tests that you'll create should help confirm that they are sufficient to resolve the issue.

Have you tested this interactively based on the description in the issue report?

I think that you can add an automated test in the GitClientCloneTest.  It already has the `test_clone_refspecs()`  test that defines a new RefSpec.  You can modify that test to sometimes include leading or trailing spaces on the RefSpec or you can copy that test and  test for spaces in the RefSpec in the copied test.

I've provided a sample of the way I would write the test and pushed it to your branch as 2b04ed4bc052ed17545022ff373a9fb99c59077c .  The test is currently failing for JGit.  I think that means there are more places that will need to be changed.

I also added the `Issue` annotation to the modified test with a comment explaining the change.

I found that location by using `git grep RefSpec -- src/test` to search for existing uses of `RefSpec` in the tests.  That test seemed like the easiest one to extend for this case.  You should do the same search and confirm that my choice was reasonable.  There may be other tests that assert RefSpec contents that could be extended in a similar way.

Jagrutiti:
> Thanks very much for the pull request. The two cases seem reasonable at first glance. The automated tests that you'll create should help confirm that they are sufficient to resolve the issue.
> 
> Have you tested this interactively based on the description in the issue report?
> 
> I think that you can add an automated test in the GitClientCloneTest. It already has the `test_clone_refspecs()` test that defines a new RefSpec. You can modify that test to sometimes include leading or trailing spaces on the RefSpec or you can copy that test and test for spaces in the RefSpec in the copied test.
> 
> I've provided a sample of the way I would write the test and pushed it to your branch as [2b04ed4](https://github.com/jenkinsci/git-client-plugin/commit/2b04ed4bc052ed17545022ff373a9fb99c59077c) . The test is currently failing for JGit. I think that means there are more places that will need to be changed.
> 
> I also added the `Issue` annotation to the modified test with a comment explaining the change.
> 
> I found that location by using `git grep RefSpec -- src/test` to search for existing uses of `RefSpec` in the tests. That test seemed like the easiest one to extend for this case. You should do the same search and confirm that my choice was reasonable. There may be other tests that assert RefSpec contents that could be extended in a similar way.


I will work on this. Thanks.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
MarkEWaite: No, I like the change.  The style guide for the plugin says that wild card imports should only be used in very narrow cases for tests.

src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
MarkEWaite: The existing `trim` should be removed.  It won't change the RefSpec definition in any way because `"/*".trim()` is the same as `"/*"`

MarkEWaite:
I've included the issue description in the pull request description because it helps reviewers to have the issue described clearly in the pull request, without requiring that they open the bug report.  That's my personal preference rather than a standard across the entire Jenkins organization, but since I'm one of the maintainers of this plugin, I applied my preference.

Jagrutiti:
> I've included the issue description in the pull request description because it helps reviewers to have the issue described clearly in the pull request, without requiring that they open the bug report. That's my personal preference rather than a standard across the entire Jenkins organization, but since I'm one of the maintainers of this plugin, I applied my preference.

I noticed. I had added the description in the PR title itself.

Jagrutiti:
I was not able to fix the test case for `jgit` and `jgitapache`. I tried.

https://github.com/jenkinsci/git-client-plugin/blob/31ecb75f53a9b222d0145b774572a87714dd5403/src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java#L1547-#L1565

MarkEWaite:
> I was not able to fix the test case for `jgit` and `jgitapache`. I know the test executes this part of the code.

At the location where you're looking, the RefSpec object already includes the leading space.  It seems like it would be better to look at the locations where RefSpec objects are constructed.  Trim the spaces before passing the string to the RefSpec constructor.

I'm making a guess by that idea, but I think it is a guess that is worth exploring.

src/test/java/org/jenkinsci/plugins/gitclient/GitClientCloneTest.java
```diff
@@ -328,11 +329,16 @@ public void test_clone_refspec() throws Exception {
...
+    }
+
     @Test
+    @Issue("JENKINS-70303") // spaces at start or end of refspec should be ignored
     public void test_clone_refspecs() throws Exception {
         List<RefSpec> refspecs = Arrays.asList(
-                new RefSpec("+refs/heads/master:refs/remotes/origin/master"),
-                new RefSpec("+refs/heads/1.4.x:refs/remotes/origin/1.4.x"));
+                new RefSpec(randomSpace() + "+refs/heads/master:refs/remotes/origin/master" + randomSpace()),
+                new RefSpec(randomSpace() + "+refs/heads/1.4.x:refs/remotes/origin/1.4.x" + randomSpace()));
```
MarkEWaite, line 341: This may be exactly the wrong place for this type of test.  We generally prefer tests that are nearer to what a user would do.  This is changing an argument to the RefSpec constructor when the real end user provides a string into a data entry field on a Jenkins form or they provide a string to an option of a Pipeline `checkout` step.

Jagrutiti: The test case did pass in my local.

Jagrutiti: Which test case tests the Pipeline `checkout` step?

MarkEWaite: I don't think that we'll be able to test Pipeline checkout from this plugin because it would require a dependency on the git plugin.  That would result in a circular dependency, since the git plugin depends on this plugin.  We'll need to test interactively.

MarkEWaite:
Interactive testing of the pull request shows that a leading space in the refspec from the UI still fails to clone. The job definition is:

```xml
<project>
  <actions/>
  <description/>
  <keepDependencies>false</keepDependencies>
  <properties/>
  <scm class="hudson.plugins.git.GitSCM" plugin="git@5.2.1">
    <configVersion>2</configVersion>
    <userRemoteConfigs>
      <hudson.plugins.git.UserRemoteConfig>
        <name>origin</name>
        <refspec> +refs/heads/3.11.x:refs/remotes/origin/3.11.x</refspec>
        <url>https://github.com/jenkinsci/git-client-plugin</url>
      </hudson.plugins.git.UserRemoteConfig>
    </userRemoteConfigs>
    <branches>
      <hudson.plugins.git.BranchSpec>
        <name>3.11.x</name>
      </hudson.plugins.git.BranchSpec>
    </branches>
    <doGenerateSubmoduleConfigurations>false</doGenerateSubmoduleConfigurations>
    <gitTool>jgit</gitTool>
    <submoduleCfg class="empty-list"/>
    <extensions/>
  </scm>
  <canRoam>true</canRoam>
  <disabled>false</disabled>
  <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
  <blockBuildWhenUpstreamBuilding>false</blockBuildWhenUpstreamBuilding>
  <triggers/>
  <concurrentBuild>false</concurrentBuild>
  <builders/>
  <publishers/>
  <buildWrappers/>
</project>
```

Jagrutiti:
> ```
> hudson.plugins.git.UserRemoteConfig
> ```

I have found this link for `UserRemoteConfig`: https://javadoc.jenkins.io/plugin/git/hudson/plugins/git/UserRemoteConfig.html

But not able to find where it is called in the code.

Jagrutiti:
Hi @MarkEWaite 

I have updated almost all the references I found for `refspecs`.

One of the test cases is failing because of: 

```

ERROR]   GitClientFetchTest.test_fetch:256 » JGitInternal Exception caught during execution of merge command. org.eclipse.jgit.errors.MissingObjectException: Missing unknown dadd5f9e276467580ddb00c2403eec0e9501a6e3
[ERROR]   GitClientFetchTest.test_fetch:256 » JGitInternal Exception caught during execution of merge command. org.eclipse.jgit.errors.MissingObjectException: Missing unknown cb6a175eac031a7acf8270842a6eeda2ef1f27ad

```

I  used debugger. It fails at this line:

` newAreaWorkspace.getGitClient().merge().setRevisionToMerge(bareCommit3).execute();`

but then the next line:

` assertThat("bare3 != newArea3", newAreaWorkspace.getGitClient().revParse("HEAD"), is(bareCommit3));`
has the following id as shown by the debugger. 

`AnyObjectId[224664d7fc3196baca45c60b06a6b269cea76932]`

 What should I fix here?

MarkEWaite:
Closing in favor of the different approach being used in pull request:

* https://github.com/jenkinsci/git-plugin/pull/3989

PR #1097, opened 2024-01-15
Calculate hashCode with arguments that are consistent with equals

AniketNS:
## Calculate hashCode with arguments that are consistent with equals

I used `String.format()` function for better consistency and it is more understandable.

Instead of relying on the `hashCode` of the superclass I've used the `Object.hash()` method to generate a hashcode. I think it is easier to understand and more explicit.

## Checklist

## Types of changes

### DIFF
--- a/src/main/java/hudson/plugins/git/Branch.java
+++ b/src/main/java/hudson/plugins/git/Branch.java
@@ -50,7 +50,7 @@ public class Branch extends GitObject {
      */
     @Override
     public int hashCode() {
-        return super.hashCode();
+        return Objects.hash(name, sha1);
     }
 
     /**
--- a/src/main/java/hudson/plugins/git/Branch.java
+++ b/src/main/java/hudson/plugins/git/Branch.java
@@ -39,7 +39,7 @@ public class Branch extends GitObject {
      */
     @Override
     public String toString() {
-        return "Branch " + name + "(" + getSHA1String() + ")";
+        return String.format("Branch %s(%s)", name, getSHA1String());
     }
 
     /**


src/main/java/hudson/plugins/git/Branch.java
```diff
@@ -50,7 +50,7 @@ public String toString() {
      */
     @Override
     public int hashCode() {
-        return super.hashCode();
+        return Objects.hash(name, sha1);
```
MarkEWaite, line 53: I disagree with this choice.  The `Branch` object inherits a name field and a sha1 field from the parent `GitObject`.  By replacing the `super.hashCode()` with ` Objects.hash(name, sha1)`, it is unnecessarily inserting detailed knowledge of the relevant fields of the `GitObject` into the `Branch` object.  If the `GitObject` were to add relevant fields in the future, then this `hashCode` method would need to be changed as well.

Can you explain further why you think this is a positive change?
```suggestion
        return super.hashCode();
```

MarkEWaite:
@AniketNS thanks for the pull request.  Please do the following:

* [ ] Restore the pull request template.  The template is placed in this repository so that every pull request submitter sees the expectations of the maintainers and clearly states which items they were able to complete and which they were not able to complete.  Pull request templates are widely used in the Jenkins project so that contributors see the expectations of maintainers while creating the pull request
* [ ] Describe the changes more clearly in the pull request summary.  A Jenkins user that reads the changelog for the git client plugin will not understand the change based on the description, "Made changes in Branch.java".  Why were those changes made and how do they improve things for the user or for the plugin maintainers or for someone else?

MarkEWaite (changes requested):

src/main/java/hudson/plugins/git/Branch.java
AniketNS: Your disagreement raises a valid point, and I appreciate your thoughtful consideration. But @MarkEWaite don't you think that `hashCode` method depends upon the `GitObject.java` file itself a wrong practice? 

That's why I wanted to make this `hashCode` method independent. By explicitly specifying the fields used for hashing (name and sha1), it avoids potential inconsistencies that might arise if the superclass's `hashCode` logic were to change.

MarkEWaite: I think that you're right.  The `equals` method in the class already includes the knowledge that `name` and `sha1` are the significant fields of the class.  Using `Objects.hash()` makes the `hashCode` method consistent with the `equals` method in the class.  That seems like a valuable and helpful consistency.

AniketNS:
Hello @MarkEWaite, could you please provide some info on the current status of this PR? I'm wondering if there are any specific improvements or additional changes I should make. I appreciate your feedback.

MarkEWaite:
> Hello @MarkEWaite, could you please provide some info on the current status of this PR? I'm wondering if there are any specific improvements or additional changes I should make. I appreciate your feedback.

Please take the actions that I listed in the checklist of the [earlier comment](https://github.com/jenkinsci/git-client-plugin/pull/1097#issuecomment-1892402916).  those actions include:

* [ ] Restore the pull request template.  The template is placed in this repository so that every pull request submitter sees the expectations of the maintainers and clearly states which items they were able to complete and which they were not able to complete.  Pull request templates are widely used in the Jenkins project so that contributors see the expectations of maintainers while creating the pull request
* [ ] Describe the changes more clearly in the pull request summary.  A Jenkins user that reads the changelog for the git client plugin will not understand the change based on the description, "Made changes in Branch.java".  Why were those changes made and how do they improve things for the user or for the plugin maintainers or for someone else?

AniketNS:
Hey @MarkEWaite, I've changed the Pull request name but can you please tell me now how I can restore the pull request template? 

Even if I copy the content from [this](https://github.com/jenkinsci/git-client-plugin/pull/1099) pull req template then how can I add that checklist?

MarkEWaite:
> Even if I copy the content from [this](https://github.com/jenkinsci/git-client-plugin/pull/1099) pull req template then how can I add that checklist? Or should I close this pull request and open a new one with the default pull request template and with a clear pull request description as you told me before?

One way is to copy the [source text of the pull request template](https://github.com/jenkinsci/git-client-plugin/blob/master/.github/pull_request_template.md) and paste it into the description of this pull request.  The source text is also on your local copy of the repository if you prefer to copy it from a local text file.

AniketNS:
Thank you so much @MarkEWaite. I've made the changes as requested. Please check and let me know if I've missed anything.

MarkEWaite (approved):

PR #1106, opened 2024-02-16
Simplify Revision.clone() implementation

AniketNS:
Simplify the implementation of `Revision.clone()`

## Types of changes

### DIFF
--- a/src/main/java/hudson/plugins/git/Revision.java
+++ b/src/main/java/hudson/plugins/git/Revision.java
@@ -5,7 +5,7 @@ import static java.util.stream.Collectors.joining;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Util;
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.List;
 import org.eclipse.jgit.lib.ObjectId;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
@@ -21,8 +21,8 @@ import org.kohsuke.stapler.export.ExportedBean;
 public class Revision implements java.io.Serializable, Cloneable {
     private static final long serialVersionUID = -7203898556389073882L;
 
-    ObjectId sha1;
-    Collection<Branch> branches;
+    private ObjectId sha1;
+    private List<Branch> branches;
 
     /**
      * Constructor for Revision.
@@ -41,7 +41,7 @@ public class Revision implements java.io.Serializable, Cloneable {
      * @param sha1 a {@link org.eclipse.jgit.lib.ObjectId} object.
      * @param branches a {@link java.util.Collection} object.
      */
-    public Revision(ObjectId sha1, Collection<Branch> branches) {
+    public Revision(ObjectId sha1, List<Branch> branches) {
         /* Defensive copy to avoid caller modifying ObjectId after calling this constructor */
         this.sha1 = (sha1 == null) ? null : sha1.toObjectId();
         this.branches = branches;
@@ -85,7 +85,7 @@ public class Revision implements java.io.Serializable, Cloneable {
      * @return a {@link java.util.Collection} object.
      */
     @Exported(name = "branch")
-    public Collection<Branch> getBranches() {
+    public List<Branch> getBranches() {
         return branches;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/git/Revision.java
```diff
@@ -21,8 +21,8 @@
 public class Revision implements java.io.Serializable, Cloneable {
     private static final long serialVersionUID = -7203898556389073882L;
 
-    ObjectId sha1;
-    Collection<Branch> branches;
+    private ObjectId sha1;
+    private Collection<Branch> branches;
```
MarkEWaite, line 25: I believe that this changes the serialization of the Revision object without changing its serialVersionUID.  Changing serialization has been a source of problems in the past.  I'd much rather that we not change serialization unless there are compelling reasons that benefit users of the git client plugin.  

This pull request provides a behavior preserving transformation (a refactoring).  A refactoring benefits maintainers of the git client plugin.  That is a desirable thing, but it is not sufficient to justify the extra effort to confirm that the change of serialization is worth the effort.  Changing serialization risks that this would not be a behavior preserving transformation, since the change of serialization may actually cause a behavior change.
```suggestion
    ObjectId sha1;
    Collection<Branch> branches;
```

src/main/java/hudson/plugins/git/Revision.java
```diff
@@ -121,14 +121,14 @@ public String toString() {
...
-            clone = (Revision) super.clone();
+            Revision clone = (Revision) super.clone();
+            clone.branches = new ArrayList<>(branches);
+            return clone;
         } catch (CloneNotSupportedException e) {
             throw new RuntimeException("Error cloning Revision", e);
         }
-        clone.branches = new ArrayList<>(branches);
-        return clone;
+
```
MarkEWaite, line 131: The ci job fails with the message `[ERROR] Run 'mvn spotless:apply' to fix these violations.`.  The pull request checklist includes the item 

* [ ] I have read the [CONTRIBUTING](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc) doc

The [contributing guide](https://github.com/jenkinsci/git-client-plugin/blob/master/CONTRIBUTING.adoc#code-formatting) says:

> Source code and pom file formatting is maintained by the spotless maven plugin. Before submitting a pull request, confirm the formatting is correct with:
> 
> mvn spotless:apply

```suggestion
```

MarkEWaite (changes requested):
Thanks for the pull request @AniketNS.  Much appreciated.

I'm not willing to hide the fields of the Revision object because it changes the serialization of the object and risks compatibility problems for users.

The changes need to be formatted with `mvn spotless:apply` so that the source code continues to use a consistent formatting style.

src/main/java/hudson/plugins/git/Revision.java
AniketNS: I'll do that.

src/main/java/hudson/plugins/git/Revision.java
AniketNS: Ohh Ok, @MarkEWaite. I didn't think about serialization compatibility. However, as per the coding standard, making those variables private as part of the OOP's encapsulation would be better for the file. 

Can we change the serialization number? Or is it not possible? will you please tell me that?

src/main/java/hudson/plugins/git/Revision.java
AniketNS: Done

src/main/java/hudson/plugins/git/Revision.java
MarkEWaite: > Can we change the serialization number? Or is it not possible? will you please tell me that?

Certainly possible, but it needs much more testing than I am willing to apply to this pull request.  A change of serialization then requires testing to confirm that the previously serialized objects are correctly handled by the new code.

Changing serialization in Jenkins plugins has been a source of repeated bugs for me.  I'm not willing to risk another serialization issue for a code improvement that is more about style than about improving things for users.  There are many style flaws in the plugin that I accept rather than risking damage to a user in order to improve style.

AniketNS: Ohh, you are right. I'd rather make the suggested changes.

AniketNS:
Hey @MarkEWaite, I've made the changes you told me. Please check and tell me if it is right or if I should make any other changes.

MarkEWaite:
@AniketNS I am out of the office for another week.  I won't look further at this until after I return.

AniketNS:
Ohh ok @MarkEWaite.

MarkEWaite (approved):

PR #1107, opened 2024-02-17
Implement config with gitclient and fix test issue with auto sign setup locally

olamy:
Add the config command for GitClient but only to set config value for now.
The primary goal is to avoid local build failure when having local (user level) setup of signing
```
cat ~/.gitconfig 
.....
[commit]
	gpgSign = true
[tag]
	gpgSign = true
...
```
## Checklist

## Types of changes

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/GitAPITest.java
@@ -14,6 +14,7 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
+import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Util;
 import hudson.model.TaskListener;
@@ -42,6 +43,7 @@ import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.lib.Ref;
 import org.eclipse.jgit.transport.RemoteConfig;
 import org.eclipse.jgit.transport.URIish;
+import org.eclipse.jgit.util.SystemReader;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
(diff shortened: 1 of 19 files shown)


rsandell (approved):
Though with this approach any new test added could break it again for those with autosign.

olamy:
@rsandell Correct. I'm not how we could run some command line in the image running the test to enable auto sign before starting the build

MarkEWaite:
> I'm not how we could run some command line in the image running the test to enable auto sign before starting the build

If you don't mind a reliance on [my infrastructure](https://github.com/MarkEWaite/docker-lfs/tree/lts-with-plugins), we could extend the [freestyle job](https://github.com/MarkEWaite/docker-lfs/blob/lts-with-plugins/ref/jobs/Jenkins-Plugins-Folder/jobs/Git-Client-Folder/jobs/git-client-plugin/config.xml) that I use for git client plugin testing to include an [XShell step](https://plugins.jenkins.io/xshell/) before the build that writes that configuration entry and add an [XShell step](https://plugins.jenkins.io/xshell/) to remove it after the build.  Not a perfect solution, but it would assure that someone checks that condition periodically, without requiring changes to ci.jenkins.io and without changes to the `buildPlugin()` Pipeline library

olamy:
@MarkEWaite yes that would be still better than depending on someone's laptop and the famous `It Works (or Not) On My Machine` 
but a solution on https://ci.jenkins.io/ would be prefer but I perfectly understand this might need some security hardening...

src/test/java/org/jenkinsci/plugins/gitclient/GitAPITest.java
```diff
@@ -91,6 +93,12 @@ public GitAPITest(final String gitImplName) {
         this.gitImplName = gitImplName;
     }
 
+    public static EnvVars getConfigNoSystemEnvsVars() {
```
MarkEWaite, line 96: Would you be willing to consider using `EnvVars` rather than `EnvsVars` since it is `EnvVars` in all the other usages?

This change by itself will break compilation.  Needs to be altered for all callers as well.
```suggestion
    public static EnvVars getConfigNoSystemEnvVars() {
```

MarkEWaite (approved):
Looks good to me, though I would prefer a rename of one test method

PR #1150, opened 2024-06-15
[JENKINS-73305] Create .ssh directory with owner only permissions

MarkEWaite:
## [JENKINS-73305](https://issues.jenkins.io/browse/JENKINS-73305) Create .ssh directory with owner only permissions

When the JGit implementation needs to create a `.ssh` directory, create it with permissions only allowing access to the directory owner.  That is the common pattern used by the OpenSSH project and by POSIX systems to reduce access to the sensitive information stored in the directory.

## Testing done

Ran the CredentialsTest in a debugger with a configured 'auth-data` directory and confirmed that the modified lines are executed on my RHEL 8 development computer.  Confirmed that the resulting directory permissions were read, write, and execute for only the owner, with no other permissions.

## Checklist

## Types of changes


src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -201,10 +204,22 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
     public SshdSessionFactory buildSshdSessionFactory(@NonNull final HostKeyVerifierFactory hostKeyVerifierFactory) {
         if (Files.notExists(hostKeyVerifierFactory.getKnownHostsFile().toPath())) {
             try {
-                Files.createDirectories(hostKeyVerifierFactory
-                        .getKnownHostsFile()
-                        .getParentFile()
-                        .toPath());
+                if (isWindows()) {
```
jtnord, line 207: Here be dragons.

Don't assume that POSIX file systems == non windows systems.

You can be on Linux and have a non POSIX FS (and the inverse).

And there was (and iirc) still is a bug where when you ask for the filesystem for a path you get the default even if it should be different!

This the only reliable way to do this is to actually try, catch the exception and do a fallback unless that big is fixed.

src/main/java/org/jenkinsci/plugins/gitclient/JGitAPIImpl.java
```diff
@@ -201,10 +204,22 @@ public class JGitAPIImpl extends LegacyCompatibleGitAPIImpl {
...
+                    Files.createDirectories(hostKeyVerifierFactory
+                            .getKnownHostsFile()
+                            .getParentFile()
+                            .toPath());
+                } else {
+                    Set<PosixFilePermission> ownerOnly = PosixFilePermissions.fromString("rwx------");
+                    FileAttribute<Set<PosixFilePermission>> fileAttribute =
+                            PosixFilePermissions.asFileAttribute(ownerOnly);
+                    Files.createDirectories(
```
jtnord, line 216: I think you want createDirectory here (and possibly a different call to create any missing parents of this directory)?
(At least you don't want all directories up to the parent to be created with these perms do you?)

PR #1166, opened 2024-08-06
Declare API's that throw `GitException`

jglick:
I noticed that methods were not consistently marked as throwing `GitException` even when they in fact could. Fixed by temporarily making this extend `Exception` rather than `RuntimeException` and then fixing resulting compiler errors.


MarkEWaite (approved):
Change looks good to me.  I want to run it in plugin BOM with the weekly test to assure that there are no surprises detected there.

PR #1214, opened 2024-11-14
[JENKINS-65032] Add option to include merge commits in changelogs

railsharipov:
### Added features
* Option to Include merge commits in the changelog

### Changes
* CliGit implementation changelog format was reverted from custom format to `--format=raw`.
* Since CliGit implementation was reverted to use `raw` format the date format has changed to default (unix time).

In the past, changelog format for CliGit implementation was changed to custom format because of the `git-plugin` date parsing issue (https://issues.jenkins.io/browse/JENKINS-27097). Since then `git-plugin` date parsing has been updated: https://github.com/jenkinsci/git-plugin/blob/master/src/main/java/hudson/plugins/git/GitChangeSet.java#L282-L300

Currently following custom changelog format is used:
```
/** Equivalent to the git-log raw format but using ISO 8601 date format - also prevent to depend on git CLI future changes */
public static final String RAW =
    "commit %H%ntree %T%nparent %P%nauthor %aN <%aE> %ai%ncommitter %cN <%cE> %ci%n%n%w(0,4,4)%B";
```
The above custom format is not equivalent to git-log `raw` format since all parent commits are placed on a single line:
```
parent sha1 sha2 ...
```
Although expected output is:
```
parent sha1
parent sha2
```
Existing JGit implementation places each parent commit on a separate line so there was inconsistency between CliGit and JGit implementations.

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/ChangelogCommand.java
@@ -109,6 +109,14 @@ public interface ChangelogCommand extends GitCommand {
     ChangelogCommand max(int n);
 
     /**
+     * Include merge commits in the changelog.
+     *
+     * @param include a boolean.
+     * @return a {@link org.jenkinsci.plugins.gitclient.ChangelogCommand} object.
+     */
+    ChangelogCommand includeMergeCommits(boolean include);
+
+    /**
      * Abort this ChangelogCommand without executing it, close any
      * open resources.  The JGit implementation of changelog
      * calculation opens the git repository and will close it when the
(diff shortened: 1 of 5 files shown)


andrew-garland:
Note that this was requested before but the PR was not merged.  I would also like to include merge commits but my preference would be to simply use git log instead of git whatchanged for CliGitAPIImpl (no hablo how to implement for JGit).

https://github.com/jenkinsci/git-client-plugin/pull/237

PR #1254, opened 2025-02-14
Skip ssh host verifier tests on ci.jenkins.io

MarkEWaite:
## Skip ssh host verifier tests on ci.jenkins.io

The GitHub.com SSH abuse prevention system rejects connections from the tests when they are run on ci.jenkins.io.  These tests should only be run as end to end acceptance tests but they are not too disruptive (for now) to continue to be included in the default tests.

Additional tests should be created that use test containers so that they can use the ssh connection on the test container instead of using the GitHub.com SSH servers.

Unfortunately, I can't prioritize that test rewrite now, so these tests are disabled on ci.jenkins.io.

Also make required spaces more evident in test data

A leading space in a text block is more difficult for me to detect than a trailing space that immediately precedes a backslash escape of the newline.

### Testing done

Confirmed that automated tests pass locally where the tests are not excluded.  Final confirmation on ci.jenkins.io where the tests will fail if not excluded.

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifierTest.java
+++ b/src/test/java/org/jenkinsci/plugins/gitclient/verifier/AcceptFirstConnectionVerifierTest.java
@@ -5,6 +5,7 @@ import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.io.FileMatchers.anExistingFile;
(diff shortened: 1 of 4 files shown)


MarkEWaite:
Superseded by a cherry pick of the commit from:

* #1255

PR #1311, opened 2025-07-07
fixes for JENKINS-75860

MonthyCurious:
Fixes for the [JENKINS-75860](https://issues.jenkins.io/browse/JENKINS-75860) Jira issue

### Testing done
passed `mvn clean verify`
executed `mvn spotless:apply`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java
@@ -1630,7 +1630,7 @@ public class CliGitAPIImpl extends LegacyCompatibleGitAPIImpl {
      */
     @Override
     public @CheckForNull String getSubmoduleUrl(String name) throws GitException, InterruptedException {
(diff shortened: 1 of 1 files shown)


MarkEWaite (changes requested):
Thanks for the pull request!

Needs tests that show the problem before the change and confirm the problem is fixed by the change.  The current test that invokes `setSubmoduleUrl` then calls `getSubmoduleUrl` to confirm that the set happened.  

This change is moving the location that is written and read.  That is good, because the test passes, but the test does not confirm that the rest of the code is functional with that change.

MonthyCurious:
I added more info to the [Jira](https://issues.jenkins.io/browse/JENKINS-75860).

Since the error is only encountered when `.git/config` and `.gitmodules` are out of sync (as are mine), I think that unifying the git commands (either use `-f .gitmodules` consistently everywhere or nowhere) would be good but nothing pressing.

I know the problem and can fix my git files. Please handle this PR and the associated Jira as you like.

